PR #24, opened 2022-01-13
Updated installation instructions

### LINKED ISSUE #23: Documentation: Expand documentation for installation
I think that it would help users to take them from a new computer to installation of nlmixr2 in an instruction documentation.

Most users come from a background where they will have R installed, but at least some will come with a computer with nothing.  Also, getting the tool-chain setup correctly can be a stumbling block for some users.  While there is lots of documentation out there of how to do these things, helping users find it would be useful.

The reason I'm writing this is because I wanted to look up the best way to install R on my new Linux system.  I have it automated for other systems, but I had to look at my own instructions to ensure that I got it right.

I would assume that there would be three main sections (Linux, Windows, and Mac) and within each main section, there could be flavors (Linux distribution, Windows version, Mac version/processor type).  Ideally, where feasible, we could also have a shell script/batch file to help along the way.

billdenney:
Fix #23 while not repeating ourselves

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -66,13 +66,17 @@ We recommend you have a look at [`rxode2`](https://nlmixr2.github.io/rxode2/arti
 
 When on CRAN, you can install the released version of nlmixr from [CRAN](https://CRAN.R-project.org) with:
 
-``` r
+```{r eval=FALSE, echo=TRUE}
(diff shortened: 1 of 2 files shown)


mattfidler:
Thanks Bill. 

One less thing on the todo :smile:

PR #71, opened 2022-02-17
Clarify assertion docs

billdenney:
This clarifies some of the assertion documentation and warning to indicate nlmixr2 instead of nlmixr.


mattfidler:
Looks good to me.  I probably need an autocorrct for nlmixr to nlmixr2

PR #106, opened 2022-03-19
Use the expected variable

### LINKED ISSUE #105: R CMD Check note about envir
There is an undefined variable in nlmixr2Est0: https://github.com/nlmixr2/nlmixr2est/runs/5605797613?check_suite_focus=true#step:19:231

(PR coming momentarily)

billdenney:
Fixes #105


billdenney:
Apparently, symengine should be left in even though it is not explicitly used: https://github.com/nlmixr2/nlmixr2est/runs/5612091522?check_suite_focus=true#step:19:131

I'm not sure if the right fix is to add symengine back in or to rework the code so that symengine is not required for `nlmixr2est` (and it's only required for `rxode2`).

mattfidler:
I would require it here.  Suggests only for rxode2 so it can be used in
rstudio shiny servers

mattfidler:
It is definitely required for focei and cwres

mattfidler:
Thanks Bill

Issue #165, opened 2022-06-03
Segfault on Mac


billdenney:
While running PR #164 , there was a segfault on Mac (that I don't think the PR causes):

https://github.com/nlmixr2/nlmixr2est/runs/6719199784?check_suite_focus=true#step:19:1286

PR #168, opened 2022-06-09
167 cross estimation censoring method

mattfidler:


mattfidler:
1995bdca0b722a61d7b1e5d1dabf98c373b1d93f is the first bad commit
commit 1995bdca0b722a61d7b1e5d1dabf98c373b1d93f
Author: Matthew Fidler <514778+mattfidler@users.noreply.github.com>
Date:   Thu Jun 9 00:30:23 2022 -0500

    Unified censoring estimation methods for normal distributions

 src/censEst.h | 158 +++++++++++++++++++++++++++++++++++++++++++++++++++++++---
 src/inner.cpp | 129 ++---------------------------------------------
 2 files changed, 156 insertions(+), 131 deletions(-)

mattfidler:
@billdenney 

Based on your comments I have:

- Unified the censoring methods for normal distributions
- Added a note about why the `M2` method was using a negative value

I believe this is ready.  The checks that are failing is the mac you already pointed out in #165, and the windows check that is failing is because the github runner doesn't believe `RCurl` exists.

In addition I have added the likelihood derivatives for the `M2` which were missing before.

PR #172, opened 2022-06-10
Add failing test for NPDE and NPD

mattfidler:


mattfidler:
Perhaps they should be the same.  If the correlation is not high the decorrelation step and the sd would be the same

PR #229, opened 2022-09-21
228 preds seem wrong when mixing normal log likelihood and non normal log likelihood

mattfidler:


mattfidler:
Expands tests and shows it isn't actually different.  May be a plotting issue.

Issue #241, opened 2022-09-28
Segfault while fitting two-compartment model


billdenney:
I was fitting a 2-compartment model with ODEs.  The model I'm fitting is pretty standard:

```
mod_2cmt$ui
 ── rxode2-based free-form 2-cmt ODE model ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ── Initalization: ──
Fixed Effects ($theta):
         lcl          lvc          lvp           lq     prop.err      add.err
-0.004342655  4.429537617  5.122684798 -0.697489836  0.500000000  1.000000000

Omega ($omega):
       etalvc
etalvc    0.1

States ($state or $stateDf):
  Compartment Number Compartment Name
1                  1           center
2                  2      peripheral1
 ── μ-referencing ($muRefTable): ──
  theta    eta level
1   lvc etalvc    id

 ── Model (Normalized Syntax): ──
function() {
    description <- "Two compartment PK model with linear clearance using differential equations"
    ini({
        lcl <- c(-2.93661770114182, -0.00434265546659215, 3.01380660241922)
        label("Clearance (CL)")
        lvc <- c(1.42540485812497, 4.42953761734493, 7.53786506453078)
        label("Central volume of distribution (V)")
        lvp <- c(2.11855203868492, 5.12268479790488, 8.23101224509072)
        label("Peripheral volume of distribution (Vp)")
        lq <- c(-3.62976488170177, -0.697489836026537, 2.32065942185927)
        label("Intercompartmental clearance (Q)")
        prop.err <- c(0, 0.5)
        label("Proportional residual error (fraction)")
        add.err <- c(0, 1)
        etalvc ~ 0.1
    })
    model({
        cl <- exp(lcl)
        vc <- exp(lvc + etalvc)
        vp <- exp(lvp)
        q <- exp(lq)
        kel <- cl/vc
        k12 <- q/vc
        k21 <- q/vp
        d/dt(center) <- -kel * center - k12 * center + k21 *
            peripheral1
        d/dt(peripheral1) <- k12 * center - k21 * peripheral1
        cp <- 1e+06 * center/vc
        cp ~ add(add.err) + prop(prop.err)
    })
}
```

And, the estimation is pretty simple saem:

```
mod_2cmt_est <- nlmixr(object = mod_2cmt$ui, data = data, est = "saem")
```

```
[...]

PR #242, opened 2022-09-28
Check for environment a bit more stricter

mattfidler:


mattfidler:
All I can do for #241 is try to make this a bit better for catching the bad environment call.  In theory, this branch should help this.

PR #256, opened 2022-10-17
Make table integration more robust

### LINKED ISSUE #254: Need to allow table step to be more robust
https://github.com/nlmixr2/nlmixr2est/blob/f948885be24d3efbf72bec13bee0a84c19a6d39e/R/resid.R#L81-L126

billdenney:
Fix #254

This will iterate through integration methods and decrease hmax/increase max steps until success or all requested fallbacks have been tested.


billdenney:
Hold off on this.  When I tried it with some real data just now, I got the error:

```
Error in rxSolve(object = NULL, params = params, events = events, inits = inits,  : 
  Assertion on 'maxsteps' failed: Must be of type 'integerish', not 'NULL'.
Error: Assertion on 'maxsteps' failed: Must be of type 'integerish', not 'NULL'.
```

mattfidler:
Use as.integer. Also internally the maxevals is never changed so this doesnt match the optimization procedure.   Only atol and rtol are adjusted

mattfidler:
Which arent ajusted in the code

mattfidler:
I am going to try to fix this.

mattfidler:
This works with the example model.  The `rxode2` repository will also need to be updated for you to check.

PR #262, opened 2022-10-19
Make .foceiPreProcessData() work with tibbles

billdenney:
Related to nlmixr2/babelmixr2#52

### DIFF
--- a/R/focei.R
+++ b/R/focei.R
@@ -1326,15 +1326,12 @@ attr(rxUiGet.foceiOptEnv, "desc") <- "Get focei optimization environment"
         return(toupper(x))
       }
   }, character(1))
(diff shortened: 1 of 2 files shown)


mattfidler:
The other code should be ok.  I think that things that arent purely data frames should be sanitized by this.  Including tibbles, data.tables nlmixr fit, rxode solves etc

mattfidler:
That is the code you used that worked around the tibbles since it works for data frames too

billdenney:
My fix in babelmixr2 will be to use as.data.frame(), but I thought while I'm at it, that making a fix for tibbles and then testing both could help.

mattfidler:
I believe that 'rxode2' should hit CRAN soon, so I will be pushing 'nlmixr2est' right after (as I promised the CRAN maintainers)

PR #273, opened 2022-10-27
Re-allow ID to be something other than numeric for focei

### LINKED ISSUE #272: Error of "no rows in event table or input data" in model that previously worked
I'm getting the error linked below in a model that previously worked:

https://github.com/nlmixr2/rxode2/blob/65b8093693846560e9df320fa01a7b51f284e216/src/etTran.cpp#L1614

My best guess is that it relates to my recent changes in .foceiPreProcessData:

https://github.com/nlmixr2/nlmixr2est/blob/7494d72191077846ebb18d674aaeba84d731b50b/R/focei.R#L1319-L1350

When I tried to determine how rows are kept for processing, I could not quickly make my way through the code.  Is there an obvious reason why my changes to `.foceiPreProcessData()` would have caused that (or any other recent changes)?

billdenney:
Fix #272

### DIFF
--- a/R/focei.R
+++ b/R/focei.R
@@ -1320,19 +1320,14 @@ attr(rxUiGet.foceiOptEnv, "desc") <- "Get focei optimization environment"
   env$origData <- as.data.frame(data)
   data <- env$origData
   .covNames <- ui$covariates
-  colnames(data) <- vapply(names(data), function(x) {
-      if (any(x == .covNames)) {
-        return(x)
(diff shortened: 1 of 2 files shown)


mattfidler:
merged, about to drop covs

mattfidler:
@billdenney 

Fixed in the files, but a different solution.  Kept your test.  This is because of my bad `git` skilss

PR #275, opened 2022-10-27
Update CI workflow to current versions

billdenney:
This updates the CI workflow to the current versions which are recommended by r-lib/actions.  As many things have changed, it may need tweaks to work correctly.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -39,21 +39,6 @@ jobs:
           brew install udunits
         shell: bash
 
-      - name: Link gfortran on MacOS
-        if: runner.os == 'macOS'
-        run: |
-          sudo ln -s /usr/local/bin/gfortran-10 /usr/local/bin/gfortran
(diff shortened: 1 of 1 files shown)


billdenney:
@mattfidler , Is the macos symengine source required or can the binaries be used since they're currently up to date?

https://github.com/nlmixr2/nlmixr2est/blob/2585715eb777251333431e2a2f784dcb38a0874b/.github/workflows/R-CMD-check.yaml#L57-L60

mattfidler:
> @mattfidler , Is the macos symengine source required or can the binaries be used since they're currently up to date?
> 
> https://github.com/nlmixr2/nlmixr2est/blob/2585715eb777251333431e2a2f784dcb38a0874b/.github/workflows/R-CMD-check.yaml#L57-L60

The github runner had used a very old mac architecture in the past which meant the binaries would crash the github runner without this step because it used newer machine code than supported by the old chip.

mattfidler:
Libgit is probably not needed

PR #292, opened 2022-11-21
Fix rxode2 #364

billdenney:
Fix nlmixr2/rxode2#364

The underlying issue was that the `append` argument was not passed to the model modification function.  The `auto` argument was also not passed, so I added that, too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # nlmixr2est (development version)
 
+- A bug in model piping which did not allow models to be appended to was fixed
+  (rxode2#364)
+
 # nlmixr2est 2.1.3
 
 - Allows `$etaH` and related family to be integrated into a `saem` fit
--- a/R/model.R
+++ b/R/model.R
@@ -1,13 +1,19 @@
 #' @export
 model.nlmixr2FitCore <- function(x, ..., append = FALSE,
-                                 auto = getOption("rxode2.autoVarPiping",TRUE),
+                                 auto = getOption("rxode2.autoVarPiping", TRUE),
                                  envir = parent.frame()) {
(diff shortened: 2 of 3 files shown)


billdenney:
@mattfidler , could the code below be simplified to the suggestion below that?

https://github.com/nlmixr2/nlmixr2est/blob/e34e1d9bdae652d135a038327c3328e2a88d97c8/R/model.R#L2-L11

```r
model.nlmixr2FitCore <- function(x, ..., append = FALSE,
                                 auto = getOption("rxode2.autoVarPiping",TRUE),
                                 envir = parent.frame()) {
  rxode2::model(x = rxode2::.copyUi(x$ui), ..., append = append, auto = auto, envir=envir)
}
```

Mainly, I don't understand the need for `.nlmixr2savePipe(x)` and `rxode2::.quoteCallInfoLines`.

mattfidler:
No, it cannot.  

- `rxode2::.quoteCallInfoLines` normalizes lines to have a standard structure that is parsed later
- `.nlmixr2savePipe(x)` saves the piping information like estimation methods, data, etc so piped models do not have to specify them.

mattfidler:
I guess it may be negligible for this method, but I prefer to keep it for consistency across the methods and to catch any sort of things that I may not have anticipated.

billdenney:
@mattfidler, that makes sense.  In that case, I think that this PR is likely ready to merge, if you're good with it.

mattfidler:
I will probably review these things next week :)

mattfidler:
Or maybe tonight.

mattfidler:
Looks good.  Thanks @billdenney

PR #308, opened 2023-01-12
Make some focei methods for simulation that are closer to est model

mattfidler:


mattfidler:
It seems like the errors seen in `nlmixr2extra` have now invaded `nlmixr2est` on the CI machines...

PR #313, opened 2023-02-10
Update compilation/linker flags for StanHeaders 2.26 compatibility

andrjohns:


mattfidler:
This breaks nlmixr2est, as before...

mattfidler:
Never mind the ci may be off.

mattfidler:
Will be fixed by #319 since the stan headers are located elsewhere and do not need to be here any more

bgoodri:
If you don't need to rely on StanHeaders anymore, can you release to CRAN this week? We need to release the StanHeaders that has been years in the making but is still years behind the current Stan development branch.

mattfidler:
Same as rxode2parse

PR #320, opened 2023-03-05
Only use absolute value of the difference

mattfidler:


billdenney:
@mattfidler I have an alternate PR in the works that also has testing built in.  Give me ~5 minutes.

billdenney:
... the fix is functionally the same.

mattfidler:
Oops...

mattfidler:
Is the test the same too?

billdenney:
The test is very similar.  I just pushed it, so you can give it a look.  My guess is that your test may be better.

mattfidler:
Its about the same.

billdenney:
I don't have a preference for which gets merged.  I'm just glad that I was able to make a reprex (and make a function modification to a C++ function!).

mattfidler:
Well since you were able to do so much, I prefer to merge yours :smile:

PR #326, opened 2023-03-10
Allow for babelmixr2::pkncaControl() not to set genRxControl

billdenney:
Fix nlmixr2/babelmixr2#75


billdenney:
I didn't see a straight-forward to add a test for this change.  I added comments to cover the rationale, and I confirmed that it worked.

PR #367, opened 2023-06-23
Fix 366

### LINKED ISSUE #366: `$parFixed` shows incorrect BSV
I have a model where I mix log-scale and linear scale IIV for different parameters.  The BSV column of $parFixed shows 1.00 even though the real CV is ~300%.

The issue is probably in this code:

https://github.com/nlmixr2/nlmixr2est/blob/42589c538c722086fdd2d839f7230e2be10cf641/R/nlmixr2output.R#L131-L184

billdenney:
Fix #366 

The underlying issue was the `sqrt(.w)` on this line which is now a `sqrt(.v)` (as `valCharPrep`):

https://github.com/nlmixr2/nlmixr2est/blob/42589c538c722086fdd2d839f7230e2be10cf641/R/nlmixr2output.R#L116

I also simplified the function's complexity.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 - Printing models with correlated omega values and omega values fixed
   to zero no longer fails (#359)
+- Values in $parFixed for BSV without exponential transformation are now
+  correctly shown (#366)
 
 # nlmixr2est 2.1.6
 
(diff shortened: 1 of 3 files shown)


mattfidler:
Thanks @billdenney 

Added some stylistic changes for consistency (and added your name to the function)

PR #381, opened 2023-07-27
Quieter compile

billdenney:
This removes most of the compiler warnings that are coming from within nlmixr2est.  It should have no effects on the functionality of the code.


mattfidler:
Seemed good to me.

PR #386, opened 2023-08-01
Remove unused code

### LINKED ISSUE #385: `exists("missingTable", envir = env)` should always be `FALSE` at this location
I was tracing a model error that I was getting, and I noticed that `!exists("missingTable", envir = env)` is referenced twice at the top of `nlmixr2Est0`:

https://github.com/nlmixr2/nlmixr2est/blob/e5d994ff25fa291e72b98cb1796f7d8ce362ee0d/R/nlmixr2Est.R#L133-L135

It seems like the second call to it would always be `FALSE`.  This is supported by the fact that the line within the `if` block is untested:

https://app.codecov.io/gh/nlmixr2/nlmixr2est/blob/main/R%2Fnlmixr2Est.R#L127

Should that if block be removed?  If not, can you please add a comment for why it could be triggered based on external effects of the code above?

billdenney:
Fix #385

### DIFF
--- a/R/nlmixr2Est.R
+++ b/R/nlmixr2Est.R
@@ -130,9 +130,6 @@ nlmixr2Est0 <- function(env, ...) {
   if (!exists("missingEst", envir=env)) {
     assign("missingEst", FALSE, envir=env)
   }
-  if (!exists("missingTable", envir=env)) {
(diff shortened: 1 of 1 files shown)


mattfidler:
Probably needs to be fixed, but this isn't the fix I think.

PR #469, opened 2024-08-16
Don't give a warning or error when eta values are NULL or not estimated

billdenney:
This resolves two issues that I have seen.  But, I worry that this is a hackish way to fix it when really the `.updateParFixedGetEtaRow()` function should not be called if there are no `.eta` values that need rows.

Also, I haven't yet found a good test for these.


billdenney:
Okay, if you like it, then please go ahead and merge.  I would like to find a test case for these rows because it seems like I'm hitting some kind of edge case in my work that should be better understood (at least by me).

PR #466, opened 2024-08-21
Stop requiring binary linkage

mattfidler:
This would help when doing reverse dependency checks from CRAN and
would likely make them happy.

The steps are:

- Find all the binary linkages in `rxode2`

- Have a function that puts them in the correct place; This function
  returns `TRUE` if the binary linkage is compatible and `FALSE` if it
  isn't compatible.

- When the binary linkage is not compatible, create a mini-package
  like `rxode2` does for models that grabs the needed binary linkages
  from `rxode2` and then compiles it.  When this compiled function
  runs, it will return a pointer list

- This pointer list will then be called in an internal `nlmixr2`
  function that assigns the updated function pointers to the correct
  place. This way `rxode2` can be updated and `nlmixr2est` will not
  complain (as much).


mattfidler:
Changed to create a C level api instead; This means that it is never
binary linked.

PR #509, opened 2025-03-15
Simplify numbers in parfixed

### LINKED ISSUE #346: Simplify numbers in `$parFixed`
When a number gets big (>1000, I think), it is converted to scientific notation for `$parFixed`.  But with that, I think that the numbers are a bit less readable.  I have a table with `1e+04` in it right now where I would find it easier to see `10000`, and I'd also prefer to see the second decimal place.

What would you think of a PR that makes numbers print with less scientific notation and simplifying the scientific notation when it is presented?

My thoughts are that it would do something like the following:

* All values show at least 3 significant figures, unless they appear to be integers and are fixed.
* If a value is between 1e-2 and 1e5 (exclusive of the boundaries), show the value as-is (up to 5 characters)
* For all scientific notation exponents, drop leading zeros ("1e-04" would become "1e-4")
* For all positive scientific notation exponents, drop the "+" following the "e" (so "1e+06" would become "1e6").

The over-arching goal is to show more precision in the same number of characters and hopefully also to make tables more readable at the same time.

billdenney:


billdenney:
The commit with the message "[Simplify creation of popDf and popDfSig for models with fixed thetas](https://github.com/nlmixr2/nlmixr2est/commit/ede8521b14f479c841122e7c87af343ccf2f355e)" is a general simplification of the code to make some of the subsequent changes easier. It's not completely related to the purpose of the PR.

mattfidler:
Are you happy with this @billdenney ?

billdenney:
It's not quite yet ready.

billdenney:
@mattfidler, I think that it is ready for review now.

I did not make the change optional. The way to make them optional would be to update `formatMinWidth()` to respect some option that doesn't exist as of now. If important, I can add that.

I tried to remove all the obsolete code, but there may be a couple of bits remaining. The interactions within the code are pretty complex with environment modifications all around. I may take a crack at simplifying the environment modifications in a subsequent PR after this is integrated.

billdenney:
@mattfidler, I now think that this is ready for review. I definitely did not think it would be a 500-line PR when I started!

billdenney:
Replaced by #645

PR #511, opened 2025-03-17
Make testing quieter

### LINKED ISSUE #510: Make testing quieter
While working on the most recent PR, I had trouble finding where the issue was arising. So, I'm trying to make testing quieter to make the location of issues easier to find.

billdenney:
Fix #510


billdenney:
I think that this is ready to merge. It cuts test output down from about 50,000 lines to about 2000. There is still some excess verbosity, but this PR makes a big dent in it.

mattfidler:
Looks good to me.

PR #529, opened 2025-06-14
Update continuous integration

billdenney:


billdenney:
@john-harrold, Do you have any ideas how to fix the missing `libintl.h` error on Mac?

john-harrold:
I haven't looked at it. Can you give me a devtools install command that will trigger the error and I look into it when I get back to my computer?

billdenney:
> I haven't looked at it. Can you give me a devtools install command that will trigger the error and I look into it when I get back to my computer?

I don't know.

Hopefully helpful details: My quick search suggests that `libintl.h` comes from `gettext` which should be installed by default but may not be in an expected location. I tried installing `gettext`, and that didn't help. The underlying issue appears to come from the `StanHeaders` R package. My guess is that you may have `libintl.h` in the correct place on your system and that it is likely a library search path issue more than anything else.

billdenney:
Good idea... I don't see anything special in their setup, do you: https://github.com/stan-dev/rstan/blob/develop/.github/workflows/main.yml

billdenney:
I wasn't looking at the error carefully enough. It appears that the library is being loaded to compile `rxode2`. See the CI update over there (nlmixr2/rxode2#864) which I just started and may be the right place to make any fixes.

mattfidler:
Rerunning after rxode2 merge

PR #621, opened 2026-02-27
Enable parallel estimation with FOCEI (second try)

billdenney:


billdenney:
This is the update for the third commit:

Parallelises the FOCEi inner ETA optimisation across subjects via #pragma omp parallel for in innerOpt(). Replaces a known race in the ODE-retry pointer-swap with a std::atomic<int> generation counter; fixes the rxOptionsFreeFocei placement-new; promotes the _innerParallel flag to atomic; adds a memory fence after resetOpBadSolve; adds a hard stop in the mceta>=1 R-API branch if entered in parallel. Depends on the upstream thread-safety work already merged in rxode2 (parallel-updates, naTime fix) and rxode2ll (parallel-non-normal, thread-safety-memory-fixes). Adds tests/testthat/test-focei-parallel.R comparing cores=1 vs cores=2 on theo_sd. Known limitation: models using state-dependent lag() still crash at cores>=2; tracked as a separate rxode2 issue.

mattfidler:
This uses the wrong interface.  It need to use the new interface in the development rxode2

mattfidler:
https://github.com/nlmixr2/rxode2/pull/1006

mattfidler:
In the revision from rxode2 only thread-local tolerances are used and they are  restored in each individual.  I would prefer that implementation in case there are other routines that may need sticky tolerances (looking forward), it is higher level and would use the DRY principle.

mattfidler:
You also can work to see if you can parallelize in the presence of mceta, I think may not be thread safe, but it could be if it draws from the multivariate normal inside of rxode2/nlmixr2est.

mattfidler:
This works, but I noticed your code in rxode2 `ind_solve` will corrupt any etas where they are used in `f()`, `dur()` or the inner problem has an issue with the solve and falls back to numerical differences.

So this solution isn't complete... yet.

billdenney:
New AI-suggested info:

## Update on PR #621 (rebased onto 0af34aa8)

Addresses Matt's 2026-04-20 and 2026-05-01 review feedback:

1. **"Uses the wrong interface"** — drops the hand-rolled
   `_badSolveGeneration` contamination counter and the parallel-vs-serial
   branching in `likInner0`.  The unified retry loop now calls rxode2's
   thread-safe `atolRtolFactor_()` and `setIndTolFactor()` from PR #1006
   (already merged on main).
2. **f()/dur()/rate() ETA corruption** — adds an `EtaRestoreGuard` RAII
   helper that snapshots ETAs at the entry of `shi21EtaGeneral()` and
   restores them on every exit path (throw, NA return, doFD fallback).
3. **mceta now parallel-safe** — pre-draws all per-subject ETA samples
   serially before the parallel region (`op_focei.mcetaSamples` cube);
   worker threads read by id without R API calls.  Drops the
   `mceta >= 1 → cores=1` clamp.
4. **Mixture models now parallel-safe** — drops the defensive
   `mixIdxN != 0 → cores=1` clamp.  Mixture probabilities are computed
   at the outer `updateTheta()` level (already serial) and per-individual
   writes inside the parallel inner loop are safe.

## Known limitation (carried in NEWS.md)

Models whose `f()/dur()/rate()/lag()` depend on ETAs still crash under
`cores >= 2` because `shi21EtaGeneral()` mutates the shared `op->neq`
mid-flight.  Reproducer at `inst/reprex_fbio_eta_parallel.R`.  The
companion rxode2 PR (`parfocei-eta-fix`) introduces the
`setIndNeqOverride()` C-API and consumes it in `getSolve()`/`getAdvan()`
/`getOpIndSolve()`; remaining `op->neq` read sites in
`rxode2/src/par_solve.cpp` will need to be routed through `rxEffNeq()`
in a follow-up to fully close this case.

## Test plan

- [x] `tests/testthat/test-focei-parallel.R`:
  - baseline FOCEi cores=2 == cores=1 (theo_sd, no f/dur/lag-with-eta)
  - mceta=0 cores=2 == cores=1
  - mceta=2 (forces sampling) cores=2 does not crash, finishes finite
- [x] focei-eta-only, focei-llik test files pass under load_all.
- [x] Full `devtools::test()` suite (CI).
- [ ] ASAN/UBSAN run.

## Dependency

Requires rxode2 branch `parfocei-eta-fix` (the per-individual
neqOverride C-API at pointer-table slots 56/57).

billdenney:
This requires some changes that are proposed (and soon to be proposed) in rxode2.

Issue #507, opened 2025-03-10
In `saem` do some sort of bounds check


mattfidler:
In `expit` / `logit` the values may drift toward boundaries and keep going that direction, but have no real change to the untransformed value.  This can drift the solution to somewhere where `saem` will have numerical problems.

Issue #521, opened 2025-04-16
Message "Theta reset (ETA drift)" thousands of times


billdenney:
I have a model that has the "Theta reset (ETA drift)" message which appears to possibly go infinitely long (I stopped it after about 1000 messages). Is it possible to have `nlmixr2()` stop attempting to estimate if messages like this happen too often?

mattfidler:
It is possible.

mattfidler:
This is fixed, I think.

Issue #454, opened 2024-07-23
Theta reset moves out of bounds


billdenney:
While running a (bad) model, a theta reset moved the values outside the allowed range.  Can we modify theta reset to ensure that values are in range-- or to change to give an error if it's impossible?

mattfidler:
Anything is possible given enough time.

Assuming if you are time-costrained you can simply remove the theta reset with:

```r
foceiControl(resetThetaP=0)
``

PR #631, opened 2026-04-03
Add preprocessing hook for automatic bounded parameter transformation

### LINKED ISSUE #496: Final estimates violates the boundaries (SAEM)
Hi!

I am building a sequential zero and first order absorption model, using SAEM, here is the code:

run1 <- function(){
  ini({
    tv <- c(0, 130)
    tcl <- c(0, 7)
    tmat <- c(0, 2)
    td1 <- c(0, 0.3, 1)

    eta.v ~ 0.9
    eta.cl ~ 0.9
    eta.mat ~ 0.9

    add_sigma <- 0.9
  })
  model({
    mu1 <- log(tcl)
    mu2 <- log(tv)
    mu3 <- log(tmat)
    
    cl <- exp(mu1 + eta.cl) 
    v <- exp(mu2 + eta.v)
    mat <- exp(mu3 + eta.mat)

    d1 <- mat * (1 - td1)
    ka <- 1 / (mat - d1)
    
    d/dt(depot) = - ka * depot
    d/dt(centr) = ka * depot - (cl/v) * centr
    dur(depot) = d1

    depot(0) = 0
    centr(0) = 0
    
    cp = log(1000 * centr / v + 0.00001)
    
    cp ~ add(add_sigma)
  })
}
model1 <- nlmixr2(run1, dt, "saem")
print(model1)

But it gives me some very strange final estimations:

image

is there any options I can set to avoid violating the boundaries？

Thanks in advance！

HajarBes:
Fixes #496
Related: #507, #454, #521

## Problem

Unbounded estimation methods (SAEM, nlm, optim, etc.) have no parameter boundary enforcement. Parameters with finite bounds specified in `ini()` can drift outside their valid range during estimation, producing invalid estimates and ODE solver crashes.

## Solution

A model-level preprocessing hook that automatically detects bounded parameters from `iniDf` and rewrites the model before estimation:

- **Two finite bounds (a, b):** `param <- expit(param_untransformed, a, b)`
- **Lower bound only (a, Inf):** `param <- a + exp(param_untransformed)`
- **Upper bound only (-Inf, b):** `param <- b - exp(param_untransformed)`

The hook uses the existing `preProcessHooksAdd` infrastructure. It applies to any method marked with `attr(nlmixr2Est.<method>, "unbounded") <- TRUE` (or a function of `control` for conditional cases like `optim` and FOCEI variants). Methods that handle bounds natively (bobyqa, lbfgsb3c, nlminb) are marked `FALSE`.

External packages (e.g., babelmixr2) can register their own estimators as bounded/unbounded by setting the same attribute on their `nlmixr2Est.*` S3 method, without modifying this package.

Parameters that are fixed, residual error, or already mu-referenced with a transform are skipped.

A post-estimation hook (`preFinalParTableHooksAdd`) runs after estimation to:
- Back-transform theta from internal to natural scale
- Restore original parameter names and bounds
- Apply Jacobian correction to the covariance matrix
- Back-transform parameter history

## What's included

- `R/preProcessBoundedTransform.R` - preprocessing hook, post-estimation hook, and parHist back-transform helper
- `unbounded` attribute on all `nlmixr2Est.*` S3 methods (TRUE for unbounded, FALSE for bounded, function for conditional)
- Original UI saved in a `boundedOriginal` environment for the post-estimation hook

## Testing

Using the model from #496 (`td1 <- c(0, 0.3, 1)`):
- **Before:** `td1 = 1.182` (violates upper bound)
- **After:** `td1 = 0.528` (within bounds, original name and bounds restored in output)

Regression checks on standard SAEM and FOCEI models produce identical estimates to the pre-patch version.


HajarBes:
Before/after comparison (same seed, same model from #496, theo_sd data):

The unpatched SAEM (red) drifts td1 above the upper bound of 1.0. The patched version with logit reparameterization (green) keeps td1 within (0, 1) throughout estimation.

_(Plot generated from test runs - can provide the test script if needed.)_
fix_comparison

mattfidler:
Thanks for the pull request. 

There has been some internal debate if these sorts of parameterizations should be allowed in general.  I'm still on the fence. NONMEM, indeed does what you say; monolix is closer to `nlmixr2` in terms of the code specification and it is a bit opaque in the gui.


As for my review:

- This does not fix #507. (I don't believe)
- While the standard errors are correct, the covariance matrix may not be correct;  A Jacobian matrix transformation needs to be applied instead
-  The DRY principle should be used, the "stable" logit/expit transform should come from the rxode2 headers.

My biggest concern for this is degradation of saem by not using logit/expit in the model for mu-referenced expressions, ie using:

```r
v = tv*exp(eta.v)
```

which would then translate to

```r
v = (tv+eta0)*exp(theta0+eta.v)
```

instead of

```r
v = exp(lv + eta.v)
``` 

which is much more stable.

HajarBes:
Thanks for the review @mattfidler.

Understood on all points. I'll update the PR to:

1. Restrict logit reparameterization to mu-referenced `expit()` parameters only (via `muRefCurEval$curEval == "expit"`) rather than auto-detecting from `iniDf `bounds
2. Apply full Jacobian transformation to the covariance matrix
3. Use `_powerDi()` from rxode2 headers instead of hand-written expit

mattfidler:
The more I think about this pull request, the more I think it is at the wrong place and incomplete since it should apply to any model estimation method, not just saem.

To be complete, any model that is an unbounded optimizer, or an unbounded optimizer on certain parameters should:

- Modify the model to `expit()` for any "theta" parameters with two bounds
- Modify the model to `(lower+exp())` for anything with a lower bound but not an upper bound 
- Modify the model to `(upper-exp())` for anything with an upper bound but not a lower bound.

This model would then run the right estimation routine and then back-transform everything (including applying the jacobian for every type of transformation).

This would be in a nlmixr2 model hook that looks at the estimation method properties to determine if it is bounded or unbounded.

I also don't understand why you would only focus on the model parameters that are already in an expit block, that is 

> Restrict logit reparameterization to mu-referenced expit() parameters only (via muRefCurEval$curEval == "expit") rather than auto-detecting from iniDf bounds

HajarBes:
Thanks @mattfidler, that makes sense. I see your point about this belonging at the model level rather than inside SAEM, and about applying it more generally based on parameter bounds rather than restricting to expit-referenced parameters.

I’ll rework this toward a model-level preprocessing step that detects bounds from iniDf and applies the appropriate transformation before estimation, rather than tying it to a specific method.

mattfidler:
The hooks haven't been documented yet;

There likely needs to be an additional type hook for parameter and covariance estimation before this works.

HajarBes:
Thanks @mattfidler, that's very helpful.

I've started reworking this using the existing hook system and have a preprocessing hook that detects bounded parameters from iniDf and applies the transformations before estimation, along with back-transforms for the output.

I haven't fully integrated the covariance Jacobian correction into a post-estimation hook yet, but I'll push the current version so you can see the approach.

HajarBes:
Just to clarify, this rework is mainly to explore the model-level hook approach you suggested and make the direction concrete. Happy to adjust the structure depending on how you'd prefer this to fit into the hook system.

mattfidler:
This is excellent beginning.  There are only a few small things on your end that need to change.

First, the determination of if the estimation method is an unbounded method is fragile;  It should be based on properties of the nlmixr2 method.  This way the hook can be applied to methods inside places like babelmixr2 without having to define them here.  

For example, 

https://github.com/nlmixr2/nlmixr2est/blob/30bd9a7b316751e0c79465284f8f2640060ce76b/R/bobyqa.R#L428

Here you would need to define something like:

```r
attr(nlmixr2Est.bobyqa, "unbounded") <- FALSE
attr(nlmixr2Est.uobyqa, "unbounded") <- TRUE
attr(nlmixr2Est.focei, "unbounded") <- function(control){ control$outerOpt %in% c("uobyqa")}
```

This way anyone can apply the hook regardless of if they are in the package or not.

Then in the package you would use something like the following code:

https://github.com/nlmixr2/rxode2/blob/61c19e6bf971617edeadc6757d69e3af0f6b3a4f/R/rxUiGet.R#L40-L55

Adapted here it would be something like:

```r
.isUnboundedMethod <- function(est, control) {
   .v <- as.character(utils::methods("nlmixr2Est"))
  .method <- paste0("rxUiGet.", .est)
  if (.method %in% .v) {
   .unbounded <- attr(utils::getS3method("nlmixr2Est", est), "unbounded")
    if (is.null(.unbounded)) return(FALSE)
    if (is.function(.unbouded)) return(isTRUE(.unbounded(control)))
    isTRUE(.unbounded)
 }  else {
   FALSE
 }
}
```

Also, not only would you have to transform the parameter estimates you will also have to transform the untransformed parameter estimate history.

You will likely need to save the original function in an environment and then I will have to create a hook the applies just before the nlmixr2 fit object is completely created.

HajarBes:
Thanks, I updated the branch following your suggestion.

The main change is that unbounded detection now uses estimator-level attributes instead of a hardcoded method list. I used function attributes for the conditional cases like `optim` and the FOCEI variants, and kept the preprocessing hook using that lookup. I also saved the original UI pieces needed for the later post-estimation step and added the parameter history back-transform helper.

One thing I ran into: I first tried setting the attribute centrally on load to keep the change in one place, but `getS3method()` was not seeing that reliably, so I switched to defining the attribute directly on the estimator methods as in your example.

I also reran a couple of regression checks on standard SAEM and FOCEI models, and they still behave as expected.

HajarBes:
Thanks for adding the preFinalParTableHooks infrastructure, this made the integration point much clearer.

I added a post-estimation hook that back-transforms theta, restores original parameter names/bounds, applies the Jacobian correction to the covariance matrix, and transforms the parameter history.

Happy to adjust based on your feedback.

mattfidler:
This is very close.  

The last step is to transform the model ui (which you modified but didn't restore).  It should be the same as what you originally provided but simply with the proper parameter estimates.  

You should be able to see the fit with `fit$ui`.  I am unsure what is shown now because it doesn't do that step.  

There also have to have some sort of configuration option to revert to prior behavior (which should be checked in the `control` option).

Once completed, there should be some tests of functionality.

HajarBes:
While implementing the `fit$ui` restoration, control option, and tests, I found one extra gap worth flagging before pushing.

The post-estimation hook appears to fire through the FOCEI finalization path, but not for SAEM/non-FOCEI methods, so the back-transform was not happening for the original SAEM case. I have a local fix for that, along with the `fit$ui` restoration, `boundedTransform` control option, and tests, and it is working locally.

Since that touches the hook wiring and the control handling for the unbounded methods, I wanted to check whether you’d prefer that integration point or a different one before I push.

mattfidler:
I probably need to handle it for wherever the hook is run.  

I have only looked at the code so far, not run it.  I was unsure if the `$ui` would be present in the place with the original hook.

mattfidler:
> The post-estimation hook appears to fire through the FOCEI finalization path, but not for SAEM/non-FOCEI methods, so the back-transform was not happening for the original SAEM case. I have a local fix for that, along with the fit$ui restoration, boundedTransform control option, and tests, and it is working locally.

I didn't see that on my environment it is calling the saem fixes.

I made sure by using a `browser()`;  You probably were skipping mu-referenced saem types of parameters.  The breaking of mu-referencing degrades the fit process still and are still problematic from a methodology point of view.

I changed this to add a warning (which seems to be eaten by nlmixr2 and not shown in the fit object), but it updates everything correctly...

mattfidler:
Still need to be done:

- [x] add a control option that enables/disables this for all methods and add to the `.isUnboundedMethod`
- [ ] Change all the documentation to roxygen2 in case they may be exported in the future
- [x] Add tests for the methods to make sure they are applying the methodology
- [x] Fix the warning so that is captured in the model output like other warnings in `nlmixr2`

HajarBes:
Thanks, that makes sense regarding the mu-referenced SAEM case and the warning.

I’ve addressed the remaining items: added the control option across methods, converted to roxygen2, added tests, and fixed warning capture so they appear in the model output.

Let me know if anything should be adjusted.

mattfidler:
@HajarBes this seems OK, though there are some odd things that seem like there are some AI assisted behavior (not fixing roxygen for inrtance), can you identify which AI assistant is used.

Since we are now part of the R consortium, (part of the Linux foundation), I think it may be needed. (See [here](https://www.tomshardware.com/software/linux/linux-lays-down-the-law-on-ai-generated-code-yes-to-copilot-no-to-ai-slop-and-humans-take-the-fall-for-mistakes-after-months-of-fierce-debate-torvalds-and-maintainers-come-to-an-agreement))

I also have to change some of the issues seen that may have been  from AI since they are unstable (like the documentation).

HajarBes:
Yes, I use Claude, with occasional GPT. I’ll read the R Consortium guidance you linked and follow it going forward.

On the documentation, the garble in `man/nlmixr2.Rd` came from a local roxygen regeneration that mis-parsed an existing docstring, which I reverted in `4cec4a43`. If you spot any other artifacts from those edits, I’m happy to clean them up myself in one pass.

mattfidler:
Its new; I'm unsure if we need to or not.

mattfidler:
The focei model is not working properly.  Even though the tests initially passed, they were actually not testing the correct behavior.

mattfidler:
focei still not working...

HajarBes:
I see it. I’ll verify the FOCEI path directly and recheck the tests

mattfidler:
These fixes work now :smile:

HajarBes:
Looks great, thanks for the FOCEI fixes and the IOV test. This came together really nicely.

Happy to have contributed. Let me know if anything else is needed on my end.

mattfidler:
The iov test is broken.  Still more work to be done :smile: 

In addition to iov needing to work, the rest of the hooks need to be tested

mattfidler:
Also the back-transformed parameter names must start with `rx` or `nlmixr2` which are documented as generated  internal `rxode2`/`nlmixr2` variables.

mattfidler:
If you want to try to fix these feel free.

HajarBes:
Had a look at the IOV test issue.

The vapply error at `R/iov.R:274` seems to come from a mismatch where `.uiIovEnv$iov$iniDf `still has tcl_untransformed while `ret$ui$iniDf` has tcl (restored by the post-est hook), so the name lookup fails.

Also, the snapshot’s `lstExpr` still references tcl_untransformed, so rebuilding `ret$ui `from it overrides the post-est hook’s name restoration.

Looks like either the IOV snapshot needs to capture the ui earlier in the pipeline, or the post-est hook needs to propagate the undo into `.uiIovEnv$iov`.

HajarBes:
I've done the rx/nlmixr2 renaming, 10/11 tests pass. Before pushing, do you have a preference for the prefix style?

- `rxBoundedTr.tcl`
- `rx.tcl.bounded` (closer to current IOV style like `rx.iov.cl.1`)
- `nlmixr2BoundedTcl`

mattfidler:
I prefer `rx` since it is smaller :smile: but have used `nlmixr` elsewhere.

Otherwise, these models are not seen so it doesn't matter.

mattfidler:
> Had a look at the IOV test issue.

I agree, just haven't figured out a good fix yet.

mattfidler:
- [x] Error (test-focei-wang2007-boxcox-lnorm.R:61:3): (code run outside of `test_that()`)
- [x] Error (test-nlme.R:84:5): nlme one compartment theo_sd; Error: unused argument (.model)
- [x] Error (test-np.R:52:5): Test naive pooled; Error in `parse(text = paste("quote(", .ls[.i], "<-", deparse1(ui$meta[[.ls[.i]]]), ")"))`: 
- [ ] Failure (test-resume-bobyqa.R:48:5): resume bobyqa Expected `grepl("bobyqa", d$extra)` to be TRUE.
- [x] Error (test-str-est.R:115:5): test nlme Error: unused argument (.model)

mattfidler:
The `resume-bobyqa` occurred before this pull request.

mattfidler:
Merged for now; the only failing test I see was pre-existing.

HajarBes:
Thanks for the merge 🎉 Appreciate the IOV fixes and the work to get this across the line. Made the feature land much cleaner than the initial PR.

Issue #641, opened 2026-06-05
Fixed effects are not updated in algebraic model


billdenney:
In the model below, the fixed effects are not estimated. A similar model (with client data) worked in a prior version of nlmixr2est:

``` r
library(nlmixr2)
#> Warning: package 'nlmixr2' was built under R version 4.5.3
#> ── Attaching packages ───────────────────────────────────────── nlmixr2 5.0.0 ──
#> ✔ lotri         1.0.4          ✔ monolix2rx    0.0.6     
#> ✔ nlmixr2data   2.0.9          ✔ nlmixr2lib    0.3.2.9000
#> ✔ nlmixr2est    6.0.1          ✔ nlmixr2rpt    0.2.2     
#> ✔ nlmixr2extra  5.0.0          ✔ nonmem2rx     0.1.9     
#> ✔ nlmixr2plot   5.0.1          ✔ posologyr     1.2.8     
#> ✔ rxode2        5.1.2          ✔ shinyMixR     0.5.3     
#> ✔ babelmixr2    0.1.11         ✔ xpose.nlmixr2 0.4.1     
#> ✔ ggPMX         1.3.2
#> ── Optional Packages Loaded/Ignored ─────────────────────────── nlmixr2 5.0.0 ──
#> ✔ babelmixr2        ✔ nonmem2rx    
#> ✔ ggPMX             ✔ posologyr    
#> ✔ monolix2rx        ✔ shinyMixR    
#> ✔ nlmixr2lib        ✔ xpose.nlmixr2
#> ✔ nlmixr2rpt
#> Warning: package 'lotri' was built under R version 4.5.3
#> Warning: package 'nlmixr2est' was built under R version 4.5.3
#> Warning: package 'nlmixr2plot' was built under R version 4.5.3
#> Warning: package 'babelmixr2' was built under R version 4.5.3
#> Warning: package 'ggPMX' was built under R version 4.5.3
#> Warning: package 'posologyr' was built under R version 4.5.3
#> Warning: package 'ggplot2' was built under R version 4.5.3
#> ── Conflicts ───────────────────────────────────────────── nlmixr2conflicts() ──
#> ✖ rxode2::boxCox()     masks nlmixr2est::boxCox()
#> ✖ rxode2::yeoJohnson() masks nlmixr2est::yeoJohnson()
library(tidyverse)
#> Warning: package 'stringr' was built under R version 4.5.3

mod <- function() {
  ini({
    e0 <- fixed(0)
[...]

PR #644, opened 2026-06-06
Make parallel FOCEI/SAEM deterministic across runs

billdenney:
speculative determinism fix, may be unnecessary once rxode2 PR lands

## Summary

After the parallel inner loop, FOCEI called `sortIds(rx, 0)` which
re-ordered subjects by the **wall-clock time** each one had taken to
solve. That made subject dispatch order at the top of the next outer
iteration depend on system load — same seed, same data, same
`cores > 1` produced different fitted parameters between runs.

Switches both call sites to `sortIds(rx, 2)` (deterministic iota
order, 1..nsub):

- `src/inner.cpp:2539` — post-parallel reorder in FOCEI `innerOpt()`.
- `src/saem.cpp:2149` — analogous reorder in SAEM's `liblsoda` path.

Combined with the per-thread tolerance arrays in rxode2 PR
https://github.com/nlmixr2/rxode2/pull/new/fix-atol-rtol-thread-safety
(which closes a separate `atolRtolFactor_` race), this eliminates the
non-determinism reported in #641 under `cores > 1`. Both fixes are
needed together; either alone is insufficient.

## Test plan
- [x] New `tests/testthat/test-focei-deterministic.R` runs the
      issue-641 reproducer twice at `cores=2L` and asserts
      `identical()` on both `objf` and `parFixedDf$Estimate`.
      Test passes once the rxode2 companion PR is on the loaded
      rxode2.
- [ ] Existing `tests/testthat/test-focei-parallel.R` continues to
      pass with its current `tolerance = 1e-8` checks; can be
      tightened to `expect_identical()` in a follow-up.

## Cost

Switching from longest-wall-time-first to iota loses
load-balancing on highly imbalanced subject populations. Expected
hit: ≤5% wall-clock on worst-case workloads, negligible on typical
PK/PD datasets.

## Related

- Companion: https://github.com/nlmixr2/rxode2/pull/new/fix-atol-rtol-thread-safety
- Independent of #641's scaleC fix (separate PR
  https://github.com/nlmixr2/nlmixr2est/pull/new/issue-641-fix).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,15 @@
+# nlmixr2est (development version)
+
+- Make parallel FOCEI and SAEM optimizer trajectories deterministic across
+  runs. The post-parallel `sortIds(rx, 0)` calls in `src/inner.cpp` (FOCEI
+  inner loop) and `src/saem.cpp` (SAEM solve) were ordering subjects by the
+  wall-time each one took to solve, leaking system load into the algorithmic
+  path. Switched to `sortIds(rx, 2)` (deterministic iota order) so dispatch
+  order at the top of each outer iteration is reproducible. Combined with
+  the per-thread tolerance arrays in rxode2 (development version), this
+  closes the non-determinism observed in
+  https://github.com/nlmixr2/nlmixr2est/issues/641 under `cores > 1`.
+
 # nlmixr2est 6.0.1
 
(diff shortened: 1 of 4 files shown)


billdenney:
Not the right solution to the problem

PR #647, opened 2026-06-07
Stable init

mattfidler:
@billdenney maybe this fixes your issue.

If so, a reprex would be helpful so it doesn't decay.

It seems you have an example that is not stable on multiple threads.


mattfidler:
... wait this is wrong.

mattfidler:
Fixed issue, maybe it will work for you @billdenney 

Please add a reprex too

billdenney:
Which issue is this solving? Possible divergence with parallel solving?

mattfidler:
In theory yes.  I don't have a reprex, so I have no clue.

PR #649, opened 2026-06-08
Flush rxode2's aggregated solve warnings per FOCEi/SAEM iter

### LINKED ISSUE #301: indy printing slows model minimization
When running an oncology nonclinical model, I see messages like "intdy -- t = 504 illegal. t not in interval tcur - _rxC(hu) to tcur" many times (thousands of times-- longer than my Rstudio scroll back buffer).

When I run the model as-is, it takes 15.1 minutes.  When I run the model within `suppressMessages()`, it takes 10.3 minutes.  The specific line of code is this one:

https://github.com/nlmixr2/rxode2/blob/7342f1deb8fb3d68411b9a750870b6f5a217401e/src/intdy.c#L57

Would it be reasonable to put some form of rate-limiter on the print statement, give a flag to suppress it, or to remove the print statement to speed things up?  (I did some looking into rate limiters, but it wasn't clear that there was one best solution.)

billdenney:
Companion to nlmixr2/rxode2 PR `fix/intdy-warn-throttle` (closes nlmixr2/rxode2#301).

## Summary
- Add a small inline helper `nmFlushRxSolveWarn(maxIds)` in [src/solveWarnHelper.h](https://github.com/nlmixr2/nlmixr2est/blob/fix/intdy-warn-throttle/src/solveWarnHelper.h) that lazily resolves `rxSolveWarnFlush` via `R_GetCCallable("rxode2", ...)`.
- Call it at the end of the FOCEi outer iteration print in `foceiOfvOptim` ([src/inner.cpp](https://github.com/nlmixr2/nlmixr2est/blob/fix/intdy-warn-throttle/src/inner.cpp)) and at the end of the SAEM per-iter print in `saem_fit` ([src/saem.cpp](https://github.com/nlmixr2/nlmixr2est/blob/fix/intdy-warn-throttle/src/saem.cpp)) so the warning summary sits next to the iteration that produced it.

## Why
rxode2 now collects `intdy` window-miss warnings into a per-package aggregator instead of printing per occurrence. Without a per-iteration flush from nlmixr2est the warnings would accumulate silently across the whole estimation; one summary line per printed iteration keeps the message aligned with the iteration that produced it and makes a flood of thousands-per-iter become one tidy line.

## Dependency
Requires rxode2 with `rxSolveWarnFlush` registered. The lazy `R_GetCCallable` lookup makes the helper a no-op on older rxode2 — bump `Imports: rxode2 (>= X.Y.Z)` in DESCRIPTION as part of the merge (next bump after the rxode2 companion PR lands).

## Test plan
- [ ] `R CMD INSTALL .` clean against the companion rxode2 branch
- [ ] `devtools::test(filter="focei-1")` passes
- [ ] FOCEi fit on `theo_sd` converges to OFV ≈ 116.99
- [ ] Run a model that historically flooded intdy and confirm: one summary line per printed iteration replaces the flood


billdenney:
This relies on nlmxr2/rxode2#1068 and #654 being merged first.

PR #651, opened 2026-06-08
Unify iteration-time print formatting across all estimators; introduce iterPrintControl()

### LINKED ISSUE #650: Feature request: Standardize iteration printing across all methods
SAEM prints iteration information to the console differently than FOCEI. I think that it would make nlmixr2est iterations easier to follow and more self-consistent within the package to have a single iteration-printing path for all algorithms. That way, we could also make printing more dynamic (if someone prefers the SAEM-formatting, we could apply that to all estimation methods optionally, for instance).

A quick AI search suggests that a lot of the infrastructure for this already exists; FOCEI and SAEM have their own, but the other methods are more standardized.

Is there interest in this change?

billdenney:
## Summary

Iteration-time progress output across every estimator (`focei`, `saem`, `bobyqa`, `nlm`, `optim`, `nls`, `nlminb`, `lbfgsb3c`, `n1qn1`, `newuoa`, `uobyqa`) now flows through one shared formatter (`scaleApplyIterPrintControl` / `scalePrintFun` / `scalePrintGrad` in `src/scale.h`) configured by a single user-facing function, `iterPrintControl()`.  Every method renders the same column layout, wrapping, ANSI handling, periodic header re-emit, and per-iteration user-interrupt check.  Closes the divergence between focei's rich `#`/`U`/`X` three-row output and saem's previous one-liner.

## What's in the PR (12 commits)

**Shared C++ pipeline (`src/scale.h`)**

- One `scaling` struct with field names that mirror `iterPrintControl()`'s R argument names (`every`, `ncol`, `headerEvery`, `useColor`, `simple`).
- One `scaleApplyIterPrintControl(&scale, ipc)` helper called by every estimator's setup code (`src/inner.cpp`, `src/saem.cpp`, `src/nlm.cpp`).
- `scalePrintFun` auto-skips degenerate rows: `U` is dropped when `scaleType == None` (it would mirror `#`); `X` is dropped when there are also no log/logit-transformed parameters.  Methods with no scaling and no transforms collapse to a single `#` row.
- `keyExtra` field lets each estimator append its own Key text (focei uses this for its G/F/C/M gradient legend + omega note) without anything in `inner.cpp` touching `RSprintf` for the header.
- `Rcpp::checkUserInterrupt()` lives inside `scalePrintFun` so every per-iteration print is interruptible without each estimator needing its own call site.

**New user-facing R API (`iterPrintControl()`)**

- Bundles all iteration-print options into one validated, classed list: `iterPrintControl(every, ncol, headerEvery, useColor, simple)`.
- Pass via the existing `print` argument on any `*Control()` function:

  ```r
  foceiControl(print = iterPrintControl(every = 5, headerEvery = 20))
  ```

- The scalar form `foceiControl(print = 5, printNcol = 8)` continues to work — internally absorbed via `.absorbIterPrintControl()`.
- Round-tripping the returned control list back through `do.call(*Control, .ctl)` is supported.

**saem now has the same transforms as focei**

- saem's default iteration print emits the same `#`/`X` layout as every other estimator (U is auto-skipped because saem has no internal optimizer scaling).
- An R-side pure helper `.iterPrintXParFromUi(ui, printNames)` inspects `ui$muRefCurEval` and returns `xPar` codes + logit bounds.  Used by both saem (per print-vector name) and focei (per `ntheta`).
- The old saem-specific simple-mode force is gone; users who want single-row output opt in with `saemControl(print = iterPrintControl(simple = TRUE))`.

**One R-side source of truth for theta transforms**

- `.foceiOptEnvSetupTransformIndexs` deleted.  Its work happens at the focei setup call site using the pure `.iterPrintXParFromUi` helper.  saem and focei now share the same R-side detection logic.

**Documentation consolidated (`R/iterPrintControl.R`)**

- A single `iterPrintParams` roxygen stub documents both the outer scalar args (`print`, `printNcol`, `useColor`) and the inner `iterPrintControl()` args (`every`, `ncol`, `headerEvery`, `useColor`, `simple`).
- Every `*Control()` function inherits via `@inheritParams iterPrintParams`.
- Existing `man/*.Rd` files are stale from before this PR series and will regenerate cleanly when `roxygen2::roxygenise()` is next run.

**`NEWS.md`** updated with a development-version section describing the unified pipeline, the new `iterPrintControl()`, the saem default change, the back-transform support, and the auto-skip behaviour.

## Example output

`foceiControl(print = iterPrintControl(every = 2, headerEvery = 3))` on a model with log-transformed thetas:

```
Key: U: Unscaled Parameters; X: Back-transformed parameters; G: Gill difference gradient approximation
F: Forward difference gradient approximation
C: Central difference gradient approximation
M: Mixed forward and central difference gradient approximation
Unscaled parameters for Omegas=chol(solve(omega));
Diagonals are transformed, as specified by foceiControl(diagXform=)
|    #| Function Val. |       tka |       tcl |        tv |    add.sd |
|.....................|        o1 |        o2 |        o3 |...........|
|-----+---------------+-----------+-----------+-----------+-----------|
|    2|     125.83249 |   -0.9618 |   -0.5715 |     1.125 |   -0.8944 |
|.....................|   -0.1650 |    0.3934 |    0.3371 |...........|
|    U|               |    0.4882 |     1.060 |     3.575 |    0.6786 |
|.....................|     1.469 |     1.938 |     2.032 |...........|
|    X|               |     1.629 |     2.885 |     35.71 |    0.6786 |
|.....................|     1.469 |     1.938 |     2.032 |...........|
|    F| Forward Diff. |     1.340 |     17.45 |     41.56 |    -6.885 |
|.....................|     18.13 |    -1.191 |    -4.839 |...........|
```

`saemControl(print = iterPrintControl(every = 1, headerEvery = 0))` on the same model — `U` is auto-skipped (no internal scaling), `X` shows back-transformed thetas:

```
Key: X: Back-transformed parameters;
|    #|       tka |rxBoundedTr.tcl |        tv | V(eta.ka) |  V(eta.v) |
|.....| V(eta.cl) |    add.sd |...........|...........|...........|
|-----+-----------+-----------+-----------+-----------+-----------|
|    1|    0.3159 |     1.192 |     3.593 |    0.5700 |   0.09500 |
|.....|    0.3819 |     3.331 |...........|...........|...........|
|    X|     1.371 |     3.294 |     36.33 |    0.5700 |   0.09500 |
|.....|    0.3819 |     3.331 |...........|...........|...........|
```

## Test plan

- [x] `R CMD INSTALL` builds clean.
- [x] Existing `test-control.R` (43 tests covering foceiControl / saemControl / nlmeControl / nlmControl / lbfgsb3cControl / saemControl int covMethod) passes.
- [x] Saem and focei smoke fits on a theo-style model verify `#`/`U`/`X` / gradient row content and the periodic header re-emit cadence.
- [x] `print = 0` produces zero stdout on every method (no native trace leaks).
- [x] Round-trip: `do.call(foceiControl, foceiControl(print = iterPrintControl(every = 5)))` returns the same object.
- [x] Deprecated arg names (`printHeader`, `printUseColor`) error cleanly with R's standard "unused argument" message.

## Follow-up (not in this PR)

A separate task chip is filed to unify the **final-fit-summary back-transformation** path (the `Back-transformed(95%CI)` column users see in `print(fit)`).  That code in `src/inner.cpp:6620+` is a second consumer of the same `logThetasF` / `logitThetasF` / `logitThetasLowF` / etc. env fields and has its own inline `exp`/`expit`/`probitInv` switch.  Pulling it through the same shared `xPar`-encoded back-transform applicator would let us drop the env fields entirely.  Scoped as a follow-up because the parFixed table is independent from iteration printing.

Fix #650


R/saemControl.R
```diff
@@ -141,6 +155,9 @@ saemControl <- function(seed = 99,
                         nmc = 3,
                         nu = c(2, 2, 2),
                         print = 1,
+                        printNcol = floor((getOption("width") - 23) / 12),
+                        printHeader = 10L,
+                        useColor = crayon::has_color(),
```
billdenney: Rename useColor to printUseColor (apply this change throughout)

src/saem.cpp
```diff
@@ -1797,13 +1840,12 @@ class SAEM {
...
-        }
-        RSprintf("\n");
-      }
+      // saem has no per-iteration objective function; NA_REAL renders as "nan"
+      // in the OFV column.  scalePrintFun increments its own counter and gates
+      // the print on (cn % print == 0); with default print=1 every iteration
+      // prints exactly like before, plus the unified focei-style column
+      // wrapping/header/colors.
+      scalePrintFun(&scale, pl.memptr(), NA_REAL);
       Rcpp::checkUserInterrupt();
```
billdenney: Should Rcpp::checkUserInterrupt(); move inside of scalePrintFun since it should likely be universally checked at each printing?

billdenney:
@mattfidler, this is now ready for review.

mattfidler:
This needs a follow up on `babelmixr2` because it breaks how the control structures are made.  `foceiControl()` is what everything turns to for this.

mattfidler:
Really it probably should work just fine, but it may be better to use the new interface if it is available....

billdenney:
I added a placeholder PR in babelmixr2 which relies on some of the mechanics here.

billdenney:
@mattfidler, this is ready for review.

mattfidler:
Looks fine to me.  Should printing a header be linked to the number of iterations printed?

I will merge since I need to change the saem parameter history in the mixture models

PR #653, opened 2026-06-08
Fix nlmSetup NULL-pointer segfault on fresh R session

### LINKED ISSUE #652: nlmSetup segfaults on first estimator call in a fresh R session (bobyqa/nlm/optim/nls/nlminb/lbfgsb3c/n1qn1/newuoa/uobyqa)
## Reproducer

In a fresh R session (no prior `rxode2::rxSolve`, no prior focei fit), every pooled estimator crashes with SIGSEGV during `.nlmSetupEnv`:

```r
library(nlmixr2est)

mod <- function() {
  ini({
    tka <- 0.45
    tcl <- log(2.7)
    tv <- 3.45
    add.sd <- 0.7
  })
  model({
    ka <- exp(tka)
    cl <- exp(tcl)
    v <- exp(tv)
    linCmt() ~ add(add.sd)
  })
}

nlmixr2(mod, theo_sd, est = "bobyqa")  # *** caught segfault ***
```

Identical crash with `est = "nlm"`, `"optim"`, `"nls"`, `"nlminb"`, `"lbfgsb3c"`, `"n1qn1"`, `"newuoa"`, `"uobyqa"`. `focei` works on the same model — it uses its own setup path (`foceiSetup_`).

## Traceback

```
 *** caught segfault ***
address 0x10, cause 'memory not mapped'

Traceback:
 1: .nlmSetupEnv(.p, ui, dataSav, .mi, .ctl,
                 lower = ui$optimParLower, upper = ui$optimParUpper)
 2: .bobyqaFitModel(.ui, .ret$dataSav)
 ...
 8: nlmixr2Est.bobyqa(env, ...)
```

## Root cause

`src/nlm.cpp:146`:
```cpp
nlmOp.stickyRecalcN2Per.assign((size_t)getRxNsub(rx), 0);
```

is called **before** `rxode2::rxSolve_()` initializes the global `rx` pointer (which happens at line 172 with `rx = getRxSolve_();`). On the first call of a fresh R session, `rx` is still NULL and `getRxNsub(NULL)` reads through a near-NULL pointer — hence the `address 0x10` (offset into the struct), "memory not mapped".

gdb confirms:
```
Thread 1 "R" received signal SIGSEGV
#0  getRxNsub (rx=0x0) at rx2api.c:218
#1  nlmSetup (e=...) at nlm.cpp:146
```

Introduced in ef3fc9b3 ("Make parallel inner-retry loop deterministic — per-subject sticky/badSolve"). The previous shared `std::atomic<int>` didn't need to know `nsub`; the per-subject `std::vector<int>` does, but was sized too early.

Verified pre-existing on `origin/main` at `6646e1c8` ("Update CRAN comments").

## Why it isn't always visible

If a focei fit (or any rxode2 solve) has run earlier in the session, the global `rx` is already set and the bug is latent. The crash only manifests on the *very first* nlm-family call of a fresh R session — which is exactly what a copy-pasted reprex from documentation triggers.

## Fix in #653

billdenney:
## Summary

- `src/nlm.cpp:146` sized `nlmOp.stickyRecalcN2Per` via `getRxNsub(rx)` *before* `rxSolve_()` initialized the global `rx` pointer. On any fresh R session — before a prior rxode2 solve / focei fit has populated `rx` — this dereferenced NULL and segfaulted the R process.
- Affected every pooled estimator that goes through `.nlmSetupEnv`: `bobyqa`, `nlm`, `optim`, `nls`, `nlminb`, `lbfgsb3c`, `n1qn1`, `newuoa`, `uobyqa`. `focei` was unaffected (separate setup path).
- Fix: move the per-subject counter sizing to right after `rx = getRxSolve_();`, with a comment at the old call site so a future refactor doesn't regress it.

Fixes #652

Introduced in ef3fc9b3 ("Make parallel inner-retry loop deterministic …") — the shared atomic counter that was replaced didn't need `nsub`, but the per-subject `std::vector<int>` does.

## Test plan

- [x] Reprex from the issue now succeeds for all 9 affected estimators (`bobyqa`, `nlm`, `optim`, `nls`, `nlminb`, `lbfgsb3c`, `n1qn1`, `newuoa`, `uobyqa`)
- [x] `focei` smoke test still passes
- [x] Added `tests/testthat/test-nlmsetup-fresh-rx.R` — uses `stats::nlm` (no `minqa`/`callr` dependency) with `iterlim = 1L` so the test is cheap; the crash is in setup, not in the optimization, so a single iteration is enough to exercise the regression path
- [x] Verified the new test crashes the R process on pre-fix code and passes on the fix
- [ ] `R CMD check` clean (please verify in CI)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,6 +62,7 @@ Imports:
     utils
 Suggests:
     broom.mixed,
+    callr,
     crayon,
     data.table,
     devtools,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,9 @@
 # nlmixr2est 6.0.1
 
 - Fix segfault in `nlmSetup` on the first estimator call of a fresh R
-  session.  The per-subject inner-retry counter was sized via
-  `getRxNsub(rx)` before `rxSolve_()` initialized the global `rx`
-  pointer, dereferencing NULL.  Affected every pooled estimator
+  session affecting every pooled estimator except `nls`
   (`bobyqa`, `nlm`, `optim`, `nls`, `nlminb`, `lbfgsb3c`, `n1qn1`,
-  `newuoa`, `uobyqa`); `focei` used a different setup path and was
-  unaffected.
+  `newuoa`, `uobyqa`);
 
 - Fix LTO violation as requested by CRAN by adding
   -DARMA_DONT_USE_OPENMP to PKG_CXXFLAGS in src/Makevars.in
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # nlmixr2est 6.0.1
 
+- Fix segfault in `nlmSetup` on the first estimator call of a fresh R
+  session.  The per-subject inner-retry counter was sized via
+  `getRxNsub(rx)` before `rxSolve_()` initialized the global `rx`
+  pointer, dereferencing NULL.  Affected every pooled estimator
+  (`bobyqa`, `nlm`, `optim`, `nls`, `nlminb`, `lbfgsb3c`, `n1qn1`,
+  `newuoa`, `uobyqa`); `focei` used a different setup path and was
+  unaffected.
+
 - Fix LTO violation as requested by CRAN by adding
   -DARMA_DONT_USE_OPENMP to PKG_CXXFLAGS in src/Makevars.in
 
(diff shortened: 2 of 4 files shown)


mattfidler:
This needs to actually be run in a fresh R session to verify that this doesn't regress.  Instead the test runs in the current R session which will never fail and show this to be an error if it ever regresses.  The test needs to be fixed.

billdenney:
This should now work with the addition of callr to the suggested package list to start the fresh session.

PR #654, opened 2026-06-08
Unify final-fit-summary back-transform with the iter-print X row

billdenney:
## Summary

PR #651 unified iteration-time print formatting across every estimator via
the shared `scalePrintFun` in `src/scale.h`, but the final-fit-summary block
that builds the `parFixed Back-transformed(95%CI)` column at
`src/inner.cpp:6620+` still had its own switch ladder doing the same log /
logit / probit back-transforms. This PR collapses both consumers onto one
shared C helper.

- New `scaleBackTransform(est, xParCode, probitCode, ...)` in `src/scale.h`:
  pure function that applies log / logit / probit (or identity), driven by
  the same xPar/probitCode encoding both paths already understand.
- `scalePrintFun` X row (every estimator), focei's `parHistData` X row, and
  the final-summary loop all call it.
- Probit is carried as a parallel `probitIdx` + `probitThetaLow/Hi` triple
  on `scaling`, not folded into xPar — positive xPar codes 2–5 are already
  reserved for omega scaling in `scaleGetScaleC`. `scaleSetup` inits the
  new fields to NULL, so existing callers that don't wire probit through
  scale keep working unchanged.
- Final-summary loop builds per-theta `thetaXPar` / `thetaProbitIdx` once
  from the existing env-side parallel arrays (`logThetasF` / `logitThetasF`
  / `probitThetasF` + bounds) and the four-branch switch collapses to one
  body.

## Test plan

- [x] `R CMD INSTALL` builds clean.
- [x] `print(fit)` parFixed Back-transformed(95%CI) column is byte-identical
  to `origin/unify-iter-print` for a focei fit on a model with both
  `log()` and `logit()` thetas — verified by diffing the parFixed data
  frame formatted to 12 digits.
- [x] Same byte-identical check for saem on the same model.
- [x] `testthat::test_dir(filter="^parFixedDf$|^bad-print$|^integerish-focei-print$")` → 4/4 pass.
- [x] Broader test suite shows the same pre-existing failures
  (`.testBoundedTransform`, `.getCorPrint`, `rxode2hasLlik` — environmental
  rxode2/nlmixr2 helpers not in the test load path, not introduced here).

## Deferred

Task description's optional "drop the F-suffix env fields entirely" — the
`logThetasF` env field has a *second* consumer at `src/inner.cpp:7119+`
(`foceiSetup_`) that builds `op_focei.xPar`. That path has its own dead
branch reading `e["logitThetas"]` (no F suffix; never set in R), so
`op_focei.xPar` never receives the negative logit codes today. Fixing
that surfaces a behavior change in focei's iteration X row for logit
thetas, which is out of scope for a byte-identical migration. The
F-suffix env fields stay as inert transport for now.


billdenney:
This depends on #651 being merged first.

billdenney:
@mattfidler, this is now ready for review.

PR #670, opened 2026-06-13
Fix CI: parallel-test compile races (6h timeouts) and focei-default test failures

billdenney:
## What was broken

The R-CMD-check matrix had three independent failures:

1. **Codoc WARNING** — `foceiControl.Rd` documented `sigdig = 3` but the code
   is `sigdig = 4` (from "Change focei defaults"). Under `error-on: warning`
   this failed every Linux job that finished the check.
2. **Three real test failures** — `test-rxuideparse.R:3,9,12` and
   `test-parFixedDf.R:31`, both fallout from the same "Change focei defaults"
   commit (default `outerOpt` `nlminb`→`lbfgsb3c`, `sigdig` `3`→`4`).
3. **6h timeouts (Windows/macOS) and exit-143 OOM (Linux)** — after
   `Config/testthat/parallel` was re-enabled. The suite isn't slow by nature
   (~9 min on Linux); the workers all resolve rxode2's model-compile dir to the
   **same shared path** (`R_user_dir("rxode2","cache")`), so concurrent
   first-compiles raced and corrupted each other → `error building model` /
   `something went wrong in compilation`, runaway retries, memory blow-up.

## Fixes

**Commit 1 — focei default fallout**
- Regenerate `man/foceiControl.Rd` (roxygen2 8.0.0) so the docs match
  `sigdig = 4` (clears the codoc WARNING; `devtools::document()` confirms this
  is the *only* doc change).
- `rxUiDeparse.foceiControl()` omitted `outerOpt` only when it equalled the
  hardcoded `"nlminb"`; compare against the *actual* default
  (`.ret$outerOptTxt`) so the new `lbfgsb3c` default (and any future one) is
  omitted. Fixes `test-rxuideparse.R`.
- `test-parFixedDf.R`: `parFixed` prints the SD to the control's `sigdig`
  significant figures (now 4); assert against `signif(., foceiControl()$sigdig)`
  instead of `round(., 2)`.
- Document the changed `foceiControl()` defaults in `NEWS.md`.

**Commit 2 — parallel test robustness**
- New `tests/testthat/helper-aaa-threads.R` (sorts first → runs before any
  fixture/test in every worker and the main process): point `rxTempDir` at a
  **per-process** directory (resetting the cached `.rxTempDir0` so it takes
  effect) and cap within-solve threads. Eliminates the cross-worker compile
  races.
- `tests/testthat.R`: size the worker pool with `detectCores()` (capped at 2
  on CRAN) via `Ncpus`/`TESTTHAT_CPUS`, and cap main-process threads.

## Verification
- **Linux**, full suite, parallel (10 workers): **0 failures, 8.6 min** (was 6h
  timeout / OOM).
- **CRAN path** (`NOT_CRAN` unset, ≤2 cores): **0 failures, 106 s** (nmTest
  blocks skip as designed; well under the 10-min budget).
- **Windows** (Server 2022, R 4.6.0 + Rtools45, 8 cores): suite now runs to
  completion instead of hitting 6h. *(verification run in progress)*

## Notes
- The old `test-focei-cens.R:53` SAEM-covariance error did not reproduce on
  current `main` before or after these changes.
- rxode2's shared compile dir not being concurrency-safe is the underlying
  cause of the races; the helper isolates it per worker for the test suite, but
  hardening rxode2's cache for concurrent compilation would be a good
  follow-up.


billdenney:
@mattfidler , The part of this that I think would benefit from a different fix is the parallel compilation fix. I think it would be best if multiple models could compile in parallel, but this sidesteps that by giving different directories.

R/foceiControl.R
```diff
@@ -1477,7 +1477,7 @@ foceiControl <- function(sigdig = 4, #
   if (object$outerOpt == -1L && object$outerOptTxt == "custom") {
     warning("functions for `outerOpt` cannot be deparsed, reset to default",
             call.=FALSE)
-  } else if (!(object$outerOptTxt %in% c("nlminb", "stats::optimize"))) {
+  } else if (!(object$outerOptTxt %in% c(.ret$outerOptTxt, "stats::optimize"))) {
```
billdenney, line 1610: @mattfidler, it's not clear to me why this would be dependent on nlminb vs something else. So, this could be wrong. But, it was causing a test to fail.

R/timing.R
```diff
@@ -48,12 +48,10 @@
   if (is.environment(nlmixr2global$currentTimingEnvironment) &
         inherits(nlmixr2global$nlmixr2Time, "proc_time")) {
     .time <- .nlmixrMergeTimeWithExtraTime(get("time", envir=nlmixr2global$currentTimingEnvironment))
-    # Keep non-zero times
-    .keep <- vapply(seq_along(names(.time)),
-                    function(i){
-                      !(.time[[i]] < 5e-5)
-                    }, logical(1), USE.NAMES=FALSE)
-    .time <- .time[, .keep]
+    # Report every recorded stage, including ~0-duration ones, so the set of
```
billdenney, line 51: This was causing steps that ran very quickly to cause tests to fail. It may be simpler to do filtering at a different stage so that the recorded timing is consistent. (Filter on output?)

tests/testthat/helper-aaa-threads.R
```diff
@@ -0,0 +1,28 @@
...
+local({
+  try(rxode2::setRxThreads(2L), silent = TRUE)
+  try(data.table::setDTthreads(2L), silent = TRUE)
+  .dir <- file.path(tempdir(), paste0("rxcompile-", Sys.getpid()))
+  dir.create(.dir, showWarnings = FALSE, recursive = TRUE)
+  Sys.setenv(rxTempDir = .dir)
+  # rxode2 caches the resolved temp dir during .onLoad; reset it so the
+  # per-process directory above is picked up.  Guarded: if a future rxode2
+  # changes this internal, we simply fall back to the default behaviour.
+  try(utils::assignInNamespace(".rxTempDir0", NULL, ns = "rxode2"), silent = TRUE)
```
billdenney: This was added due to compilation collisions causing problems. I think that a better fix would be if there were definitively no collisions. I am not sure if the collisions happened during the original helper setup of all the models running simultaneously (I think yes) in which case we could possibly move those to occur in some way by just one of the test workers, perhaps. If it is for other models, that may be a bigger issue we should try to solve.

mattfidler:
Hi @billdenney 

Another solution is to have different environmental variables that test different things.  focei long running and saem long running making it split into more tests, but be able to test the whole project.

I am thinking thi may have to be what is implemented.  What confuses me is that these used to work before and for some reason they have slowed down even though many speed enhancements have been introduced into the core of both nlmixr2est and rxode2.

R/foceiControl.R
mattfidler: I don't recall. It may now be wrong.

R/timing.R
mattfidler: This is fine with me.

tests/testthat/helper-aaa-threads.R
mattfidler: There is lock/unlock mechanisms for building to allow parallel running, it may work...?

The CRAN machine may not like this approach though, I simply cannot recall.

billdenney:
@mattfidler, What Claude is telling me about the slowness is that it is likely due to resource contention with either too many threads (GitHub runners have 4 cores or more likely hyperthreads with 16 GB RAM). We may want to scale back the parallelism a bit so that there is less resource contention. Would it be helpful for me to do a test running speed bisection of nlmixr2est/rxode2 locally to see if something pops up?

mattfidler:
Maybe. Sometimes it is super helpful sometimes it doesn't help at all.

As a note I will away for 2 weeks.  I can't really review on a phone...

billdenney:
@mattfidler , I had Claude replicate the build infrastructure as close as possible locally, and it seems like the underlying issue is OpenBLAS causing memory/threading issues. I think that if the current CI succeeds, that likely proves the issue. It's not clear to me if the CRAN fixes are right, but we may want to simplify it to always test single-threaded on CRAN.

mattfidler:
Still not sure where this is falling down.

billdenney:
I think it may have been the rxode2 memory issues. I'll rerun to see if it works now that the rxode2 memory fix is merged.

PR #682, opened 2026-06-15
Fix infinite recursion in .cloneEnv() on circular/shared environment references

billdenney:
## Problem
`.cloneEnv()` (the deep-cloner behind `nlmixrClone()`) recurses into every
environment-valued binding with no record of what it has already visited. On an
environment graph that contains a **cycle** or a **shared environment**, it
recurses forever and aborts with:

```
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
```

This surfaced when building output for **`est="saemix"`** (babelmixr2): the fit's
`meta$.simModelBase` carries an rxode2 model whose compiled-model cache is
self-referential (`.rx` ↔ `.rxModels` ↔ `rx_<hash>__$.rx`). Cloning the fit hit
that cycle and the fit silently came back `NULL` (the stack-overflow error was
swallowed by the cache-reset error collector), so `est="saemix"` produced no
usable fit.

## Fix
Track the environments already being cloned by `data.table::address()` (the same
approach `rxode2`'s `.udfAddToSearch()` already uses) and return the in-progress
clone on revisit. This:
- terminates on cyclic/shared graphs,
- preserves the shared reference / cycle in the cloned graph (rather than
  duplicating or diverging), and
- avoids needlessly deep-copying a cache of C-compiled models.

~6 lines; environments are kept (no architectural change).

## Test
Adds `tests/testthat/test-cloneEnv-cycle.R` (two-node cycle + direct
self-reference). Verified TDD-style in an isolated container:
- **Before** this change: `FAIL=1` — the test aborts with
  `evaluation nested too deeply: infinite recursion`.
- **After**: `FAIL=0 PASS=9`.

End-to-end, the previously-failing babelmixr2 `est="saemix"` comparison tests now
pass (`FAIL=0`).

## Notes
General robustness fix — any deep-clone of a fit whose simulation-model cache is
populated could hit this, not only `est="saemix"`.


mattfidler:
I think this duplicates some code in rxode2, but this is fine.

PR #668, opened 2026-06-12
Fix covMethod="r" and "s" standard errors (constant-factor inflation)

hiddevandebeek:
## Summary

`covMethod="r"` standard errors are √2 too large and `covMethod="s"` SEs are 2× too large. The default sandwich `"r,s"` was already correct. Fixes #666.

## Root cause

With `OFV = −2·logL`, the R matrix is `R = ½·Hessian(OFV) = I_obs` (observed information) and the S matrix is the score cross-product `S = J`. The correct covariances vs what the code returned:

| method | correct | returned |
|---|---|---|
| R | `R⁻¹` | `2·R⁻¹` |
| S | `S⁻¹` | `4·S⁻¹` |
| R,S | `R⁻¹SR⁻¹` | `R⁻¹SR⁻¹` ✓ |

The sandwich never used `covR`/`covS`, so only `"r"`/`"s"` were affected.

## Fix

In `foceiCalcCov`: `covR = Rinv` (was `2*Rinv`) and `covS = Sinv` (was `4*Sinv`). These also feed the sandwich-selection heuristic, so fixing them at source keeps that comparison on consistently-scaled covariances. `covRS` is unchanged.

## Validation

**NONMEM `$COV MATRIX=R`** (combined-error sim), mean nlmixr2/NONMEM SE ratio: `"r"` 1.345 → **0.951**; `"r,s"` 0.958 (unchanged).

**Empirical sampling covariance of a known data-generating model** (mean method SE / true `cov(θ̂)`, scale-free relative ratios): `r ≈ r,s ≈ s(S⁻¹)`, while the old `s(4·S⁻¹)` runs at **2.04× / 2.09× / 2.05×** for tka/tcl/tv. So `"s"` goes from ~2.0× the true SE to ~1.0×.

(NONMEM `MATRIX=S` is not used as the reference — it runs ~1.5× its own `MATRIX=R` here, so it would mislead.)

## Scope

Default `"r,s"` and `covRS` are unchanged; the selection heuristic and the `$covR`/`$covS` diagnostics now operate on correctly-scaled covariances. Affects the explicit `covMethod="r"`/`"s"` outputs and the R/S fallbacks.

Issue #666, opened 2026-06-12
covMethod="r" standard errors look ~sqrt(2) larger than NONMEM $COVARIANCE MATRIX=R


hiddevandebeek:
I was prototyping a faster covariance step and, to sanity-check it, started from the existing `covMethod="r"` — where I noticed `"r"` itself doesn't line up with NONMEM.

### Background (the faster method, just for context)

It builds the structural-θ information analytically from the FO/FOCE per-subject marginal covariance `C_i = a_i·Ω·a_iᵀ + R_i`, as `I = Σ_i S_iᵀ·C_i⁻¹·S_i` (where `a_i = ∂f_i/∂η`, `S_i = ∂f_i/∂θ`), reusing the inner sensitivities estimation already computes — no finite differences, so it's a lot faster. Before wiring it up I compared against `covMethod="r"`, and that's where the discrepancy showed up.

### Observation

On `theo_sd`, FOCEI, the same model/data fit in nlmixr2 and NONMEM (OBJF 116.806 vs 116.803):

```r
library(nlmixr2)
one.cmt <- function() {
  ini({ tka <- log(1.5); tcl <- log(2.7); tv <- log(31.5)
    eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1; add.sd <- 0.7 })
  model({ ka <- exp(tka+eta.ka); cl <- exp(tcl+eta.cl); v <- exp(tv+eta.v)
    linCmt() ~ add(add.sd) })
}
fit_r  <- nlmixr2(one.cmt, theo_sd, "focei", foceiControl(covMethod = "r"))
fit_rs <- nlmixr2(one.cmt, theo_sd, "focei", foceiControl(covMethod = "r,s"))
p <- c("tka","tcl","tv")
data.frame(SE_r  = fit_r$parFixedDf[p,"SE"],
           SE_rs = fit_rs$parFixedDf[p,"SE"],
           ratio = fit_r$parFixedDf[p,"SE"] / fit_rs$parFixedDf[p,"SE"])
```

Output (nlmixr2est 6.0.1, rxode2 5.1.3):

```
      SE_r     SE_rs    ratio
tka 0.26877  0.19504  1.378
tcl 0.11735  0.07508  1.563
tv  0.06544  0.04365  1.499
```

[...]

PR #691, opened 2026-06-23
FOCEI/FOCE covariance: scaling fix (#666) + analytic observed information + FD fallback

hiddevandebeek:
Fixes and extends the FOCEI/FOCE covariance, addressing two distinct defects in the legacy finite-difference path and adding an exact analytic path. (This supersedes #668, whose √2/×2 scaling fix is included here.)

## 1. Scaling fix for the legacy FD covariance (was #668, #666)

`covMethod = "r"` and `"s"` standard errors were inflated by constant factors (`sqrt(2)` and `2`). With the objective on the `-2*logLik` scale, the R matrix is the observed information and the S matrix is the score cross-product, so the covariances are `R^-1` and `S^-1` -- but they were returned as `2*R^-1` and `4*S^-1`. Fixed at source (`covR`/`covS`), so both the final output **and** the sandwich-selection heuristic compare consistently-scaled covariances. The default sandwich `"r,s"` was already correct and is unchanged.

## 2. Analytic observed-information covariance (new)

The observed information is the exact-gradient construction (Almquist, Leander & Jirstrand 2015) **continued by one differentiation**. Writing the FOCE/FOCEI objective as a **data term** plus the **log-determinant of the first-order inner Hessian**, the population Hessian splits into:

- the **data term** -- reuses the second-order sensitivities already assembled for the gradient;
- the **log-determinant term** -- carries the **third-order** sensitivities, entering in exactly two places.

This is finite-difference-free: no step to tune, and none of the catastrophic-cancellation blow-ups of a differenced Hessian. New functions:

- **`foceiCovAnalytic(fit)`** -- the analytic R matrix for the structural thetas. Runs on a **released** rxode2 via a bundled pure-R third-order sensitivity generator (rxode2's C++ `rxExpandSens3_`, nlmixr2/rxode2#1092, is used when present).
- **`foceiCovFD(fit)`** -- the **full** covariance (thetas + Omega + residual) as the inverse FD Hessian over the natural **variance-covariance (non-Cholesky)** Omega parameterization, producing the Omega/residual SEs that `covMethod="r"` does not.
- **`foceiCov(fit, omega = FALSE, residual = FALSE)`** -- orchestrator.

## 3. Finite-difference fallback when sensitivities don't load

`foceiCov()` uses a three-step fallback so a covariance is **always** produced:

1. analytic structural-theta block (needs 2nd/3rd-order sensitivities);
2. → `foceiCovFD` (only 1st-order sensitivities, so it survives where the large augmented solve fails);
3. → the fit's **built-in finite-difference covariance** (`fit$cov`) when the sensitivity machinery does not load at all (no symengine, or rxode2 cannot build the sensitivity model).

The analytic path is never partially applied, and the FD it falls back to is the now-correctly-scaled covMethod result from part 1.

## Validation

Analytic SEs match NONMEM `$COVARIANCE MATRIX=R` and ferx to ~1% across theophylline (additive), warfarin (proportional), and a simulated combined-error model, for **both FOCEI and FOCE**; the full FD path reproduces ferx's Omega/residual SEs to ~1%; the analytic R matrix matches finite differences of the objective to ~1e-7. Tests in `test-cov-analytic.R` and `test-cov-focei.R`. Man pages were hand-written; re-run `devtools::document()` in a full build to regenerate.


hiddevandebeek:
I'd like your read on the **theoretical approach** for the analytic FOCEI/FOCE covariance before I commit to porting it to C++.

The idea: instead of finite-differencing the objective for the R matrix, assemble the observed information in closed form. I take Almquist's (2015) exact-gradient construction — the outer gradient `∂F/∂θ` written in closed form from ≤2nd-order sensitivities — and continue it by **one more differentiation** to get the full observed-information matrix. 

Structural thetas enter via mu-referencing (`∂f/∂θ_k = ∂f/∂η_k`); the residual sigma via symbolic partials of the residual variance; and Omega on the natural **variance–covariance** scale (non-Cholesky), using `∂Ω⁻¹/∂ω_ab = −Ω⁻¹ E_ab Ω⁻¹` — so it yields Omega variance *and* block-covariance SEs directly (the ones `covMethod="r"` doesn't give).

For robustness it's a three-tier ladder (the augmented ODE gets likelier to fail the higher its order): 
(1) the exact 3rd-order above; 
(2) a middle tier that differences the analytic **2nd-order** sensitivities with **Shi (2021)** adaptive central differences — the same scheme the inner problem uses — to build the 3rd-order tensor without the heavier ODE, reproducing tier 1 to ~1e-5; 
(3) a last-resort finite-difference of nlmixr2's own objective. Every tier returns the same full parameter set.

I've validated it end-to-end against NONMEM `$COV MATRIX=R`, ferx, and finite differences — agreement ~1%, analytic vs FD ~1e-5 — for diagonal and block Omega and add/prop/combined error. (FOCE here matches nlmixr2's FOCE, i.e. Laplace without interaction.)

**Question: are you happy with this theory — does the construction look correct to you?** It's all in R right now as a reference implementation. If the approach is sound in your view, the plan is to move it to **C++ (into `inner.cpp`, RcppArmadillo)** for the native covMethod and performance.

hiddevandebeek:
@mattfidler so also incorporating the shi already in tier 2, lmk what you think

hiddevandebeek:
Superseded by a consolidated reimplementation rebased on the latest `main`: one minimal PR combining the #666 covariance scaling fix, the analytic observed-information covariance over theta + sigma + Omega (covariates, non-mu-referenced etas, and IOV), the sigdig-derived cov ODE tolerance (#694), and the finite-difference fallback. Closing in favour of that replacement (PR to follow).

PR #697, opened 2026-06-26
focei: covariance over theta + sigma + Omega, tightened cov ODE tolerance (#694)

### LINKED ISSUE #694: Consider tighetr tolerances for SEs
https://github.com/FeRx-NLME/ferx-core/issues/520

@hiddevandebeek if you want to do this too, it relates to your other pull request.

In rxode2/focei this can be accoplished by the sigdig and having different atol/rtol based on the requested ODE estimates.

hiddevandebeek:
## Summary

The FOCEI/FOCE covariance (`$cov`) now spans the **full parameter set** — population thetas, residual error (sigma), and the natural-scale **Omega** variances/covariances — instead of the structural thetas only. This applies to `covMethod = "r"`, `"s"`, and `"r,s"`. Omega/residual standard errors are extractable via `sqrt(diag(fit$cov))`; the printed parameter table is unchanged.

It also fixes the original #694 problem: the covariance finite-difference Hessian was dominated by ODE solver noise at the default `sigdig`, giving unreliable SEs. The cov step now solves the ODE at a tighter (bounded) tolerance than estimation.

## What changed

- **Parameter set** — the cov skip mask now drops only literally-fixed parameters, so `npars = thetan + omegan − fixedn` spans theta + sigma + Omega (incl. IOV). The existing `foceiCalcR`/`foceiS`/`foceiCalcCov` machinery computes over it unchanged (no new covariance routine).
- **Omega scale** — the cov's Omega block comes out on the estimation (Cholesky) parameterization and is delta-transformed to the natural variance-covariance scale: `cov_natural = T·cov·Tᵀ`, with `J[k,c] = ∂Ω_pairₖ/∂θ_c = −[Ω·∂Ω⁻¹/∂θ_c·Ω]`, using rxode2's `d.omegaInv` (the same primitive the gradient uses). Theta + sigma are untouched. Handles diagonal, diagonal-with-fixed, and full block Omega.
- **Cov ODE tolerance** — an RAII guard re-runs the inner-model `rxSolve_` setup at `covSigdig`-tight atol/rtol (`clamp(sigdig+5, sigdig, 10)`, atol floor ≈5e-13) for the cov step and restores the estimation tolerance after.
- **Bounded params** — the existing post-fit Jacobian back-transform is padded with identity for the new sigma/Omega block.

## Validation

Validated against **NONMEM 7.5** (FOCEI, ADVAN2 TRANS2) on `theo_sd`: estimates/OFV land on the same optimum, and R / S / sandwich SEs agree to **~1–3%** (S-matrix to <1%, including the genuine OPG inflation of the score method). The `test-cov-focei.R` regression test is anchored to those NONMEM reference SEs.

## Notes

- `parFixed` (the printed table) is unchanged — Omega/residual SEs live only in `$cov`.
- A *partially*-fixed correlated Omega block (fixing an individual off-diagonal of a block) is left on the estimation scale rather than mis-scaled (documented in-code); diagonal-with-fixed and full free blocks both transform correctly.
- `covSigdig` is currently derived from `sigdig`; a control hook is in place for a future `foceiControl(covSigdig=)` argument.

Closes #694.


hiddevandebeek:
So this adds Omegas, Errors and IOV. question remains whether all are indeed wanted in all scenarios (causing slow-down), or that user can specify.

right now i just tightened the sig digits by 5. I read your comment on the heuristic  and what to choose, I'll think about it for a bit.

Also not sure of naming yet of etas in the $cov block

I've tried to make it a bit accessible so we can adjust it easily.

Let's discuss this next week!

hiddevandebeek:
## `covType="analytic"` — exact FOCEI observed-information covariance (+ `covFull`, #694 cov ODE tol)

Adds an exact, finite-difference-free observed-information covariance for **FOCEI**, plus a `covFull` switch and a tightened cov-step ODE tolerance. Default behaviour is unchanged (`covType="fd"`, `covFull=FALSE`).

### Control knobs (`foceiControl`)

| arg | default | meaning |
|---|---|---|
| `covType` | `"fd"` | `"fd"` = finite-difference Hessian (unchanged); `"analytic"` = exact analytic observed information (FOCEI only) |
| `covFull` | `FALSE` | `FALSE` = `$cov` over structural θ only (original scope); `TRUE` = θ + residual σ + natural-scale Ω |

### How it works

The FOCEI Laplace `-2LL` Hessian splits into a **data term** (2nd-order prediction sensitivities, already needed for the gradient) and a **log-det term** (3rd-order sensitivities, from `∂²log|H̃|/∂θ²`). State sensitivities come from rxode2 `.rxSens`; Ω derivatives from rxode2 `rxOmegaVarCovDeriv`.

**Uniform direction assembly** (the core idea): each structural θ contributes one sensitivity direction — a mu-referenced θ reuses its eta's direction (`∂/∂θ ≡ ∂/∂η`); every other θ (covariate coefficient *or* eta-less) gets its own `THETA_j_` direction. No covariate detection or scaling.

Pipeline: `.foceiCov` → `.foceiCovAnalytic` (loop subjects, sum R, invert) → installed as `$cov`.

### New file `R/foceiCovAnalytic.R`

| function | role |
|---|---|
| `.hasRxExpandSens2` | availability gate (needs rxode2 `rxExpandSens2_` + `symengine`) |
| `.omegaVarCovDeriv` | thin wrapper over `rxOmegaVarCovDeriv` (Ω⁻¹ & log\|Ω\| 1st/2nd derivs) |
| `.foceiAnalyticErrFull` | symbolic f-derivatives of the error model ρ/p for add / prop / combined1 / combined2 |
| `.foceiAnalyticAugModelDirs` | builds the augmented rxode2 sensitivity model over the eta+θ direction set (order 2 or 3) |
| `.foceiAnalyticSolveDir` | per-subject solve of the **order-3** model → f, a, A, Ath (exact3 tier) |
| `.foceiAnalyticSolveDirFD3` | **order-2** solve + Shi(2021) FD of the exact 2nd-order sens for the 3rd-order term (fd2 tier) |
| `.shiHarmonicMean` / `.shiRC` / `.shi21CentralR` | R port of nlmixr2's Shi(2021) adaptive central-difference step (the C++ `shi21Central` isn't R-callable) |
| `.foceiAnalyticSubjectR` | assembles one subject's observed-information R-matrix (data + log-det; θ/σ/Ω blocks) |
| `.foceiCovAnalytic` | top level: assemble directions, loop subjects, sum R, invert → cov/se/params |

### New file `R/foceiCov.R`

| function | role |
|---|---|
| `.foceiCov` | **two-tier ladder**: `fd2` (default) then `exact3` (fallback); returns the same full param set from either |
| `.omegaBlocks` | connected components of Ω (block detection) |
| `.omegaFixed` | which Ω element pairs are fixed |
| `.foceiEtaThetaMap`\* | eta rows + their mu-ref θ (shared helper) |
| `.iniIsFixed`\* | which named params are fixed (shared helper) |
| `.foceiOmegaCovNames`\* | `om.<x>` / `cov.<x>.<y>` naming (shared helper) |

\* extracted in the audit pass to de-duplicate logic shared with `focei.R`.

**Two tiers** — same direction-set model, same assembly, differ only in the 3rd-order tensor:

| tier | 3rd order via | build cost | accuracy |
|---|---|---|---|
| `fd2` (default) | Shi(2021) FD of *exact* 2nd-order sens | ~4 s (2nd-order model) | matches `exact3` ~1e-3 |
| `exact3` (fallback) | exact analytic 3rd-order sens | ~40 s (~80 KB model) | exact; used only if fd2 non-finite |

### Changes to existing files

| file | change |
|---|---|
| `R/focei.R` | `.foceiAnalyticInScope` (pre-fit gate) · `.foceiAnalyticCovOverride` (post-fit: install analytic cov or announce FD) · `.foceiAnalyticFillParFixed` + `.scaleBackTransformR` (rebuild the SE/%RSE/CI table, since the C++ cov step is skipped) · `.foceiNameOmegaCov` (name the Ω block of the FD `$cov` under `covFull`) · two hooks in `.foceiFamilyControl` (skip FD cov when analytic is in scope) and `.foceiFamilyReturn` (run the override) |
| `R/foceiControl.R` | add `covType` and `covFull` params |
| `src/inner.cpp` | `covSolveArgs_`/`storeCovSolveArgs_` (cache solve args) + `CovSolveTolGuard` (RAII: tighten cov-step ODE tol to `0.5·10^(−covSigdig−2)`, restore on exit — **#694**) · `covFull` natural-scale Ω transform in `foceiCalcCov` (delta-method off the Cholesky scale) |

### Gating — **FOCEI only**

| condition | in scope? |
|---|---|
| FOCEI + add/prop/combined error, ≥1 eta, sens primitives present | ✅ analytic |
| FOCE (`interaction==0`) | ❌ → FD (see below) |
| FO / FOI | ❌ → FD (first-order marginal, not conditional) |
| `lnorm` / `propT` / `propF` error | ❌ → FD |
| fixed structural θ (breaks eta indexing) | ❌ → FD |
| augmented ODE fails to integrate | ❌ → next tier, else warn / FD |

**Why FOCEI only:** FOCE's marginal objective is *non-smooth in θ* — its inner EBE solve is fed an inconsistent (value, gradient) pair (the FOCE gradient drops the `∂R/∂η` terms but the value keeps true `R(f(η))`), so `η̂(θ)` jitters deterministically and the observed-information surface is jagged. This is inherent to the inner problem (analytic sensitivities don't fix it), and is why the *FD* FOCE R-matrix is also unreliable — tracked separately under #694. FOCEI has a genuine smooth objective, so the engine is scoped to it.

### Edge cases (all covered + tested)

| case | handling |
|---|---|
| non-mu-referenced eta | keeps its own `ETA_i_` direction; Ω named by the eta |
| covariate coefficient / eta-less θ | own `THETA_j_` direction (true sensitivity) — no heuristic |
| block Ω | connected-component detection; off-diagonal cov SEs |
| single random effect | handled (no `sapply` collapse) |
| fixed σ | dropped from the σ block |
| `combined1` vs `combined2` | `(sa+sp·f)²` vs `sa²+sp²·f²` |
| non-PD information | `NaN` SEs flag it (no silent `nearPD`) |
| per-subject solve failure | guarded → next tier → FD |

### Validation

| check | result |
|---|---|
| `fd2` vs `exact3` (full cov) | ~1e-3 |
| structural-θ SEs vs NONMEM `$COV MATRIX=R` (`theo_sd`) | `0.18868 / 0.08351 / 0.04617` |
| covariate / eta-less / block-Ω vs FD full cov | tol 5e-3 |
| FOCE bows out | asserted |
| tests | `test-cov-analytic.R` (54) · `test-cov-focei.R` (`r`/`s`/`r,s` vs NONMEM, #666) |

hiddevandebeek:
So my thoughts are:

fd2 (fd'ing over 2nd order) is so much quicker than 3rd order, it might not be worth doing 3rd order for now --> perhaps later? it uses the Shi harmonic means, but since its fd over analytical its very precise.

The logic I use now doesnt make sense (fallback of 2nd order compilation fail tries 3rd order), but I sort of kept it since I was working on it. Ill remove this, right?

FOCE was not acting correctly for FD; also after some trying, so i left that out for now. NONMEM using SB likelihood for FOCE, nlmixr2 is a bit different including the log|Htilde|, but I guess on purpose?

2 new arguments in foceiControl, but by default the same as it used to be, so backwards-compatible.
1. minimal SEs (thetas only) or full matrix (only in matrix; not in print function)
2. fd or analytical covariance R matrix

@mattfidler let me know what to clean up, remove or something! Ill be adjusting the code to ensure everything works as promised over the course of the coming days. But this is the core idea.

hiddevandebeek:
I dont like this implementation, opting for one in the C++ optiimzer

mattfidler:
Ok.  I will wait.  Still not done merging in the branches in nlmixr2est...

hiddevandebeek:
Superseded by a consolidated reimplementation rebased on the latest `main`: one minimal PR combining the #666 covariance scaling fix, the analytic observed-information covariance over theta + sigma + Omega (covariates, non-mu-referenced etas, and IOV), the sigdig-derived cov ODE tolerance (#694), and the finite-difference fallback. Closing in favour of that replacement (PR to follow).

PR #700, opened 2026-07-04
focei/foce: analytic observed-information covariance + FOCE objective fix (+ #666 scaling)

### LINKED ISSUE #666: covMethod="r" standard errors look ~sqrt(2) larger than NONMEM $COVARIANCE MATRIX=R

hiddevandebeek:
## Summary

Three related pieces of FOCEI/FOCE work, rebased on the latest `main`:

1. **`foceiControl(covType = "analytic")`** — an exact analytic observed-information
   covariance for **FOCEI and FOCE** fits, computed while the optimizer is live.
2. **A FOCE objective/EBE fix** — FOCE (`interaction = FALSE`) evaluated its residual
   variance at the wrong point; ODE and `linCmt()` FOCE now agree with each other and
   with the NONMEM FOCE reference.
3. The **#666 covariance scaling fix**.

Consolidates the covariance work from #691 and #697 into one PR. Supersedes #691 and
#697. Fixes #666.

## FOCE objective / EBE fix

FOCE (`interaction = FALSE`) must evaluate the residual variance `R` entering the inner
likelihood at the `eta = 0` population prediction and hold it constant, because the
truncated Sheiner-Beal inner gradient drops `dR/deta`. The previous code froze `R`
symbolically (`rxRepR0_`), which only removed *explicit* `eta` symbols:

- **ODE models** kept a live model state in `R`, so `R` stayed `eta`-dependent — the
  FOCE-ODE objective was wrong (disagreed with the closed form and NONMEM).
- **`linCmt()` models** had a second `linCmt`/`linCmtB` call injected by the
  substitution, corrupting the inner sensitivities and stalling the `eta` optimization.

`R` is now supplied at `eta = 0` directly from the inner model in C++ (`getPopR`), and
the FOCE inner likelihood/gradient use that fixed population `R`. FOCE-ODE now matches
the `linCmt` closed form and both match the NONMEM FOCE reference (Wang 2007). The
`eta = 0` `R` depends only on `theta`, so it is **cached per subject** across the inner
iterations (recomputed once per parameter update), keeping FOCE close to the FOCEI
per-iteration cost. FOCEI/FO are untouched (byte-identical).

## Analytic covariance (covType = "analytic")

The R-matrix (observed information) is assembled in closed form: a data term from the
analytic 2nd-order sensitivities (`rxode2::.rxSens`) plus a log-determinant term whose
3rd-order tensor is recovered by Shi (2021) central differences of those sensitivities
(keeping the augmented ODE at O(ndir^2), not O(ndir^3)). It reports a full covariance in
`fit$cov` over the structural thetas, the residual sigma, and the Omega
variances/covariances — including the Omega/residual SEs the finite-difference `"r"`
matrix does not provide (via `rxode2::rxOmegaVarCovDeriv`).

Parameter coverage, all detected automatically from the model structure:

- mu-referenced thetas — reuse the eta sensitivity direction (no extra ODE cost);
- eta-less thetas and covariate coefficients — their own sensitivity direction;
- non-mu-referenced etas (e.g. `exp(theta)*exp(eta)`) — eta-named Omega;
- IOV (inter-occasion variability) — via the shared-variance Omega branch.

**FOCE** (interaction off) uses the general total-derivative Hessian and re-solves the
FOCE empirical-Bayes estimates to the FOCE inner stationarity at the same `eta = 0` `R`
before assembly, so the analytic cov is the observed information of the corrected FOCE
objective; additive-error FOCE reduces exactly to the FOCEI result.

Any fit outside scope (FO, `nAGQ > 1`, censoring, pure-proportional or DV-transformed
error, bounded-parameter transforms, a structural theta shared by two etas, a non-SD
`iovXform`) warns and falls back to the finite-difference Hessian, identical to
`covType = "fd"`.

## covMethod handling

`covType = "analytic"` defaults to `covMethod = "r"` (the observed-information inverse).
An explicit `covMethod = "r,s"` or `"s"` is honored, with the analytic R feeding the
native finite-difference sandwich / S-matrix.

## New foceiControl arguments

- `covType = c("fd", "analytic")` — default `"fd"` (unchanged behaviour).
- `covSolveTol` — override the `sigdig`-derived augmented-solve tolerance.
- `covFull` — `FALSE` (default) installs the structural-theta block only, matching the
  finite-difference `fit$cov` shape for backwards compatibility; `TRUE` installs the
  full theta+sigma+Omega covariance (identical theta SEs).

## Covariance scaling fix (#666)

The FOCEI `covMethod = "r"` / `"s"` / `"r,s"` standard errors were inflated by a constant
factor (sqrt(2) for `"r"`, 2 for `"s"`) because the R- and S-matrix covariances used
`2*R^-1` / `4*S^-1` instead of `R^-1` / `S^-1`. They now match NONMEM `$COV`. This
changes existing `covType = "fd"` SEs.

## Validation

- Analytic FOCEI theta SEs match NONMEM `$COV MATRIX=R` (theo_sd: 0.18868 / 0.08351 / 0.04617).
- Covariate and eta-less thetas match tight finite differences (~0.5%).
- A non-mu-referenced eta is byte-identical to the equivalent mu-referenced form.
- IOV observed information matches a profile finite difference to ~0.02%.
- FOCE analytic R matches an `eta = 0` gold finite-difference to ~3e-5; FOCEI byte-identical.
- `test-cov-analytic.R` 55/0; `test-focei-wang2007-basic.R` (real FOCEI+FOCE fits) 198/0.

## Reuse

Delegates to existing machinery throughout: `rxode2::.rxSens` / `.rxJacobian` /
`rxStateOde` / `rxFromSE` for sensitivities, the templated `shi21Central` for the
3rd-order tensor, `rxOmegaVarCovDeriv` for the Omega derivatives, `ui$muRefDataFrame`
for mu-referencing, and `nmNearPD` for the PD correction.

## Known follow-up

The extended Wang2007 regression files that are `skip_on_ci` (bounded, boxcox,
boxcox-half, boxcox-lnorm, combined, lognormal, power, yeojohnson) hold hard-coded
FOCE-ODE reference values that the FOCE fix changes. The CI-visible `basic` file is fully
updated, and the shared error-model vectors in boxcox/combined/power are updated to the
verified values; the transform-specific vectors in these `skip_on_ci` files still need a
one-time regeneration on Linux (they do not run on CI, so they do not affect this PR's
checks — Windows deliberately skips them due to DLL overloading).


hiddevandebeek:
Thanks -- addressed all four (pushed in `8811d3f7`); they were all documentation/robustness, no behaviour change:

1. **`covType` roxygen scope** -- corrected: the analytic path applies to additive or combined additive+proportional error and covers mu-referenced, covariate, and other non-mu-referenced structural parameters plus IOV; it falls back for FO/FOCE/`nAGQ>1`/censoring/pure-proportional/DV-transformed error.
2. **NEWS scope** -- fixed the same way (IOV and covariate/non-mu-referenced parameters were incorrectly listed as out-of-scope; pure-proportional noted as the error case that falls back).
3. **`man/foceiControl.Rd`** -- regenerated to match the corrected roxygen.
4. **`foceiCovAnalytic()` theta source** -- the standalone oracle now reads `fit$theta` (converged), falling back to `iniDf$est`, matching the live hook's use of the converged estimates.

Full suite green (42 pass).

mattfidler:
As a note, I will change analytic to be the default since there are other things that are changing and will be breaking changes.

hiddevandebeek:
Worked through the full review (pushed in `573765b0`). All 12 findings addressed; the 5 refuted ones need no action (the two `covFull` items are moot now that `covFull` defaults to `FALSE`).

**Correctness**
1. **DV/EVID==2 misalignment** — restored the `nrow(solve) == length(obs times)` guard (it was a regression from an earlier port); a stray EVID==2/covariate row sharing an obs timestamp now bails to FD instead of silently recycling `y`.
2. **Stale `.analyticCov` after a failed inversion** — `.analyticCov` is now stashed only *after* the deciding `solve(.ct)` succeeds, so a NULL return leaves no stale cov for the FD path to pick up.
3. **Success path never restored the global solve** — `restoreFitSolve_()` now runs on the analytic-success branch too; verified `nrow(fitData)` and OBJF match the FD fit (tables no longer truncated to the last subject).
4. **IOV oracle read a process-global** — the standalone `foceiCovAnalytic()` now derives the IOV flag from the `fit` (occasion `condition` on the eta), not `.uiIovEnv`.
5. **Swallowed restore failure** — `restoreFitSolve_` returns a status; if the re-solve fails, the decline path now reports a failed covariance instead of running FD against a freed solve.
6. **Omega blocks from converged values** — `.omegaBlocks` now derives block structure from the *declared* `iniDf` off-diagonals, so a block whose covariance converges near zero keeps its off-diagonal parameter.

**Cleanup / performance** (all proven numerically identical — R matrix and SEs bit-for-bit)
7. `storeCovSolveArgs_` guarded on `covType=="analytic"` (no per-fit dataset retention on `fd` fits).
8. Pre-split subject data once. 9. Invert `Omega` once, not per subject. 10. Hoisted the index-independent `Mcol`/`Cpe` out of the R-fill loop. 11. Extracted the duplicated Omega-pairs assembly into `.foceiOmegaPairs`. 12. Vectorized the 3rd-order symmetrization with `aperm`.

Added regression tests for #3 (table integrity) and #6 (declared-block). Full suite green (all cov + focei suites, 0 fail).

mattfidler:
@hiddevandebeek as a note, I'm unsure why `foce` is gated, it should use the same machinery.

mattfidler:
Also let me know if you think this is ready.

hiddevandebeek:
i will try to do foce the coming 24 hours, ran into estimation problems earlier with FOCE altogether (for FD and Analytical). I think it has to do with the likelihood, not sure completely yet. If that doesnt work tomorrow I will focus on finalizing this PR. then FOCE will come later.

I want to do one last nonmem run with some models to test robustness.

But so far its almost done, I'm more happy with this implementation since it seams into the existing one.

Tomorrow evening at the latest ill give you an update on what worked and whether we're ready to merge

mattfidler:
Ok.  As a note, a major rewrite of the inner gradients will probably hit before your patch goes live  BUT it is likely not going to be default.  I was hoping after your patch but it seems to be working out that way.

mattfidler:
Also as a note, for the warm starting of the EBEs, it isn't terrible to calculate the hessian (because it is part of the objective function).  `calcEtaHessian()` should be able to do the warm start pretty easily.  tit may be good to have some caching in this method.  It isn't inserted into the Zm right now (it uses the optimizer estimate of the Hessian).  I think it may not be the same structure (row or column major lower triangular; I think it is opposite of rxode2/NONMEM convention if I remember correctly).

hiddevandebeek:
okay its ready for review.
I worked out the bug (I think) for the FOCE optimization. However, its now slower than FOCEI. Did add caching to negate it a bit.
It would be nice if you had a look at whether this was truely a bug and robustly fixed now.

If you feel like its not tested enough/robust enough, otherwise I'll gate FOCE again for the analytical R-cov and rip out the FOCE inner update.

Let me know! 

btw:
For the next nlmixr2 version I think the analytical R-cov is definitely within scope. I'll start working on warm-start EBE and looking at outer gradients too somewhere in the following days, but I don't suspect to have it ready for the next CRAN version.

hiddevandebeek:
## Independent NONMEM validation of the FOCE objective fix (Wang 2007)

To confirm the FOCE fix in this PR against an external reference, I re-ran the
Wang 2007 error-model suite in **NONMEM 7.5.1** (`nmfe75`) and compared the
objective function value against nlmixr2 for **10 error structures × {FOCEI, FOCE}**.

**Result: 20/20 objective values match to 3 decimals.**

### Setup

- Same monoexponential as the test suite (`ke = tke·exp(eta.ke)`,
  `ipre = 10·exp(-ke·t)`), evaluated **at the initial estimates** — nlmixr2
  `foceiControl(maxOuterIterations = 0)` ↔ NONMEM `MAXEVAL = 0`, all parameters
  `FIX`.
- Closed-form `$PRED` (so NONMEM is a genuinely independent implementation, not
  an ODE round-trip), `METHOD=1` with `INTER` (FOCEI) / without (FOCE).
- Each NONMEM error model was translated from nlmixr2's **exact** residual
  variance (`ui$foceiModel0` `rx_r_`), not guessed — e.g. `propF` variance is
  `(f2·prop.sd)²`, proportional to `f2` rather than to the prediction.

### Objective function values (NONMEM = nlmixr2, all exact to 1e-3)

| error model | variance `R` | NONMEM FOCEI | nlmixr2 FOCEI | NONMEM FOCE | nlmixr2 FOCE |
|---|---|--:|--:|--:|--:|
| prop | `(f·prop.sd)²` | 39.458 | 39.458 | 39.207 | 39.207 |
| propMod | `(f·f2)²` | 63.353 | 63.353 | 63.063 | 63.063 |
| add | `add.sd²` | −2.059 | −2.059 | −2.059 | −2.059 |
| addMod | `f2²` | 3.238 | 3.238 | 2.438 | 2.438 |
| propF | `(f2·prop.sd)²` | 6.496 | 6.496 | 6.275 | 6.275 |
| propFMod | `(f2·lipre)²` | 19.177 | 19.177 | 18.202 | 18.202 |
| add+prop (comb2) | `add.sd² + f²·prop.sd²` | 39.735 | 39.735 | 39.499 | 39.499 |
| add+prop (comb1) | `(add.sd + f·prop.sd)²` | 43.554 | 43.554 | 43.394 | 43.394 |
| addMod+propMod (comb2) | `f2² + f²·f3²` | 106.308 | 106.308 | 106.079 | 106.079 |
| add+propF (comb2) | `add.sd² + f2²·prop.sd²` | −2.321 | −2.321 | −2.454 | −2.454 |

`f = ipre` (prediction), `f2 = ipre/(ipre+5)`, `f3 = 3·f2`, `lipre = log(ipre)`.
nlmixr2 columns are the closed-form fits; the ODE fits are identical to them (below).

### What this establishes

- **FOCEI is unchanged and correct.** NONMEM reproduces every FOCEI value,
  including the eta-dependent-variance cases (`addMod`, `propMod`, the `Mod`
  combined models) where the interaction term is non-trivial.
- **The FOCE fix is correct against NONMEM.** Every FOCE value — the numbers this
  PR changes — matches NONMEM's `METHOD=1` (no `INTER`) exactly, across pure
  proportional, additive, `f`-referenced (`propF`), and combined (`comb1`/`comb2`)
  error, including models where the residual variance depends on the prediction
  or a prediction-derived quantity.
- **ODE = closed-form = NONMEM.** With the fix, nlmixr2's ODE FOCE objective now
  equals its closed-form objective for all 10 models, and both equal NONMEM. The
  eta-dependent-variance FOCE cases (`addMod` FOCE 2.438 vs FOCEI 3.238;
  `add+propF` FOCE −2.454 vs FOCEI −2.321) show the interaction correctly *off*
  for FOCE while FOCEI keeps it, matching NONMEM on both.

### Minor fixture note

`test-focei-wang2007-basic.R` hard-codes the propMod FOCE-ODE reference as
`63.001`; the current build (and NONMEM) give `63.063`. The test passes because
`testthat::expect_equal(tolerance = 1e-3)` is *relative* (0.062/63 ≈ 9.8e-4), but
the exact NONMEM-confirmed value is `63.063` if we want to tighten it.

mattfidler:
> test-focei-wang2007-basic.R hard-codes the propMod FOCE-ODE reference as
63.001; the current build (and NONMEM) give 63.063. The test passes because
testthat::expect_equal(tolerance = 1e-3) is relative (0.062/63 ≈ 9.8e-4), but
the exact NONMEM-confirmed value is 63.063 if we want to tighten it.

This is fine with me.  

The value **was** that in an earlier version of NONMEM.  It just shows that NONMEM reference values shift.  It may even be in this repository (My guess is that we used 7.4.3).

Perhaps you should (in ferx and here) reference the version, compiler, and reference for NONMEM.  I guess it makes a difference.

You do in the pull request, but maybe we should note it in the test. 

I do not yet have access to NONMEM 7.5.1 so I have no way to verify this :smile:

mattfidler:
@hiddevandebeek is this ready?

mattfidler:
Another possibility @hiddevandebeek for this is to have a build that runs nonmem to generate these objective function vales if they really need to be pinned to a specific version (and possibly save them as a CSV file). 

For the inner problem it is likely very similar as shown in this repository.

hiddevandebeek:
sure ill add that.

Im doing one last run to check if everything is clean.

Can you review it? Otherwise give me one last look at it.

What do you think about the FOCE difference?

mattfidler:
I will review it when you say you are ready.

hiddevandebeek:
okay, one last adjustment

hiddevandebeek:
@mattfidler okay done

hiddevandebeek:
okay now really done, adjusted it so it uses same gill83 step for full and minimal.

mattfidler:
Hi @hiddevandebeek; I couldn't edit your pull request so I moved it to another one (#703)

PR #703, opened 2026-07-05
Hiddevandebeek focei analytic cov

mattfidler:
Supercedes #700 .

I have some edits @hiddevandebeek.  As a note linerization with the `foce` performed quite well but I doubt it will perform as well with NONMEM's `foce` equivalent, so I'm keeping this and calling it a `focei+` variant.  I could make it into a separate method.  Right now I have turned off the analytic gradient, though it may be able to be calculated too.  It doesn't have to be calculated for the release.


mattfidler:
There are a few brittle things that need to be fixed here (covariates may not work and prop only models do not work).  Working through it and adding the fast method as a cross check.

If you want to look after I'm done let me know.

hiddevandebeek:
Yeah let me know!

mattfidler:
Currently the analytic gradient covers all models that transformably linear in all the `*foce*` methods, as well as ones with censored observations.  The `*foce*f` methods are the analytical gradient methods.  This also applies to ODEs and pred models (I think; the pred models is simply the gradients without the ODEs).

Also on the list are DDEs and inductive linearization models (which will also be integrated before CRAN).

However, these are the ones that you were working on.  Let me know if you have any comments.

Thanks for your help @hiddevandebeek

mattfidler:
@hiddevandebeek 

One last point, you did not have analytic as the default; was that because of backward compatibility or another reason?

hiddevandebeek:
sorry had a thing with work, yeah for backward compatibility. I have some additions for this PR in terms of efficiency and for the analytical outer gradient.

hiddevandebeek:
I will add and describe it in a minute or 10. its compiling now

hiddevandebeek:
btw, this looks very nice! very cool how fast the outer grad was implemented!

mattfidler:
> btw, this looks very nice! very cool how fast the outer grad was implemented!

Well, I had it in mind for awhile so many of the items were already built.

mattfidler:
The DDE fast-family / analytic-covariance path here depends on upstream rxode2 fixes in nlmixr2/rxode2#1106 (correct delay history when a delay model's Jacobian can't be generated; enable `ros4` for delayed sensitivity models). The augmented-cov solve forces `method=dop853` as a belt-and-suspenders workaround, and `tests/testthat/test-dde-focei.R` guards the pipeline end to end.

mattfidler:
Follow-up rxode2 fix nlmixr2/rxode2#1107 repairs the calcJac param-mangling for THETA/ETA-named delay models, so the DDE augmented-covariance solve now also works via the stiff `ros4`/`dop853+ros4` composite (previously it fell back to the finite-difference sandwich). The nlmixr2est cov solve still forces `dop853` here -- it is exact for the augmented model and version-independent -- but the composite is no longer broken for these models.

mattfidler:
Now DDEs go through analytic path too.

To avoid merging issues again, I am merging now.  The CI here seems to not be working well.  I will have to verify on my windows/pc computer before submitting to CRAN.

I do not have a mac computer 😄

PR #705, opened 2026-07-08
Efficiency: cache + batched/parallel solves for the analytic gradient & cov

hiddevandebeek:
Efficiency additions on top of the analytic gradient/cov in #703 (a clean fast-forward from the current branch tip — 3 files, `+184/−84`: `foceiCovAnalytic.R`, `foceiGradAnalytic.R`, one cov test). **Numerically inert: SEs are bit-identical; no change to the gradient math or model coverage.**

### What it does
**Solve/reader path (gradient + cov)**
- Augmented-model session cache (same `foceiModelDigest` key) — repeated fits skip the per-fit symbolic rebuild of the cov's augmented model.
- Matrix-extraction reader in `.foceiAnalyticSolveAll`: each sensitivity column extracted once and sliced by row, instead of per-subject `data.frame [[` + `paste0` (profiling: the ODE solve was ~4% of a gradient call — the reader was the cost).
- Event-table reuse (`etTrans` once, ~40%/solve) + explicit `rxControl$cores` on every analytic solve.

**Cov 3rd-order tensor**
- Batched Richardson FD3 (`.foceiAnalyticSolveAllFD3`): `1 + 2·neta` population solves instead of the per-subject Shi (~O(nsub·neta)), for FOCEI and FOCE alike (2-pass `AssembleR` restructure). Censored (f,R) path stays per-subject. `FOCEI_NO_FD3_BATCH` env var falls back to per-subject if ever needed.

### Measured
A/B on the identical base (`rxClean()`'d so each recompiles fresh; 12-subject theo model):

| | unoptimized | optimized |
|---|---|---|
| steady-state fit+cov | 9.3s | **2.1s (~4.4×)** |
| first-ever fit (recompile) | 27.9s | **18.1s** |

SEs bit-identical; running with FD3 batching on vs off gives identical suite results (no regressions). `test-focei-cens` 38/0; fast-grad and cov clean apart from a pre-existing FOCE-combined case.

### One scope decision to flag
Censored observations (M2/M3/M4) now fall back to the **FD cov** — censoring stays fully supported in the analytic **gradient**; only the cov's censored determinant is pulled. Trivial to revert if you'd rather keep it.


hiddevandebeek:
I routed censored observations (M2/M3/M4) out of the analytic observed-information cov and back to the FD cov, not a 100% sure about that yet. 

This doesn't touch censored fitting

hiddevandebeek:
will add some optimized support for other error models

mattfidler:
The censored gradients are exact so they should be added :smile:

mattfidler:
~~@hiddevandebeek  what other error models do you want to support, a t-distribution?~~

Nevermind, optimized values.  I understand :smile:

hiddevandebeek:
**Follow-up: extended the batched FD3 to the general `(f,R)` cov path (`AssembleRFR`).** Now FOCEI and foce+ get the batched/parallel analytic cov for *every* error model, not just add/prop:

- **add / prop / combined2** → `AssembleR` (batched — first commit)
- **combined1 `(sa+sp·f)²`, lognormal, box-cox/yeo-johnson, multi-endpoint, any general variance** → `AssembleRFR` (batched — this commit)

Both `AssembleRFR` branches mirror the `AssembleR` 2-pass restructure: the eta=0 population solve is batched (nonmem frozen R0), the EBE re-solve stays per-subject (Newton), then one `.foceiAnalyticSolveAllFD3` delivers `f/a/A/Ath` for all subjects — `+ R/aR/AR/AthR` for FOCEI (`withR=TRUE`); `R/aR/AR + Ath` for foce+ (`withR=FALSE`, since foce+ never needs `AthR`) — replacing the per-subject Shi. `FOCEI_NO_FD3_BATCH` falls back to per-subject.

**Validated on combined1** (which used to fall back to FD): cov=analytic for both FOCEI and foce+, batched-vs-per-subject SEs match to **5.7e-5 / 1.2e-5**. Cov suite **117/7 unchanged** — no regressions (the 7 are the pre-existing FOCE-combined case + the boxCox tests that need the dev-rxode2 `_powerDLambda`).

Censored (M2/M3/M4) still routes to the FD cov; the analytic gradient is untouched.

One diagnostic note: combined1 *looked* like it fell back to FD, but only at an unconverged θ (`maxOuter=0`) — the observed-information Hessian is indefinite there (a negative variance), so the fit correctly rejects the non-PD cov. At convergence it's PD → analytic. So the general-variance forms were in scope all along; this commit just batches them.

hiddevandebeek:
okay fine! Ill remove gating in the next commit for M methods, so its clean again 👍 
Ill do an automated profiling of the code during the night to search where parallel, r vs c++ or other stuff can be optimized further

mattfidler:
In theory the laplace censoring gradient is exact too, but it has numerical issues at extremes.  It isn't so clear-cut for the laplace censoring methods.

mattfidler:
This was merged into #703  it said it had divergent branches.

PR #706, opened 2026-07-09
Analytic FOCEI/FOCE covariance: efficiency layer (shared grad/cov model, sigma-skip, batched FOCE EBE)

hiddevandebeek:
## Summary

Efficiency + coverage layer on top of the analytic FOCEI/FOCE observed-information covariance from #703. The covariance and gradient **numbers are unchanged** — every change here is either speed-only (bit-identical) or test precision. Touches `R/foceiCovAnalytic.R`, `R/foceiGradAnalytic.R`, and `tests/testthat/test-cov-analytic.R`.

## What's in it

1. **Shared gradient/cov model, on by default (Route A + sigma-skip).** The `(f,R)` covariance now reuses the gradient's already-compiled `dirs` model (sigma tensor slots rebuilt from the model's own `rsig` outputs) instead of compiling a second model with explicit sigma directions. Previously behind `FOCEI_RSIG`/`FOCEI_SIGMA_SKIP`; now **default-on** with opt-out escape hatches `FOCEI_NO_RSIG` / `FOCEI_NO_SIGMA_SKIP`. One augmented-model compile instead of two, and the sigma directions' all-zero state-sensitivity columns vanish. Concretely (combined1, 2 sigmas): a separate 62-state cov model → the gradient's 26-state model (shared).

2. **Batched FOCE EBE re-solve in the gradient.** The analytic outer gradient's FOCE path re-solved each subject's EBEs with a per-subject Newton loop; it now uses the same batched `.foceiAnalyticFoceEbeBatch` the covariance uses — one rxode2 population solve per Newton step. Bit-identical to the per-subject loop (verified 0.0e+00 on combined1 FOCE-nonmem and foce+).

3. **Un-gated Gaussian censored gradients** (from earlier in the stack): M2/M3/M4 gradients are exact and run through the optimized batched solve.

4. **Test precision fix.** The analytic-vs-FD / gold-FD comparison tests fit at `sigdig = 6` so the analytic augmented solve (`.foceiAnalyticSolveTol = 10^-(sigdig+6)` → `1e-12`) matches the gold-FD / `covMethod="r"` side (which already solves at `1e-12`). Without it, two comparisons sat just past their tight thresholds on near-zero matrix entries (FOCE-combined R vs gold-FD 0.00346 vs 3e-3; censored-M3 SE 0.090 vs 0.05). No change to the covariance itself.

## Numbers unchanged

Route A / sigma-skip produce bit-identical results to the explicit-sigma-direction path (validated vs finite differences at tight tolerance, and identical with the fast path on vs off). Only compile/solve cost drops.

## Error-model coverage (analytic, across FOCEI / FOCE-nonmem / foce+)

- ✅ **add, prop, combined1, combined2, boxCox** — analytic. `prop` correctly falls back to FD only when a prediction actually reaches zero (the `canVanish` guard; e.g. theo's pre-dose point), and is analytic wherever predictions stay away from zero.
- ⏳ **lnorm, logitNorm** — currently fall back to FD. Root-caused to an upstream rxode2 bug in `.rxFromSEnum` (crashes on an empty operand produced when the 2nd-order sensitivity of a `log`/`logit`-transformed prediction leaves an unsimplified `exp(A-(A))`). Filed as **nlmixr2/rxode2#1109**. No change needed here once that lands; they fall back safely to FD in the meantime.

## Validation

`test-cov-analytic.R` passes (0 failures), `test-focei-fast-grad.R` and `test-focei-foce-plus.R` pass, when built against a dev rxode2 that has the required symbolic-sensitivity features.

**CI note:** the analytic-cov feature (this PR and #703) needs the dev rxode2 symbolic features; CI's current rxode2 lacks them, so the analytic path falls back to FD and the `covMethod="analytic"` tests report failures (they don't fail `R CMD check` — `stop_on_failure = FALSE`). Same behavior on main. The `R CMD check` ERROR on ubuntu-release is a pre-existing `nlmControl` (nlm) example failure, unrelated to this PR.

### DIFF
--- a/R/foceiControl.R
+++ b/R/foceiControl.R
@@ -86,10 +86,10 @@
 #'     finite differences, and use the Eq-48 random-effect extrapolation for the
 #'     next inner-problem starting values.  Requires an analytic-scope model
 #'     (single additive/proportional Gaussian endpoint); out-of-scope models fall
-#'     back to the finite-difference gradient with a message.  When unspecified,
-#'     the outer optimizer defaults to \code{"lbfgsb3c"} (vs \code{"nlminb"} for
-#'     \code{fast=FALSE}); pairing \code{fast=TRUE} with a derivative-free
-#'     \code{outerOpt} reverts to \code{fast=FALSE}.  The \code{*f} methods (e.g.
+#'     back to the finite-difference gradient with a message.  The outer
+#'     optimizer defaults to \code{"nlminb"} (as for \code{fast=FALSE}), which
+#'     converges robustly with the analytic gradient; pairing \code{fast=TRUE}
+#'     with a derivative-free \code{outerOpt} reverts to \code{fast=FALSE}.  The \code{*f} methods (e.g.
 #'     \code{foceif}) default this to \code{TRUE}.
 #'
 #' @param covTryHarder If the R matrix is non-positive definite and
@@ -1052,11 +1052,15 @@ foceiControl <- function(sigdig = 4, #
   if (!is.null(.xtra$outerOptFun)) {
     outerOptFun <- .xtra$outerOptFun
   } else if (rxode2::rxIs(outerOpt, "character")) {
-    # mode-dependent default when the user did not specify outerOpt: base methods
-    # use nlminb, fast methods use the gradient-friendly lbfgsb3c.  An explicit
+    # Default outer optimizer (when the user did not specify one): nlminb for both
+    # base and fast methods.  nlminb converges robustly with the analytic ("fast")
+    # gradient; lbfgsb3c is faster on well-conditioned problems but can settle at a
+    # worse local optimum on ill-conditioned models (e.g. multi-compartment), which
+    # both changes the reported objective and leaves the observed-information
+    # covariance indefinite (the analytic cov then falls back to FD).  An explicit
     # outerOpt (a single string) skips this.
     if (missing(outerOpt)) {
-      outerOpt <- if (isTRUE(fast)) "lbfgsb3c" else "nlminb"
+      outerOpt <- "nlminb"
     }
     outerOpt <- match.arg(outerOpt)
     .outerOptTxt <- outerOpt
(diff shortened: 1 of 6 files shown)


mattfidler:
Requested review from copilot

mattfidler:
@hiddevandebeek The review has returned.

Also fixed the issue you flagged in rxode2

mattfidler:
Pushed c526b6a8 on top of a87b97bb (which already addressed all four Copilot review comments -- digest cache key, bounded caches, up-front column checks in `.foceiAnalyticSolveAll()`, and named `events =`).

The new commit adds a regression test enabling the previously-pending transform coverage: **lnorm and logitNorm now run the analytic covariance** (FOCEI and FOCE) against an rxode2 with the `.rxFromSEnum` empty-operand fix (nlmixr2/rxode2#1111 for nlmixr2/rxode2#1109). Verified locally against that branch:

- logitNorm: `covMethod == "analytic"` for focei and foce; SEs within 5% of `covMethod="r"`.
- lnorm: same, once predose observations are excluded (a zero prediction gives `log(f) = -Inf`, so theo's t=0 points are dropped in the test); SEs within ~6% (focei) / ~4% (foce) of `covMethod="r"`.

No nlmixr2est-side gate needed removing -- the path already falls back only when the rxode2 symbolic build fails, so this lights up automatically once rxode2#1111 lands.

One unrelated observation from running the full `test-cov-analytic.R` locally: the `iovXform="var"` leg of the IOV test (`expect_true(is.matrix(fVAR$cov))`) is flaky -- the FD "r,s" fallback covariance for that fit intermittently fails on identical code (thread-order sensitivity), independent of this PR's changes.

mattfidler:
@hiddevandebeek is this ready.  

Also do you think the analytic `r` is enough (or should we still do a sandwich estimator).  I think analytic is enough, but I wanted your thoughts since you have explored this a bit more.

hiddevandebeek:
I think that it is valid right now and quite robust. over the coming period i will do small improvements in terms of code cleanliness etc., cause i do think its a bit cluttered.

hiddevandebeek:
But I do like that the fast path combined with analytical cov is optimized now

Issue #711, opened 2026-07-10
mufocei / muModel="lin": "undefined columns selected" with >=2 mu-referenced covariate expressions


hiddevandebeek:
**Summary**

The mu-referenced FOCEI family (`muModel="lin"`/`"irls"` — i.e. `mufocei`/`irlsfocei`/…) errors with `Error: undefined columns selected` when a model has **two or more** mu-referenced covariates that are algebraic **expressions** (e.g. `log(WT/70)`) rather than bare data columns. Affects both `fast=TRUE` and `fast=FALSE`.

**Reprex**

```r
library(nlmixr2est)
library(nlmixr2data)

mod <- function() {
  ini({
    tka <- log(1.5); tcl <- log(2.7); tv <- log(31.5)
    wtcl <- 0.75; wtv <- 1
    eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1
    add.sd <- 0.7
  })
  model({
    ka <- exp(tka + eta.ka)
    cl <- exp(tcl + eta.cl + wtcl * log(WT/70))   # covariate #1: expression log(WT/70) on CL
    v  <- exp(tv  + eta.v  + wtv  * log(WT/70))   # covariate #2: same expression on V
    d/dt(depot)  <- -ka * depot
    d/dt(center) <-  ka * depot - cl/v * center
    cp <- center / v
    cp ~ add(add.sd)
  })
}

nlmixr2(mod, theo_sd, "focei", foceiControl(muModel = "lin"))
#> Error: undefined columns selected
```

**Scope**

| model | result |
|---|---|
| ≥2 covariate **expressions** (`log(WT/70)` on CL and V) | ❌ error |
| 1 covariate expression | ✅ ok (never reaches the failing path) |
| ≥2 **raw-column** covariates (`WT` directly) | ✅ ok |

Independent of `fast`.

**Diagnosis**

[...]

PR #714, opened 2026-07-11
focei(analytic): reuse eta sensitivities for mu-referenced covariate coefficients

hiddevandebeek:
## Summary

The analytic FOCEI/FOCE outer gradient (`fast=TRUE`) and observed-information covariance (`covType="analytic"`) build an augmented sensitivity model. A mu-referenced covariate coefficient `b` enters through an eta's parameter (`cl = exp(tcl + eta.cl + b*cov)`), so for a covariate constant within each subject every sensitivity of `b` equals the linked eta's scaled by the covariate value (`df/db = cov*df/deta`, and likewise for the 2nd-order terms). Those covariate directions are now emitted as algebraic scaled copies of the linked eta's columns instead of integrating their own state-sensitivity ODEs.

This shrinks the compiled augmented model ~25–58% and cuts its (super-linear) build time substantially. The result is **bit-identical** — same optimum, same covariance.

## Speedup (measured, theo_sd)

| model | aug lines | build (CSE + compile) | end-to-end fit |
|---|---|---|---|
| 1 covariate | 126 → 91 | 5.7 → 3.8s (−33%) | 18.9 → 17.7s |
| 3 covariates | 243 → 122 | 10.9 → 5.0s (−55%) | 30.1 → 22.6s |
| 2-cmt, 5 covariates | 996 → 420 | 106 → 24.6s (**4.3×**) | 248.7 → 55.8s (**4.5×**) |

The saving grows with model size (the `rxOptExpr` CSE pass is super-linear in the model size).

## How it works

- **Detection reuses rxode2's own mu-reference classification** — `muRefDataFrame` (theta↔eta), `muRefCovariateDataFrame` (bare covariates), and `mu2RefCovariateReplaceDataFrame` (algebraic covariates, the same frame the `muRefCovAlg` / #711 machinery uses) — so it covers bare (`b*WT`, `b*SEX`) and algebraic (`b*log(WT/70)`, `b*(WT-70)`) covariates without re-parsing the model.
- **Guards.** The eta-scaling identity holds only when the eta enters solely through that one additive combination. It falls back to the full symbolic build for a time-varying covariate (a subject-constant gate computed from the raw data before the outer model is built) or for an eta shared/reused across parameters (which rxode2's classifier does not flag); a scaled eta (`0.5*eta`) is already excluded by `muRefDataFrame`.
- Also restricts the analytic covariance to **Gaussian endpoints**: a non-normal likelihood (`t`/`cauchy`/count/ordinal) or multiple estimated transform lambdas now fall back to the finite-difference covariance instead of being assembled as Gaussian.

## Validation

- **Bit-identical** to the non-reused build: objf + every theta + every SE identical, verified across add / prop / combined1 / combined2 / allometric / two-eta / bounded-coefficient models and a 2-compartment 5-covariate model, and against the current `main`.
- **Same optimum**: the objective is unchanged at every theta (it is computed by the inner EBE model, which the reuse never touches).
- Analytic-covariance and fast-gradient test suites pass; adds a dedicated covariate-reuse test.


hiddevandebeek:
let me know what you think. optimization was already made for muref theta earlier

hiddevandebeek:
eta guard is better for mu-ref theta now

Issue #716, opened 2026-07-12
Parallelize rxOptExpr chunk optimization in the analytic augmented-model build


hiddevandebeek:
### Summary

The analytic FOCEI/FOCE augmented sensitivity model (`covType="analytic"` / `foceiControl(fast=TRUE)`) is optimized with `rxOptExpr` in **contiguous line-chunks** (`.foceiAnalyticAugModelDirs`, `R/foceiCovAnalytic.R`) because `rxOptExpr` is ~O(n^3.5) in model size. The chunks are independent (each optimized separately, its introduced vars renamed to a per-chunk `rx_expr_c<g>_` prefix), so the current sequential `vapply` can be parallelized. Two independent tuning knobs turned out to matter: the **parallel backend** and the **chunk size / balancing**.

### Benchmark (Windows, 14 cores → 8 workers, rxode2 5.1.3; 266-line augmented model)

**Backends (size-40 chunks, k=7):**

| backend | compute | startup |
|---|---|---|
| sequential (current) | 7.86s | — |
| PSOCK cluster | 2.97s | 2.04s **per build** |
| **mirai** (persistent daemons) | ~3.5s | 0.78s **once per session** |

mirai's per-call compute is comparable to PSOCK, but its **daemons persist** — repeated builds (bootstrap / SCM / parallel workers) pay startup once, vs PSOCK re-paying ~2s every build — and it is a single cross-platform path (no PSOCK-vs-`mclapply` split). Correctness: mirai output is equivalent — the `rx_expr_N` counter is daemon-stateful so numbering differs, but the per-chunk `c<g>_` prefix keeps namespaces distinct and the assembled model identical.

**Chunk size is the larger lever** (mirai daemons warm):

| chunk size | k | sequential | mirai | model growth (cross-chunk CSE lost) |
|---|---|---|---|---|
| 40 (current) | 7 | 8.14s | 3.47s | +10% |
| 25 | 11 | 6.35s | 2.07s | +14% |
| 15 | 18 | 4.77s | 1.29s | +24% |
| 10 | 27 | 4.30s | 1.07s | +33% |

[...]

PR #723, opened 2026-07-13
focei(analytic): cost-balanced + optional mirai-parallel rxOptExpr chunking

hiddevandebeek:
## Summary

Reworks the analytic FOCEI/FOCE augmented-model `rxOptExpr` pass (built in `.foceiAnalyticAugModelDirs`) to be faster, building on the compartment-scoped gate from #716.

`rxOptExpr` is ~O(n^3.5) in model size, so the augmented model is optimized in contiguous line-chunks. Two changes:

1. **Cost-balanced chunking (always on, free).** Chunks are split by equal character-sum instead of equal line count, so the largest chunk — which drives the O(n^3.5) cost and the parallel floor — is minimized. The 2nd-order sensitivity equations are long and uneven, so equal-line-count chunks left one dominant chunk.

2. **Optional parallel optimization via `mirai` (`foceiControl(optExprParallel = TRUE)`).** The chunks are independent (each `rxOptExpr`'d separately, its introduced vars namespaced by a per-chunk `rx_expr_c<g>_` prefix), so they dispatch across an active `mirai` daemon pool. Persistent daemons (set up once with `mirai::daemons()`) make this ideal for bootstrap / covariate search, where the augmented model is rebuilt in many workers; it's cross-platform with no per-build cluster startup. Default `FALSE` uses the cost-balanced chunking sequentially.

Compartment-scoped models (a parameter-dependent state IC or dosing modifier, gated in #716) continue to optimize whole and un-chunked — the chunker only sees safe models, so no line-holdout is needed.

## Correctness

Mathematically identical to the whole-model optimization — validated bit-identical:
- parallel vs sequential fit SEs match to 0
- optimized-vs-whole solve `max|diff| = 0` on the augmented model

Each chunk is collapsed to a single string, so a parse error degrades that chunk to *unoptimized-but-correct* (length-safe) rather than aborting the pass.

## Chunk-size choice

Kept at ~40 lines (the parallel build-time optimum). The full-build tradeoff is `rxOptExpr` (falls with smaller chunks) vs `eventSens`/compile (rise as the model grows from lost cross-chunk CSE). The per-solve cost of that growth is negligible (the extra lines are algebraic intermediates, not ODE states — ~3% at +85% lines, crossover ~300k solves).

## Notes

- `mirai` added to Suggests; the option is a no-op without it (and without an active pool).
- `man/foceiControl.Rd` was hand-updated (couldn't run `document()` in my env, which needs the unreleased rxode2) — worth regenerating.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -84,7 +84,8 @@ Suggests:
     fastGHQuad,
     rmarkdown,
     rlang,
-    numDeriv
+    numDeriv,
+    mirai
 LinkingTo:
     BH,
     n1qn1 (>= 6.0.1-12),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,17 @@
 # nlmixr2est (development version)
 
+- The analytic FOCEI/FOCE augmented-model `rxOptExpr` pass now splits the model
+  into **cost-balanced** contiguous chunks (equal character-sum rather than equal
+  line count) so the largest chunk -- which drives the ~O(n^3.5) cost -- is
+  minimized.  With `foceiControl(optExprParallel = TRUE)` and an active `mirai`
+  daemon pool (`mirai::daemons()`), the independent chunks are optimized in
+  parallel, cutting the pass several-fold on larger covariate models; the result
+  is mathematically identical (bit-identical solves).  Most valuable for
+  bootstrap / covariate search, where the augmented model is rebuilt across many
+  workers and the daemon pool is set up once.  (Compartment-scoped models -- a
+  parameter-dependent state IC or dosing modifier -- continue to optimize whole
+  and un-chunked, as before.)
+
 - Added an automatic differentiation variational inference method
   (`est = "advi"`, Kucukelbir et al. 2017) with `adviControl()`.  The variational
   gradient is obtained from the FOCEi forward sensitivities (the inner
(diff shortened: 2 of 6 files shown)


mattfidler:
This is really the wrong place for this (and #724)  They should go in rxode2 so it can be applied anywhere.  Also that is where the mirari dependency lives.

hiddevandebeek:
okay will adjust

hiddevandebeek:
Closing per your review, @mattfidler — agreed, this belongs in **rxode2** next to `rxOptExpr` (and that's where the `mirai` dependency already lives) so any caller benefits, not just the FOCEI analytic path. Re-opening there; #724 (the compartment-scoped ungating that builds on this) is closed for the same reason and moves with it.

For the rxode2 version I'll follow the convention already in `rxOom.R` — `parallel` as an integer daemon count (0 = serial), the call managing its own `mirai::daemons()` pool with `on.exit(daemons(0))`, and a hard `stop()` on a failed chunk — rather than the boolean + externally-managed persistent pool used here.

Carrying over the validated results: cost-balanced chunking is ~14× faster than the whole-model `rxOptExpr` call on a 275-line augmented model (O(n^3.5)), parallel adds ~2.6× on top, and the chunked model is **bit-identical** on solve with **no solve-time cost** (1.00×), so it's a pure build-time win.

PR #724, opened 2026-07-13
focei(analytic): ungate compartment-scoped models (chunk via in-place disguise) [stacked on #723]

hiddevandebeek:
## Summary

**Stacked on #723** — this builds directly on that branch. Until #723 merges, the diff here includes #723's commits too; the **net-new change is the two ungating commits** and it collapses to just those once #723 lands. Please review after #723.

Ungates the compartment-scoped models that #723 (following `34cec2d0`) sends to a single whole-model `rxOptExpr` call. Those models — a parameter-dependent state initial condition (`state(0)=`) or a dosing modifier (`f/alag/lag/rate/dur(cmt)=`) — now get the chunked/parallel speedup too.

## How

`rxOptExpr` parses each chunk as a standalone model, so a `state(0)=` / modifier line in a chunk without its `d/dt(cmt)` is a parse error, and the lines can't be *moved* (their RHS may read a variable a later line reassigns — position is load-bearing).

So instead of moving them, **disguise each such line's LHS as a unique plain name in place** for the optimization pass and restore it after:
- `W(0)=W0` → `rx__disg_ic__W__=W0` → optimize → `W(0)=W0`
- `f(depot)=…` → `rx__disg_mod__f__depot__=…` → … → `f(depot)=…`

The line never changes position, so semantics are preserved; it parses fine in any chunk as a plain assignment; and restore rewrites the LHS only (split at first `=`), leaving the optimized RHS untouched.

## Validation (bit-identical, `solve max|diff| = 0`)

- **The neonatal weight model (`W(0)=W0`) that motivated the gate** — its 95-line augmented model built through the real `.vaeDecoderModel` path: the old chunker orphans `W(0)` and prints `'W(0)' present, but d/dt(W) not defined`; the disguise runs **clean and fully optimized**.
- A param-dependent-IC + `f()`/`lag()` model, **sequential and mirai-parallel**.
- The real augmented model and a plain model — unaffected.
- Disguise/restore round-trips exactly on state ICs, sensitivity-compartment ICs (`rx__sens_..._(0)=`), and every dosing-modifier form.

Adds a round-trip unit test + a param-dependent-IC fit test; drops the "only non-special models" caveat from the `optExprParallel` docs; NEWS. The per-chunk fallback keeps any *unforeseen* construct correct-but-unoptimized.

## Not covered locally

The full end-to-end VAE fit needs `neonatal_wt` + the unreleased rxode2, so that's left to CI — but the **augmented-model build**, the exact step that orphaned, is validated here.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -84,7 +84,8 @@ Suggests:
     fastGHQuad,
     rmarkdown,
     rlang,
-    numDeriv
+    numDeriv,
+    mirai
 LinkingTo:
     BH,
     n1qn1 (>= 6.0.1-12),
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,9 +8,13 @@
   parallel, cutting the pass several-fold on larger covariate models; the result
   is mathematically identical (bit-identical solves).  Most valuable for
   bootstrap / covariate search, where the augmented model is rebuilt across many
-  workers and the daemon pool is set up once.  (Compartment-scoped models -- a
-  parameter-dependent state IC or dosing modifier -- continue to optimize whole
-  and un-chunked, as before.)
+  workers and the daemon pool is set up once.  Compartment-scoped models -- a
+  parameter-dependent state initial condition (`state(0)=`) or a dosing modifier
+  (`f/lag/rate/dur(cmt)=`) -- are now chunked too: those lines are disguised as
+  plain assignments **in place** for the optimization pass and restored after, so
+  they never orphan a chunk (which would be a parse error) and never move (which
+  would change what a later reassignment reads), yet still get the chunked/parallel
+  speedup -- previously they fell back to a single whole-model call.
 
 - Added an automatic differentiation variational inference method
   (`est = "advi"`, Kucukelbir et al. 2017) with `adviControl()`.  The variational
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,17 @@
 # nlmixr2est (development version)
 
+- The analytic FOCEI/FOCE augmented-model `rxOptExpr` pass now splits the model
+  into **cost-balanced** contiguous chunks (equal character-sum rather than equal
+  line count) so the largest chunk -- which drives the ~O(n^3.5) cost -- is
+  minimized.  With `foceiControl(optExprParallel = TRUE)` and an active `mirai`
+  daemon pool (`mirai::daemons()`), the independent chunks are optimized in
+  parallel, cutting the pass several-fold on larger covariate models; the result
+  is mathematically identical (bit-identical solves).  Most valuable for
+  bootstrap / covariate search, where the augmented model is rebuilt across many
+  workers and the daemon pool is set up once.  (Compartment-scoped models -- a
+  parameter-dependent state IC or dosing modifier -- continue to optimize whole
+  and un-chunked, as before.)
+
 - Added an automatic differentiation variational inference method
   (`est = "advi"`, Kucukelbir et al. 2017) with `adviControl()`.  The variational
   gradient is obtained from the FOCEi forward sensitivities (the inner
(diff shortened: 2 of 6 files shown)


hiddevandebeek:
## Robustness validation

Stress-tested the disguise/restore + chunking on the **raw** augmented models the chunker actually receives in production (the un-optimized sensitivity text assembled at `foceiCovAnalytic.R:~1420`, before the single `rxOptExpr` pass). Four structurally diverse models, each compared whole-model optimization vs chunked optimization:

| model | lines | chunks | struct (states+params) | seq cl=40 | seq cl=10 | **parallel cl=40** | orphan errs |
|---|---|---|---|---|---|---|---|
| growth, param IC | 95 | 2 | match | **0** | **0** | **0** | 0 |
| 3-cmt linear, 6 ICs + `f`/`alag` | 275 | 7 | match | **0** | **0** | **0** | 0 |
| `rate`/`dur` modeled infusion | 119 | 3 | match | **0** | **0** | **0** | 0 |
| Michaelis-Menten, mid-cmt IC + `lag` | 149 | 4 | match | **0** | **0** | **0** | 0 |

Every model solves **bit-identically** (`max|diff| = 0`) to the whole-model optimization across sequential chunking at two chunk sizes *and* parallel chunking, with states+params matching exactly and zero orphan (`'X(0)' present, but d/dt(X) not defined`) errors.

### Whitespace-exactness (motivated the latest commit)

The disguise originally rebuilt each target line as `paste0(newLhs, "=", rhs)`, which dropped whitespace between the LHS token and `=` (`peri(0) = q/vp` → `peri(0)= q/vp`). Always solve-identical (rxode2 ignores that whitespace), but not byte-exact. The fix captures/re-emits leading indent and the gap before `=` verbatim on both sides. Round-trip `restore(disguise(x))` is now **byte-identical** for every spacing variant tested — leading spaces/tabs, spaces around `=`, spaces inside the LHS token (`depot (0)`), spaces inside the parens (`f( depot )`), trailing `\r` — and remains idempotent.

Belt-and-suspenders, because the input can't actually carry user whitespace: `rxNorm` canonicalizes all of it upstream. Verified — a user writing `f (depot) = ff`, `alag  (depot) = tlag`, `depot (0) = ic0` (spaces before the paren, around `=`) all normalize to `f(depot)=ff;` / `alag(depot)=tlag;` / `depot(0)=ic0;` before the sensitivity builder ever assembles the model text.

### Note on the earlier confusion

An initial pass appeared to show failures on the multi-construct models. That was a **test-fixture artifact**, not a code bug: `rxNorm(am$augMod)` returns the *already*-chunk-optimized model (contains `rx_expr_c1_...`), so re-feeding it double-optimized into `rx_expr_c2_c1_...`. Rebuilding the fixtures with `optExpression=FALSE` gives the true raw input (0 `rx_expr_`), on which everything is bit-identical as above.

hiddevandebeek:
Closing per @mattfidler's review on #723 — this belongs in **rxode2**, alongside `rxOptExpr` (where the `mirai` dependency already lives), so it applies to every caller rather than just the FOCEI analytic augmented model.

Re-opening against rxode2. Carrying over what was validated here, so nothing is lost:

- **In-place disguise** (`state(0)=` / `f|alag|lag|rate|dur(cmt)=` → unique plain name → optimize → restore) is what lets a compartment-scoped model be chunked at all: the line never moves (its RHS may read a variable a later line reassigns — position is load-bearing), and it parses standalone in any chunk.
- **Bit-identical** (`solve max|diff| = 0`) vs whole-model optimization on 4 structurally diverse raw augmented models (growth+IC; 3-cmt with 6 ICs + `f`/`alag`; modeled `rate`/`dur` infusion; Michaelis-Menten with mid-cmt IC + `lag`), across sequential chunking at two chunk sizes and parallel chunking, with states+params matching and zero orphan errors.
- **Round-trip is byte-exact** for every whitespace variant, and idempotent. (In practice moot — `rxNorm` canonicalizes all user spacing upstream — but it makes the invariant unconditional.)
- **Chunking costs no solve time**: chunk-opt vs whole-opt solve in 0.0300 s vs 0.0300 s (1.00×) despite weaker per-chunk CSE, so it's a pure build-time win.

The rxode2 version will follow the existing `rxOom.R` mirai convention (integer daemon count, call-managed `mirai::daemons()` pool, hard-stop on a failed chunk) rather than the boolean/persistent-pool design used here.

PR #733, opened 2026-07-14
Fix fast=TRUE analytic outer gradient for prediction-dependent residual variance (prop/pow/combined)

hiddevandebeek:
## Summary

`foceiControl(fast=TRUE)`'s analytic outer gradient is **wrong for every FOCEI model whose residual variance depends on the prediction** — `prop()`, `add()+prop()`, `combined1`, `pow()`, `add()+pow()`. Errors range from **20% to 620%**. Since proportional/combined error with interaction is the default FOCEI setup, this affects `foceif` / `mufoceif` on essentially every real PK model.

The symptom is that `foceif` stalls with `false convergence (8)` at a worse objective than plain `focei` — the optimizer is handed a gradient that isn't the derivative of the objective it's evaluating.

## Root cause

`foceiGradSubjectFR_` (`src/foceiGrad.cpp`) chains the Laplace determinant coefficients through a direction as

```cpp
double ddfr = pffR[o] * a(o, s) + pfRR[o] * aR(o, s);   // d(dfr)/ddir_s
```

reusing `pffR` (`= d(dff)/dR`) as `d(dfr)/df`. Those are the same quantity only if `(dff, dfr, drr)` are second partials of a potential, so mixed partials commute. That is true for the **exact censored Laplace determinant** (`cp[2..4]`/`cp[5..8]` are genuine 2nd/3rd derivatives of the censored log-density) — which is presumably how the aliasing got written — but **false for a normal observation**, where the determinant uses the Gauss-Newton *expected* information:

```
(dff, dfr, drr) = (1/R, 0, 0.5/R^2)
```

This triple is not integrable. `dfr` is identically zero in both arguments, so `d(dfr)/df = 0`, while `d(dff)/dR = -1/R^2`. Symmetry would force those to be equal; they aren't. The aliasing injects a spurious

```
-a(s)/R^2 * (a_l*aR_m + aR_l*a_m)
```

into `dHt/ddir`.

## Why the failure pattern looks the way it does

The spurious term is proportional to `aR = dR/ddir`, so it **cancels identically whenever the variance doesn't move with the direction**. That exactly predicts what's observed:

| error form | FOCEI (before) | FOCE (before) | FOCEI (after) |
|---|---|---|---|
| `add` | 0.00% ✓ | 0.00% ✓ | 0.00% |
| `lnorm` | 0.03% ✓ | 0.01% ✓ | 0.03% |
| `boxCox` / `yeoJohnson` | 0.01% ✓ | 0.00% ✓ | 0.01% |
| **`prop`** | **138.7%** ✗ | 0.00% ✓ | **0.04%** |
| **`add+prop`** | **19.8%** ✗ | 0.01% ✓ | **0.00%** |
| **`combined1`** | **185.0%** ✗ | — | **0.02%** |
| **`pow`** | **623.8%** ✗ | 0.00% ✓ | **0.01%** |
| **`add+pow`** | **313.9%** ✗ | 0.02% ✓ | **0.03%** |

Constant-variance forms (`add`, `lnorm`, transforms on an additive endpoint) and **all FOCE variants** (frozen variance ⇒ `aR = 0`) were already correct and are unchanged.

Reference is Richardson-extrapolated central differences with a per-parameter step at `atol/rtol/atolSens/rtolSens = 1e-12`; `|D(h) − D(h/2)| / |D|` is ≤ 2e-4 throughout, so the reference is trustworthy to 5–6 significant figures.

## Fix

Introduce `pfrf = d(dfr)/df` as its own coefficient — `0` for a normal observation, `cp[6]` for the exact censored Laplace determinant (where the potential *is* integrable, so the previous aliasing was legitimate there). `dHtSg` needed no change: sigma moves only `R`, so it already used `pfRR = d(dfr)/dR = 0` and never took the bad path.

## Why the existing tests missed it

`test-focei-fast-grad.R` **does** have a FOCEI + `prop()` test ("matches FD for a covariate model"), but it evaluates far from the optimum, where the data-fit gradient (`prop.sd` ≈ −5.2e4) swamps the corrupted `log|Ht|` determinant term. It passed at 0.5% error against a 1% tolerance **with the bug present**. On the fixed build it now comes out exact (ratio 1.0000 on every parameter).

Added a regression test that evaluates **near the optimum** with three etas, where the determinant term is a first-order contributor. It fails by **>800%** with the bug and is exact without it. It also uses a per-parameter FD step — a flat `h=1e-3` is a 1% perturbation of `prop.sd=0.1` and makes the reference itself carry ~10% error, enough to mask the signal.

Note these gradient tests are `skip_on_cran()` + `skip_on_ci()`, so they don't run in CI.

## Verification

- Fixed build: exact (≤0.04%) across `add`, `prop`, `add+prop`, `combined1`, `pow`, `add+pow`, `lnorm`, `boxCox(est)`, `yeoJohnson`, under both FOCEI and FOCE.
- Existing add-only and FOCE/censored gradient tests unchanged.
- The analytic covariance already agreed with the FD covariance on prop models (0.1%), which is what localized the bug to the gradient kernel rather than the shared solved R-sensitivities.


hiddevandebeek:
Was working on the mu-ref fo-family methods and noticed slight underconvergence which wasn't closable with tolerances.
In terms of code: done, I will trim the comments now for the second commit, then you can review it.

hiddevandebeek:
so the bug isnt extremely large in terms of final found nll, but simple and does speed up convergence

mattfidler:
Let me know when you are done, I have merged with main (and it is passing).

hiddevandebeek:
Done, implementation/theory is now correct

mattfidler:
Also do you want to add analytic gradients for agq and generalized likelihoods (not for this release).  If so I think the likelihoods need the Hessian calculated too.

hiddevandebeek:
yeah sounds good, will work on that the coming week then!

PR #739, opened 2026-07-15
cran-07-2026: full r,s sandwich covariance + CRAN test fixes

mattfidler:
## Summary

CRAN-prep branch. Two threads of work:

**1. Full theta+sigma+Omega finite-difference covariance (true r,s sandwich).** Merges `focei-full-rs-cov`: with `covFull=TRUE` (the consistent default) the FD covariance methods report the full theta + residual sigma + Omega covariance. `covMethod="r,s"` is now a true full sandwich `solve(Rfull) %*% Sfull %*% solve(Rfull)` (previously the full shape was only the Hessian inverse), `"s"` is `solve(Sfull)`, `"r"` is `solve(Rfull)`; `fit$covR/covS/covRS` carry the matching full shape. This also applies when `covMethod="analytic"` is out of scope and falls back to the FD sandwich (previously stayed theta-only). FD-full bows out under bounded-parameter transforms (the theta-sized Jacobian hook cannot correct a full cov).

**2. Test fixes for CRAN robustness.**
- `test-cov-analytic.R` (boxCox lambda): compare the analytic cov to `covMethod="r"` (same estimand) at `sigdig=6`. The old comparison to `"s"` only matched because the full-cov install mislabeled `"s"` as the Hessian inverse; now `"s"` is the true OPG `solve(Sfull)` and legitimately disagrees on 12-subject data.
- `test-mu-plain-fit.R`: iteration-print check updated to the standard-column mu layout (the bolt-on `|   mu|` row was removed).
- `test-focei-foce-plus.R` (mfocep/ifocep): pin `warm="save"`. The default `warm="calc"` (always-recalc n1qn1 seed Hessian) steers the multi-modal foce+ mu-profiled restart into a shallower basin; `warm="save"` keeps the deeper-mode invariant the test guards.
- `test-muRefResetOptOut.R`: since plain-mu profiling, the plain mu eta is also regression-managed/protected, and `warm="calc"`'s full-Newton first inner step erases the retention signal -- pin `warm="save"`, assert both mu-group etas retain >2x drift vs the `muModel="none"` baseline.
- `test-matexp.R`: `rxWithSeed` for the simulated data (no bare `set.seed` leak); loosen the mu/irls SE ballpark to `8e-2` (the mu-regression re-derives the mu-thetas per form and the near-collinear MM parameters inflate that into a deterministic ~5-6% SE difference; the `covMethod=="analytic"` and objf checks stay the real guards).
- `test-saem-mix.R`: `rxWithSeed` for the split-ETA simulated data.

## Test plan
- `test-cov-analytic.R`: 177/177 pass
- `test-matexp.R`: 44/44 pass
- `test-focei-foce-plus.R`: 11/11, `test-muRefResetOptOut.R`: 6/6, `test-mu-plain-fit.R`: 28/28, `test-saem-mix.R`: 43/43
- Related cov files (cov-focei, table-cmt, cov-robust, fix-cov, cov-mu, cov-fdfull-install) green

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -150,6 +150,11 @@
   install skipped them because the fit tables were rendered before the
   full-model covariance was recomputed.
 
+- Converting a fit to a different covariance (`setCov()`, `getVarCov()`)
+  now refreshes `Condition#(Cov)`/`Condition#(Cor)` and the eigen
+  diagnostics from the newly installed covariance instead of leaving the
+  previous method's values in place.
+
 - SAEM `covMethod = "fim"` adds the mu-block Hessian (was indefinite / NaN SEs),
   and `"fim"`/`"sa"` report off-diagonal Omega and combined residual SEs.
   Fixed `covMethod = "linFim"` and the SAEM covariance erroring for a single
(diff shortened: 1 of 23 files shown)


mattfidler:
- [ ] irlsfoceif not working having a condition number.

PR #743, opened 2026-07-16
Reuse the fit's persistent rxInv for the analytic Omega derivatives

hiddevandebeek:
`.foceiEstOmegaDeriv()` builds a fresh `rxSymInvChol` env on every analytic outer-gradient call (`foceiControl(fast = TRUE)`). A fresh env pays a one-time symbolic setup on its first `$d.omegaInv`; a repeat read on the same env is free, and a 3x3 diagonal Omega has no 60ms of arithmetic in it. Measured, that setup was **~40% of each analytic gradient**.

The fit env already holds a persistent `rxInv` that C++ keeps current — `foceiSetup` takes it as `_rxInv`, `updateTheta` pushes each new Omega in via `setOmegaTheta`, and `analyticOuterGrad` runs `foceiOfv0` before calling back into R. This passes the live env down and reuses it, paying the symbolic setup once per fit instead of once per gradient.

| | before | after |
|---|---|---|
| `.foceiEstOmegaDeriv` per call | 45ms | **0.2ms** |

Benefits the whole FOCEI fast family, not one estimator.

### Correctness

Reused-env and fresh-env derivatives are **bit-identical** (`max|d.omegaInv diff| = 0`, `max|tr.28 diff| = 0`).

The reuse is **guarded**: the env is used only when its own omega is the Omega being asked about, otherwise it falls back to a fresh build. In a live fit the check always passes (measured 15/15 gradient calls, guard never fired), so it is deliberate insurance rather than load-bearing — `rxSymInvCholEnv` is a `list`, not an environment, and C++ takes `as<List>(rxInv)`, so the two only stay in sync because the list references an inner environment (`create.env = TRUE`). If that invariant ever broke, this degrades to a no-op instead of to a silently wrong gradient.

### Tests

New test in `test-focei-fast-grad.R` asserting the reuse is exact and that a stale env (wrong Omega), a junk `rxInv`, and an absent `rxInv` all fall back to the fresh result.

Existing suite: `test-focei-fast-grad.R` passes 47 assertions, 0 failed, 0 skipped.

### DIFF
--- a/R/foceiGradAnalytic.R
+++ b/R/foceiGradAnalytic.R
@@ -586,7 +586,7 @@
 #' post-fit gradient paths.  Returns a list of the assembled pieces, or `NULL`
 #' (out of scope).  `thVals` is the named converged theta vector.
 #' @noRd
-.foceiAnalyticGradSetup <- function(ui, thVals, Om) {
+.foceiAnalyticGradSetup <- function(ui, thVals, Om, e = NULL) {
   if (!isTRUE(rxode2::rxGetControl(ui, "fast", FALSE))) return(NULL)
   if (!.hasRxSens()) return(NULL)
   if (isTRUE(any(ui$predDf$linCmt))) return(NULL)   # linCmt(): no symbolic state sensitivities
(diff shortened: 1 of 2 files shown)


mattfidler:
Thanks @hiddevandebeek

PR #773, opened 2026-07-17
Warn about linCmt() dosing properties on the central compartment (#358)

### LINKED ISSUE #358: However, as you pointed out, the bio-availability to the central compartment in `focei`'s linear compartment seems to not be working at all:

mattfidler:
Fixes #358.

## Investigation

Issue #358 reported that bioavailability applied to the central compartment of a `focei` `linCmt()` model "seems to not be working at all". I reproduced the underlying behavior:

- A first-order absorption `linCmt()` model has two implied compartments, `depot` (the dosing compartment) and `central`.
- `rxSolve`/`linCmtB` only applies a dosing property (`f`/`alag`/`rate`/`dur`) to the compartment that actually receives the dose. With a depot dose, `f(central)` is **silently ignored** -- the `bio_cov=0` and `bio_cov=1` curves come out identical (ratio 1.0 instead of the expected 0.6).
- A `focei` fit therefore leaves the bioavailability parameter frozen at its initial value (its objective is flat in that parameter), which looks like a "spurious recovery" with an implausibly tiny SE.

This matches the maintainer's own diagnosis in the linked nlmixr2/nlmixr2#160 thread: bioavailability must be applied to the dosing compartment (`f(depot)`), and `f(central)` does nothing here. The remaining problem is that it was ignored **silently**.

## Fix

Add a preprocessing hook (`.preProcessLinCmtDosing`) that runs before estimation for every method and warns when a dosing property (`f()`/`alag()`/`rate()`/`dur()`) is applied to `central` in a first-order absorption `linCmt()` model, pointing the user to `depot`.

Detection uses `ui$props`:
- only fires for `linCmt()` models that have an implied `depot` compartment (first-order absorption). A pure IV `linCmt()` doses `central` directly, so `f(central)` there is legitimate and is left alone.
- checks `ui$props$cmtProp` for a dosing property on `central`.

The warning surfaces through the fit's `runInfo` (the standard nlmixr2 warning channel), and does not fire for the correct `f(depot)` form or for IV models.

## Tests

`tests/testthat/test-issue-358.R`:
- unit checks the hook directly for `f(central)`/`alag(central)` (warns), `f(depot)` (no warn), and IV `f(central)` (no warn);
- an end-to-end `focei` fit asserts the warning reaches `fit$runInfo`.

Note: `est="foi"` currently errors ("cannot find foi related control object") on the base branch independent of this change, which trips the fixture prefit helper; that is a separate pre-existing issue.


mattfidler:
Terriblei

PR #783, opened 2026-07-17
Fix restart crash for mu-referenced / variant FOCEi controls

mattfidler:
## The bug

A FOCEi-family fit whose control is not literally a `foceiControl` aborted on
any **restart** with:

```
focei$control must be a focei control object
```

Every mu-referenced or method variant — `ifocei`, `mfocei`, `foce`, `focep`,
`agq`, `laplace`, and the `*f` fast variants such as `ifoceif` — builds its
control with `foceiControl(...)` and then reclasses it to its own class
(`class(.control) <- "ifoceiControl"`, etc.). So the object carries all of
`foceiControl()`'s fields but **not** `"foceiControl"` in its class vector.

The restart path re-validates the environment with
`.nlmixrCheckFoceiEnvironment()`, whose control check was:

```r
if (!inherits(ret$control, "foceiControl")) {
  stop("focei$control must be a focei control object", call.=FALSE)
}
```

That is `FALSE` for the reclassed family controls, so a restart — e.g. after a
zero/bad-gradient theta reset — died even though the fit had been set up from a
perfectly valid control. A plain `focei` fit was unaffected (its control *is* a
`foceiControl`), which is why this stayed hidden.

## The fix

Recognise the whole family via a new `.nlmixrIsFoceiFamilyControl()` predicate,
backed by an enumerated list of the classes `foceiControl()` builds. The check
only runs on the restart path, so the blast radius is limited to the exact
failure.

## Verification

- New `tests/testthat/test-focei-family-control.R` (22 checks): the predicate
  accepts all 17 family controls and rejects `saem`/`nlme`/`nlm`; and
  `.nlmixrCheckFoceiEnvironment()` no longer rejects an `ifoceiControl` while
  still rejecting a `saemControl` — the exact assertion that failed.
- Surfaced while fitting a delay-differential-equation model with
  `est="ifoceif"`: the fit hit a theta reset and then this crash on restart.
  With the fix the same fit runs to completion.

## Note

The underlying design smell — family controls dropping `"foceiControl"` from
their class vector — is left as-is here; making them inherit `foceiControl`
would be cleaner but touches ~17 control constructors and their conversion
branches. This change fixes the user-visible crash minimally. Found alongside a
companion rxode2 fix (nlmixr2/rxode2#1138) for the same DDE workflow.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -489,6 +489,15 @@
 
 ### Estimation
 
+- A mu-referenced or method-variant FOCEi fit (`ifocei`, `mfocei`, `foce`,
+  `focep`, `agq`, `laplace`, and the `*f` fast variants such as `ifoceif`) that
+  needed to restart -- for example after a zero/bad-gradient theta reset -- died
+  with `focei$control must be a focei control object`.  These controls are all
+  built by `foceiControl()` and then reclassed to their own class, so they do
+  not carry `"foceiControl"` in their class vector, and the restart-path
+  environment check rejected them even though the fit had been set up from a
+  valid control.  The check now recognises the whole FOCEi control family.
+
 - Models that combine `linCmt()` with ODEs (for example a solved PK driving an
   effect-compartment ODE) now estimate correctly with the FOCEi and nlm
   families; the linear compartments are solved as ODEs for those methods.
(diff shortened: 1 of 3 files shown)


mattfidler:
@copilot resolve the merge conflicts in this pull request

PR #785, opened 2026-07-17
Fix AGQ quadrature node scaling (converges to the wrong limit); gate analytic covariance on cholSECov

hiddevandebeek:
## 1. AGQ quadrature node scaling

`est="agq"` scales its quadrature nodes wrongly, so adding nodes does **not** converge to the marginal likelihood — it converges to a wrong value. AGQ still improves on Laplace (~3x), which is why this was easy to miss: it looks like it's working.

**Cause.** `.agq()` builds nodes and weights from `fastGHQuad::gaussHermiteData()` with the weights divided by `sqrt(pi)` — Gauss–Hermite for the `e^{-x^2}` kernel, normalised to sum to 1. The integral being approximated has an `e^{-z'z/2}` kernel, so the node substitution is `z = sqrt(2)*x`. `inner.cpp` omitted it:

```cpp
arma::vec etaCur = etahat +  Ginv_5 * x;
lik += sum(log(w) +  0.5 * x % x);
```

so the nodes explore `N(etahat, Ht^-1/2)` instead of `N(etahat, Ht^-1)`.

**Fix.**

```cpp
arma::vec etaCur = etahat + M_SQRT2 * Ginv_5 * x;
lik += sum(log(w) + x % x);
```

No extra constant: the `2^(neta/2)` Jacobian from `dz = sqrt(2)*dx` cancels against the normalised weights. (The commented-out `//lik = 0.5*op_focei.neta * M_LN2 + ...` just below is a different formulation of the same missing factor.)

**Evidence.** Verified against references independent of any Gauss–Hermite rule — 1-D adaptive integration (one random effect) and nested adaptive quadrature (two) — with the objf constant calibrated on Gaussian-marginal models, where AGQ is exact at any nAGQ, as exactly `nobs*log(2*pi)`. Before the fix the error freezes as nAGQ grows (many nodes no better than a few); after, it converges to the exact marginal likelihood. Laplace is unchanged.

**Scope.** Every `nAGQ > 1` objective value moves, and any AGQ standard errors derived from it. `focei`/`foce`/`fo` (`_aqn == 0`) and `laplace` (`_aqn == 1`) are unaffected — the edit is inside `if (_aqn > 1)` — as is the `x = 0` node.

## 2. Analytic covariance under `cholSECov`

`covType="analytic"` did not account for `cholSECov=TRUE`. Before the covariance step, `foceiCalcCov` resets each subject's `doChol` from `cholSECov`, so `log|Ht|` then comes from the generalized (Schnabel–Eskow) Cholesky rather than `chol()`. For a non-positive-definite `Ht` that differs from what the analytic observed information differentiates, so the covariance now falls back to finite differences under `cholSECov=TRUE`. Both flags default `FALSE`, and the generalized factor equals `chol()` for a sufficiently PD `Ht`, so this changes results only for `cholSECov=TRUE` on a non-PD eta Hessian. (`cholSEOpt` governs only the optimization phase and does not affect the cov step.)

## Test baselines

Wang2007 AGQ regression baselines record self-generated values, not analytical ones (Wang 2007 gives closed forms for FO/FOCE/FOCEI, not AGQ), so they were recording the bug and are updated to the converged values. `test-focei-wang2007-basic.R` is updated here — positions 7–8 (`agq ode`/`agq`) only; all other entries unchanged.

@mattfidler — the remaining Wang2007 files (`combined`, `lognormal`, `boxcox*`, `power`, `yeojohnson`, `bounded`) are `skip_on_os("windows")` and their transform error models are platform-divergent, so I couldn't regenerate their AGQ baselines from Windows without risking values that fail on CI. Their positions 7–8 shift by the same fix. **Could you regenerate those on Linux?** CI will show which actually break — the assertion uses a mean-relative tolerance over the whole vector, so large-objf cases may stay under it.

## Also

Removes a duplicate contributor entry in `DESCRIPTION` (listed both alphabetically and appended at the end).


mattfidler:
Addressed all four Copilot review comments in a9a4f9e0:

- **inner.cpp AGQ comment**: now references `Ginv_5` (`etahat + sqrt(2)*Ginv_5*x`) and spells out that `Ginv_5 = inv(chol(Ht))`, the inverse eta-Hessian Cholesky factor `Ht^-1/2`, matching the actual computation.
- **Brittle `inner.cpp:7175` reference**: dropped the line number; the comment now points at the stable identifier `foceiCalcCov()` and its `fInd->doChol = !op_focei.cholSECov` code path.
- **Duplicated `cholSECov` gate + reason string** (both call sites): centralized into `.foceiCholSECovActive()` and `.foceiCholSECovReason`, so `.foceiCalcRanalytic` and `.foceiCovAnalyticCalc` can no longer diverge.

Verified `devtools::load_all()` recompiles and loads cleanly.

mattfidler:
Thanks @hiddevandebeek

PR #786, opened 2026-07-18
fix(vae): scalable exact covariate selection + alpha ramp + AMT/EVID data prep

mattfidler:
Fixes two defects reported by the VAE-NLME authors (Rohleff et al., CPT:PSP 2025) whose method `est="vae"` ports, and completes the reference covariate-selection ramp.

## Issue 1 -- no `AMT`/`EVID` columns crashes data prep
On dose-free data (e.g. the neonate weight dataset) `nlmixr2(..., est="vae")` failed with `Error: replacement has 0 rows, data has N`. `.vaeDataPrep` derived `EVID` from `AMT`, but when there is no `AMT` column `d$AMT` is `NULL`, so `ifelse(is.na(d$AMT) | d$AMT == 0, ...)` returned a length-0 vector. Such rows are now treated as observations (`EVID = 0`).

## Issue 2 -- covariate selection silently selects nothing at 32 covariates
The covariate M-step enumerated all `2^nCov` subsets. This is undefined behavior at `nCov = 32` (`1u << 32` masks to `1u << 0 == 1`, so only the empty model was ever tried -> no covariate selected) and intractable well before that. The Python reference uses an exact MIQP (cvxpy + GUROBI) for the same L0/BIC objective.

Replaced the enumeration with a self-contained C++ **branch-and-bound** (`vaeBestSubsetL0`) that returns the exact minimizer of the identical objective `RSS_S/omega + penalty*|S|`, using a full-remaining-fit RSS lower bound to prune. It scales to a few dozen covariates -- the `nCov = 32` case recovers the exact sparse support in ~0.03 s. A new `vaeBestSubset_` export (registered in `src/init.c`) exposes the kernel for isolated testing.

## alpha_pen warmup ramp (reference-matching, tunable)
Ported the reference covariate-selection penalty ramp: the per-covariate L0 cost multiplier ramps linearly from `covSelectAlpha` at `it=1` toward `1`, then holds at `1` for `it >= klWarmup` -- matching the reference `linspace(alpha, 1, klWarmup)` indexed 0-based by `(it-1)`. Exposed as the tunable `vaeControl(covSelectAlpha=)` (default `2`, the reference value; `klWarmup` is the ramp length). The ramp iterations -- which evaluate and print the ELBO objective -- are labeled `CovSel ramp` in the shared scale.h iteration table, with a legend entry.

## Testing (no Python/GUROBI at run time)
- `tools/vaeGenMiqpFixtures.R` + `tests/testthat/baselines/vae-miqp-golden.rds`: expected supports are the authoritative brute-force optimum of the same objective the MIQP solves.
- `test-vae-miqp.R` (fast/essential): golden match, an independent in-test brute-force cross-check on random problems, `covSelectAlpha` control tunability, and the `nCov = 32` overflow regression (exact recovery + fast).
- `test-vae-dataprep.R`: regresses the no-`AMT`/no-`EVID` path.
- `test-vae-covariate.R`: still selects `{ka, V}` end-to-end with the ramp active, and asserts the `CovSel ramp` step is shown in the iteration print.


mattfidler:
Thanks @Copilot -- addressed in 8ee76066:

- **Out-of-bounds write in `vaeBestSubsetL0`** (`out.coef[0]` on a size-0 vector): now falls back to the intercept-only model (mean of `y`) when no finite-scoring candidate is ever recorded, so `out.coef` always has at least the intercept.
- **Unchecked `omega`/`isFree` lengths in `vaeBestSubset_`**: length-1 inputs are recycled to `zDim` (mirroring the caller-side `rep_len`), and any other length mismatch now throws a clear error instead of indexing past the end.
- **`alpha_pen` ramp**: switched to `(it-1)/(klWarmup-1)`, so the multiplier is exactly `covSelectAlpha` at `it=1` and no longer collapses to `1.0` for `klWarmup=2`. This stays faithful to `linspace(alpha, 1, klWarmup)` indexed 0-based by `(it-1)`, held at `1` for `it>=klWarmup`, and needs no `klWarmup<=2` special case.

On the optional lower-bound performance note: I left the per-node full OLS solve as-is. The full-remaining-fit RSS bound is the intended design and already solves the `nCov=32` case in ~0.03 s, so the QR-downdate/Gram-caching complexity isn't warranted here.

Validation: rebuilt clean; `vaeBestSubset_` run directly against `vae-miqp-golden.rds` matches all five golden supports (incl. `nCov=32`), scalar recycling selects correctly, and a length-2 `omega` against `zDim=3` errors as intended.

mattfidler:
Second round addressed in fc4ca868:

- **Redundant leaf re-evaluation in `vaeBnbRec`**: the exclude child inherited the parent's `inSet` (already fit at the parent), so every internal node re-solved the same OLS. Added an `evalSelf` flag -- the include child evaluates its new subset, the exclude child skips the leaf. Same exact optimum, no duplicated solves. All five golden supports still match and `nCov=32` runs in ~0.07 s.
- **Hard requirement on `covSelectAlpha` in `control`**: older serialized control objects predate this field, so `as<double>(control["covSelectAlpha"])` would throw before training. Now guarded with `containsElementNamed`, defaulting a missing field to `1.0` (ramp disabled) -- which preserves the pre-ramp behavior those objects were built under.
- **Sink cleanup on error in `test-vae-covariate.R`**: replaced the trailing `sink()`/`close()` with `withr::defer({ sink(type="message"); close(tc) })` right after the sink is opened, so a failing fit can't leave an active message sink for later tests. (Used `withr::defer` rather than `on.exit`, since `on.exit` isn't reliable inside `test_that`.)

On the ramp comments: the earlier `(it-1)/(klWarmup-1)` change already makes the multiplier exactly `covSelectAlpha` at `it=1` and removes the `klWarmup=2` collapse. I've also updated the PR description so it matches the implementation -- the ramp now holds at `1` for `it >= klWarmup` (rather than reaching `1` at `it=klWarmup-1`).

PR #790, opened 2026-07-18
fix(vae): show nonMuTheta="regress" thetas in the iteration print

mattfidler:
## Summary

Two related fixes to how `est="vae"` surfaces non-mu-referenced (`nonMuTheta`) population parameters in the iteration print.

### 1. `nonMuTheta="regress"` thetas were estimated but never shown

The user reported that the non-mu thetas the VAE regresses each M-step (`vaeControl(nonMuTheta="regress")`, the default) were **not shown in the iteration print**. The regression **was** working -- the M-step re-optimizes them with a bounded `bobyqa` regression and writes the result back into the theta vector (`th[regIdx]`) -- but the parameter-history row only carried the latent-space structural thetas, the omega diagonal, and the residual error. They are now appended to the printed row (R `.vaeParRow` + C++ `parRow`, with the correct back-transform from `.iterPrintXParFromUi`), so their progress is visible in the iteration table / `parHistData`.

### 2. `nonMuTheta="fix"` only fixed the omega, not the fixed effect

While verifying (1), the `"fix"` mode was found to still **estimate** the typical-value theta (it moved off `ini`) -- only the injected eta's omega was held fixed -- so `"fix"` behaved almost like `"eta"` and the theta appeared in the print. Per the mode's name and intent, `"fix"` now holds the **fixed effect fixed too**:

- `.vaeInjectNonMuEtas(fix=TRUE)` also marks the paired theta rows `fix`.
- `.vaeDataPrep` exposes a per-latent-dim `zPopFix` flag.
- The C++ M-step (burn-in, plain, and covariate-selection branches) holds those `zPop` entries at their `ini` value.
- Fixed latent dims are dropped from the printed structural columns.

A `"fix"` parameter is now held exactly at `ini`, reported marked-fixed with the injected eta dropped, and not shown in the iteration table.

## Behavior after the change

| mode | in iteration print? | estimated? |
|------|--------------------|------------|
| `regress` (default) | yes | yes (bounded bobyqa) |
| `eta` | yes | yes (typical value + omega) |
| `fix` | **no** | **no** (held at ini) |
| `none` | no | no (frozen, no eta) |

## Testing

- `test-vae-nonmutheta.R` extended to assert the mechanism (regress thetas appear in `parHistData` and track the fit; `fix` thetas stay at `ini`, are reported fixed, and are absent from `parHistData`). 41 pass.
- `test-vae-parhist.R` (9), `test-vae-fixbounds.R` (8), `test-vae-fit.R` (11) all green; clean recompile.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -557,6 +557,13 @@
 
 ### Estimation
 
+- `est="vae"` with `nonMuTheta="regress"` now shows the regressed non-mu-referenced
+  thetas in the iteration table and parameter history.  The M-step `bobyqa`
+  regression already estimated them, but they were omitted from the printed
+  parameter walk (only the latent-space thetas, omega, and residual error were
+  shown), so their progress was invisible; they are now appended to each row with
+  the correct back-transform.
+
 - `est="vae"` covariate selection no longer silently selects nothing at 32
   candidate covariates.  The best-subset step enumerated all `2^nCov` subsets,
   which is undefined behavior at `nCov = 32` (`1u << 32` wraps to `1`, so only
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,9 +11,11 @@
   (bounds from the `ini()` lower/upper, blended with the M-step gain), recovering a
   no-random-effect population parameter without adding a spurious random effect.
   The eta-injection alternatives estimate it as `theta + mean(eta)` (the temporary
-  eta is dropped from the output model): `"eta"` estimates the injected omega,
-  `"fix"` holds it fixed at `nonMuEtaOmega`; `"none"` keeps the old freeze behavior.
-  A `$runInfo` note lists which parameters were converted.
+  eta is dropped from the output model): `"eta"` estimates the injected omega and
+  the typical value; `"fix"` holds both the injected omega AND the typical-value
+  theta fixed at their `ini()` values (nothing about the parameter is estimated, so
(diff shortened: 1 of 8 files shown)


mattfidler:
Addressed the review feedback in 372a3920:

- **`.vaeParRow` mixing values into `parInfo`** (R/vaeFit.R): `parInfo` now carries only column metadata (names/indices); the regressed-theta *values* are passed as an explicit `regressVals` argument, and the `.vaeTrain` call site passes the `ini()` starting values directly.
- **Fragile walk-vs-fit tolerance** (test): loosened `1e-3` -> `1e-2` and documented that the assertion's purpose is that the printed column IS the regressed parameter (not a stale/zero placeholder), not a bit-exact match.

Also completed the requested follow-up: `nonMuTheta="fix"` now drops the fixed parameter's **omega** column from the iteration print too (previously only the typical value was hidden). A `zPopFix` latent dim is fully fixed -- theta held at `ini`, omega fixed -- so it contributes no columns at all. Driven by a new `omegaIdx` in R with a matching skip in the C++ `parRow` omega loop; new test asserts `o(eta.lke)`/`o(eta.lV)` are absent while the real `o(eta.ka)` is still tracked.

All VAE tests green (`nonmutheta` 43, `parhist` 9, `fit` 11, `fixbounds` 8, `covariate` 18).

PR #794, opened 2026-07-18
Decouple sa/imp covariances (usable on any method) + re-default covMethod per family

mattfidler:
## Summary

Breaks the SAEM Louis stochastic-approximation FIM (`covMethod="sa"`) and the
importance-sampling Monte-Carlo observed information (`covMethod="imp"`) out of
their C++ kernels (`src/saem.cpp`, `src/imp.cpp`) so **any** estimation method can
use them, then re-defaults `covMethod` per family. Previously `sa` was tied to
`est="saem"` and `imp` to the imp/impmap family, and no other method could adopt
them.

## What changed

**Decoupled recompute engine (`R/covRecompute.R`)** computes `sa`/`imp` at a
completed fit's converged estimates:
- `.covRecomputeSa` runs a short SAEM (`nBurn=100, nEm=100`) at the pinned
  theta/omega, then the dedicated `nSaCov` FIM phase (a cold `nBurn=0/nEm=0` start
  leaves the SA running-sums uninitialized and yields a non-finite FIM).
- `.covRecomputeImp` runs the impmap kernel with the EM frozen (`nIter=1`; the
  kernel requires >= 1 iteration).
- `.covInstallResult` PD-guards the incoming covariance (a non-finite / non-PD
  matrix is **not** installed) and refreshes SE/%RSE/CI on the fit's own parameter
  table, preserving its point estimates (mirrors `.saemInstallAnalyticCov`).

**Two ways to select them:**
- **Fit time:** `covMethod="sa"/"imp"` is accepted by every control; the foreign
  request is stashed in `covMethodDeferred` and a central post-fit hook
  (`R/nlmixr2Est.R`) recomputes + installs it after the fit finalizes.
- **Post hoc:** `setCov(fit, "sa")` / `setCov(fit, "imp")` on any completed fit.

**Re-defaulted `covMethod` per family:**

| Family | Old | New |
|---|---|---|
| FOCEI (`focei`/`foce`/`laplace`/`agq`) | analytic | **r,s** |
| `vae` | analytic | **r,s** |
| `nlme` | analytic | **nlme** (keeps nlme's own vcov) |
| nonparametric (`npag`/`npb`) | analytic recompute | **imp** |
| `saem` (`sa`), imp family (`imp`), NLM (`r`), `advi` (`advi`), `fo`/`foi` (off) | -- | unchanged |

**`setCov(fit, "analytic")` no longer silently downgrades to `r,s`.** It uses the
exact analytic primitive (`.foceiCovAnalyticCalc`); when analytic cannot be
computed for the model the covariance is left **unchanged** (and never mislabeled
`"analytic"` while actually being the finite-difference sandwich).

## Implementation notes

- `foceiControl` stashes `covMethodDeferred` via the existing `...`/whitelist (no
  new formal). `saemControl` and the 8 NLM controls needed a
  `covMethodDeferred=NA_character_` **formal** because they round-trip through
  `do.call(control, list)`, which errors on unknown args. The NLM controls keep
  `covMethod="sa"/"imp"` (a valid `match.arg` choice, so it round-trips) and
  `.nlmFinalizeList` skips the native Hessian cov for them.
- The central hook reads `fit$control$covMethodDeferred` (the uniform
  `nmObjGetControl` accessor), falling back to the internal `foceiControl` for the
  families that finalize through `.foceiFamilyReturn` (vae/advi/impmap/np).

## Tests

- `tests/testthat/test-cov-decouple.R` (essential) -- family defaults, `sa`/`imp`
  accepted by every control, and the `setCov("analytic")` no-fallback guarantee.
- `tests/testthat/test-cov-decouple-saimp.R` (weekly batch 2) -- fit-time `sa`/`imp`
  across families + `setCov` round-trips.
- Updated the old "analytic is the default" assertions in
  `test-cov-analytic-defaults`, `test-cov-analytic`, `test-dde-focei`,
  `test-mfocei`, `test-nlme-cov`.

## Note for reviewers

The FOCEI default flip (analytic -> r,s) can shift reported **SE values** in any
test that fits with the default and asserts exact SEs. Every `covMethod` *label*
assertion was found and updated; the focei/cov test files do not assert
default-fit SE values. Remaining SE-tolerance churn, if any, would surface in the
**weekly** slow-batch suite.


mattfidler:
Addressed the review feedback (Copilot) in 645e297b, and merged `origin/main` into the branch so it stays merge-ready.

### Review fixes
- **`%RSE` guard (`R/covRecompute.R`)** — `.covInstallResult` now sets `%RSE` to `NA_real_` when the point estimate is `0`/non-finite instead of producing `Inf`/`NaN`, keeping the parameter table finite after a covariance install.
- **`.nlmCovMethodDefer` export (`NAMESPACE`/`R/nlmShared.R`)** — its roxygen block had merged with `.nlmFinalizeList`'s, leaking `@export` onto the internal `@noRd` helper and dropping `.nlmFinalizeList`'s man page. Split the blocks: `.nlmCovMethodDefer` is no longer exported and `.nlmFinalizeList` keeps its export + `man/dot-nlmFinalizeList.Rd`. (Ultimately `.nlmCovMethodDefer` is removed entirely — see below.)
- **Recompute hook (`R/nlmixr2Est.R`)** — the post-fit hook now triggers off the presence of `covMethodDeferred` (skipping only when a valid matching covariance is already installed) rather than comparing `fit$covMethod` to the deferred token.
- **`covMethod` docs (Copilot uobyqa note)** — the NLM controls that inherited `foceiControl`'s `covMethod` description were being documented with `r,s`/`analytic`/`sa`/`imp` choices they don't actually offer. Each NLM control now carries an accurate `covMethod` `@param`.

### Design correction: NLM family drops `sa`/`imp`
While verifying the hook fix I hit a real problem: the **NLM family (`bobyqa`/`nlm`/`nlminb`/`newuoa`/`n1qn1`/`lbfgsb3c`/`optim`/`uobyqa`) is population-only** — it rejects random-effects models. `sa` (SA-FIM) and `imp` (importance sampling) are inherently random-effects covariances, so the recompute (a nested `saem`/`imp` run) can never succeed there and errored out. The PR's own tests never exercised NLM + `sa`/`imp`.

Per maintainer decision, `sa`/`imp` are **removed from the NLM `covMethod` menus** (reverting the `covMethodDeferred` formal, the `.nlmCovMethodDefer` helper, and the finalize skip). The decoupled `sa`/`imp` remain fully available on the mixed-effects families (`focei`/`saem`/`imp`/`vae`/`advi`/`np`), which is where they're meaningful and tested.

- `NEWS.md` / `setCov()` docs now state `sa`/`imp` apply to mixed-effects fits only.
- `test-cov-decouple.R` now asserts the NLM family **rejects** `sa`/`imp`.

### Verified
- `bobyqaControl(covMethod="sa")` is rejected (menu back to `r`/`""`).
- `est="focei"` with `covMethod="sa"` and `"imp"` each install a PD, finite covariance with finite SEs.
- `test-cov-decouple.R`: 16/16 assertions pass.

PR #796, opened 2026-07-19
perf(imp): parallelize the importance-sampling covariance and E-step proposal build

mattfidler:
## Summary

Speeds up the `imp` / `impmap` / `qrpem` importance-sampling family by parallelizing two serial hotspots over subjects, **bit-identically** at any thread count. The E-step weight loop was already parallel; this adds:

- **`impComputeCov`** — the Monte-Carlo covariance (`covMethod="imp"`, the default). It is called O(np^2) times by the FD Hessian, each looping all subjects x `isample` inner solves, so it was the dominant serial cost. Now the per-subject scalar contributions are accumulated in parallel and reduced in id order.
- **E-step proposal build** — the per-subject mode / Hessian / Cholesky prologue (the serial cost for `impmap`), now parallelized (per-subject slots, no reduction).

Both reuse the E-step thread-safety contract (per-subject solve rows; serial for mixtures / multi-endpoint pool-sizing via the shared `cores` guard) and are bit-identical to the serial result. Measured ~1.7-2x on the 12-subject `theo_sd` at 4 threads, growing with subject count.

## Bit-identity

Every change reduces per-subject scalars/arrays in a fixed id order, so the floating-point sum matches the serial code exactly. Verified `identical()` on `objf`, `fixef`, `omega`, and `cov` for `imp` / `impmap` / `qrpem` (incl. `sir=TRUE`) against the pre-change build at both 1 and 4 threads.

## Tests

- New essential (non-slow) `tests/testthat/test-imp-parallel.R` asserts a full `imp` + `impmap` fit (objective, parameters, covariance) is `identical()` between 1 and 4 threads.
- Full `test-impmap.R` (158), `test-qrpem.R` (72), and the new test (8) pass, deterministically.

## Not included

The theta-score **M-step** is left serial: parallelizing its theta-sensitivity solves races on rxode2's shared solve-pool per-thread work arrays (the same limitation that already forces the E-step serial under pool-sizing), which surfaced as nondeterministic divergence only under full-suite conditions. A bit-identical parallel M-step needs rxode2-level thread-safety for `neqOverride` solves.

https://claude.ai/code/session_01GzWCpQSvXPReaXDZ3XLoGD


mattfidler:
This has issues. working on another request...

PR #808, opened 2026-07-22
feat(vae): analytic outer-gradient nonMuTheta="grad"; fix est="vae"+IOV, sigdig and unbounded-ini divergence

mattfidler:
## Summary

Adds `vaeControl(nonMuTheta = "grad")`: a structural population `theta` with no
random effect is estimated with the **exact analytic outer gradient** (the
Almquist sensitivity machinery behind `foceiControl(fast = TRUE)`) instead of the
derivative-free bounded `bobyqa` regression. Along the way this fixes several
independent defects found while validating it, including `est = "vae"` + IOV,
which was broken for every `nonMuTheta` mode.

## `nonMuTheta = "grad"`

One augmented sensitivity solve per M-step replaces the bobyqa sweep, run **in the
shared FOCEi solve pool** (the augmented model sizes the pool; the inner MAP runs
under `ind->neqOverride`, the arrangement `est="impmap"` already uses). Because
nothing is freed, `restoreFitSolve_` is gone from the VAE path.

`foceiSetup_` needed one supporting change: the augmented model declares
`THETA_1_`/`ETA_1_` (those names *are* its sensitivity directions, so it cannot be
renamed), while `params` is built as `THETA[1]`/`ETA[1]` and `rxSolve_` binds by
name. Missing pool-model names matching `THETA_n_`/`ETA_n_` now get an alias
column. Verified inert for imp/advi, whose pool model already uses the bracket
spelling.

**This is not a speed feature.** Measured on `theo_sd`, it is *slower* than
`"regress"` (1.47x with one non-mu theta, 1.13x with three; the gap narrows as the
count grows but does not close). It is chosen for accuracy: `tv` 3.4294 vs
3.4324, against a FOCEi maximum-likelihood value of 3.4293. The docs say so
plainly.

## Objective consistency

The non-mu theta M-step now optimizes the **full outer objective** (Laplace
determinant + `0.5*log|Omega^-1|` + the DV-transform Jacobian) for *both*
`"regress"` and `"grad"`, with every mu-referenced theta held at its current
M-step value. Previously `"regress"` minimized the joint likelihood at frozen
encoder etas, whose optimum is displaced. The two modes now differ only in
optimizer and are directly comparable; `"regress"` moved 3.4175 -> 3.4324.

## Fixes found while validating

* **`est = "vae"` + IOV was completely broken** (every `nonMuTheta` mode). Two
  causes, both required:
  1. the IOV **magnitude theta was never estimated** -- `.uiApplyIov` makes it a
     theta paired with per-occasion etas fixed at variance 1, which put it in
     `muRefDataFrame` so `.vaeNonMuThetas` dropped it as "mu-referenced", while
     `.vaeDataPrep` marked those etas free and forced their theta to 0;
  2. `vaeControl()`/`adviControl()` defaulted `sigdig = NULL` while
     `foceiControl()` defaults 4, and `.uiFinalizeIov` calls
     `signif(x, digits = control$sigdig)` -- `signif`'s **second** argument -- so
     NULL raised `invalid second argument of length 0`. **`advi` carried the same
     latent crash.**
* **Unbounded `ini()` divergence**: a non-mu theta with no bounds had nothing
  constraining the M-step and ran away (`tv` ~1e68 on `theo_sd`). Unbounded
  structural thetas now get the finite scale-aware fallback that covariate
  coefficients already had; a user `ini()` bound still wins, and the fallback does
  not bind (unbounded and bounded fits now agree).
* **`foceiControl(fast=)` scope doc** was stale -- it claimed "single
  additive/proportional Gaussian endpoint", but the gate is on *distribution*;
  multiple endpoints, combined/power error, both-sides transforms and a single
  estimated lambda are all in scope.

## Tests

New: `test-vae-nonmutheta-grad.R` (35), plus IOV coverage of the real
`returnVae = FALSE` user path -- `test-vae-iov.R` previously asserted only on the
raw VAE object, which is exactly how a fully broken IOV fit stayed green.

Passing: vae-nonmutheta-grad 35/35 · vae-iov 18/18 · vae-nonmutheta 43/43 ·
vae-errmodel 11/11 · focei-fast-grad 47/47 · focei-fast-methods 43/43 ·
agq-fast-grad 32/32 · impmap 158/158.

## Scope / notes for review

* IOV remains out of scope for the analytic **gradient** itself (the direction set
  is ui-level while runtime etas expand per occasion); `nonMuTheta="grad"`
  downgrades to `"regress"` with a `$runInfo` note when out of analytic scope.
* `plans/vae-outer-gradient-nonmutheta.md` carries the full design record,
  including the eliminated hypotheses -- useful context for reviewing the IOV
  fixes, whose root cause was not what the surrounding code suggested.

https://claude.ai/code/session_01KcP8o3WZYYtiVF4vGFSm4a


mattfidler:
Addressed the three review comments in caa2f1e5 — all still applied, since the intervening work was in the residual/ELS path rather than the grad plumbing they target.

- **`R/vaeGrad.R` header** — it claimed the augmented solve calls `rxSolveFree()` and that the C++ caller must `restoreFitSolve_()`. Neither is true now: `vaeOuterSolve_` runs in the shared FOCEi pool and frees nothing, which `inner.cpp` states explicitly. Rewritten to describe the pooled path.
- **`.vaeGradReset()`** — now clears all nine `.vaeGradEnv` fields instead of four. One correction: the *stale state* half doesn't apply, because `.vaeGradInit` overwrites every field including the sticky `failed`. This is memory hygiene (`data`/`ids` were retained for the session after a fit), not a correctness bug — worth being precise, since `failed` is sticky by design within a fit.
- **`.foceiAnalyticSolveAll` roxygen** — now documents the `est="vae"` pooled branch and that it's gated on an active-call flag rather than cached state, which is what keeps focei's own fast gradient unaffected.

This branch has moved a long way since the review: it now also carries the two-stage ELS residual estimator (`residOptimize=`), encoder covariate conditioning, and several reference-alignment options. A fresh pass over the newer code would be more useful than re-reviewing the above.

@copilot please review again.

PR #814, opened 2026-07-23
feat(vae): L0Learn-accelerated covariate selection for large candidate sets

mattfidler:
## Why

The `est="vae"` covariate M-step solves an exact L0/BIC best-subset problem per
latent dimension per training iteration (`vaeBestSubsetL0`, `src/inner.cpp`):

    score(S) = RSS_S / omega[k] + penalty * |S|

The exact branch-and-bound is fine at the case studies' 5 covariates and blows up
past a few dozen -- **87s for a single 30-covariate dimension**, multiplied by
`zDim * iters`. Rohleff et al. report the same wall in their own implementation and
suggested [L0Learn](https://github.com/hazimehh/L0Learn) at >= 25 covariates.

## What

`vaeControl(covSelectMethod = c("auto", "bnb", "l0learn"))` and
`vaeControl(covSelectMaxExact = 17L)`.

**L0Learn proposes, it never decides.** Supports harvested from its `L0` and `L0L2`
paths are re-scored in C++ with the same exact objective, the same OLS
(`vaeOlsRss`) and the same lexicographic tie-break the branch-and-bound uses, then
improved by an add/drop/swap local search. L0Learn's internal centering, scaling
and lambda grid therefore cannot shift a selection -- they only decide which
subsets get looked at. `omega`, the `covSelectAlpha` penalty ramp, `pinCovariates`
and the `zPop` bound clamp all keep working unchanged.

`"auto"` (the default) switches a latent dimension only when it holds at least
`covSelectMaxExact` candidates, counted **after** `pinCovariates` trimming, so the
threshold is the size of the search actually run. Below it, nothing changes.

L0Learn is a **Suggests**. When the exact search would be impractical but L0Learn
is not installed (`"auto"` at or over the threshold, or an explicit `"l0learn"`),
the fit errors rather than run the slow search silently; `covSelectMaxExact = Inf`
forces the exact branch-and-bound everywhere. Below the threshold a missing
L0Learn is fine -- the exact search runs there regardless.

A fit whose search was approximate says so in `$runInfo` and records
`fit$vae$covSelectMethodUsed`.

## Results

Parity sweep, 200 problems, `nCov` 8..18, half independent and half rho=0.7
correlated: **200/200 exact matches** against the branch-and-bound optimum.

`tools/benchVaeCovSelect.R`, N = 189, one selection call (selection identical at
every measured size):

| nCov | bnb | L0Learn+polish | speedup |
|---|---|---|---|
| 10 | 0.002s | 0.015s | 0.1x |
| 15 | 0.006s | 0.019s | 0.3x |
| 20 | 0.204s | 0.027s | 7.6x |
| 25 | 2.874s | 0.038s | **75.6x** |
| 30 | 87.393s | 0.053s | **1648.9x** |

The wall-clock crossover is near `nCov` 17, so `covSelectMaxExact` defaults to 17
(the authors suggested L0Learn at >= 25; measuring here put the crossover lower).
It is tunable, and `Inf` forces the exact search everywhere.

## Notes for review

- An `Rcpp::Function` cannot be called from inside the M-step's OpenMP loop, so
  candidate generation is **one R crossing per iteration on the main thread**,
  before the loop; scoring and polish still run in parallel.
- Candidates come back indexed into each dimension's reduced design, which is what
  the loop's `Xuse` already is, so the existing `pinCovariates` global mapping is
  reused untouched.
- `vaeBestSubsetL0` / `vaeBestSubset_` are unchanged apart from factoring result
  shaping into a shared `vaeFinishSubset`; the golden MIQP fixtures still pass.
- `src/init.c` prototype and arg count hand-added for the new `vaeScoreSupports_`
  export, per `CLAUDE.md`.

## Verification

- Full essential push/PR subset: **134/134 files, 0 failures**.
- `test-vae-l0learn.R` (essential) green **with** L0Learn (216 pass) and
  **without** it (64 pass / 5 skip).
- `test-vae-l0learn-fit.R` (weekly batch 6) asserts the mechanism --
  `covSelectMethodUsed` and the `$runInfo` note -- not just result agreement.
- Existing `vae-neonatal` / `vae-covariate` / `vae-fit` / `vae-errmodel` /
  `vae-train` / `vae-elbo` unchanged and green.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -84,7 +84,8 @@ Suggests:
     rmarkdown,
     rlang,
     numDeriv,
-    rstan
+    rstan,
+    L0Learn
 LinkingTo:
     BH,
     n1qn1 (>= 6.0.1-12),
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,8 @@
   runs one per dimension per iteration).  With the suggested `L0Learn` package
   installed, `covSelectMethod="auto"` (the default) has `L0Learn` propose
   candidate supports for any latent dimension holding at least
-  `covSelectMaxExact` (25) candidates, counted after `pinCovariates` trimming.
+  `covSelectMaxExact` (default 17, the measured wall-clock crossover) candidates,
+  counted after `pinCovariates` trimming.
   Those are candidates only: each is scored with the same exact
   `RSS/omega + penalty*|S|` objective, the same OLS and the same tie-break the
   branch-and-bound uses, then improved by an add/drop/swap local search -- so
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,8 +15,10 @@
   `RSS/omega + penalty*|S|` objective, the same OLS and the same tie-break the
   branch-and-bound uses, then improved by an add/drop/swap local search -- so
   `L0Learn`'s own objective and scaling cannot shift a selection.  Below the
-  threshold, and whenever `L0Learn` is not installed, the search stays exact and
-  unchanged.  A fit that used the approximate search says so in `$runInfo` and
+  threshold the search stays exact and unchanged.  When the exact search would be
+  impractical but `L0Learn` is not installed, the fit errors rather than run it
+  silently; `covSelectMaxExact = Inf` forces the exact branch-and-bound
+  everywhere.  A fit that used the approximate search says so in `$runInfo` and
   records it in `fit$vae$covSelectMethodUsed`.
 
 - `est="vae"` gains `vaeControl(nonMuTheta="grad")`, which estimates a structural
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,22 @@
 
 ## New features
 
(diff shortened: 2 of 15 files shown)


mattfidler:
Thanks -- addressed both in 39714919f:

1. **`covSelectMaxExact` truncation** -- `assertNumeric` (needed so `Inf` gets through) also admitted `17.9`, which `as.integer()` would truncate silently. Re-added `assertIntegerish` on the finite branch, so a fractional threshold now errors. Added `vaeControl()` validation tests covering `17.9`, `0`, `-5`, `Inf` and integer/double whole numbers.

2. **Default 25 vs 17** -- the code default is intentionally 17 (the measured wall-clock crossover; nCov 20 is already 7.6x faster). The PR description was stale from before that change -- I've updated it, and the same edit corrects the missing-package behavior (it now errors with a `covSelectMaxExact=Inf` hint rather than silently running the slow search).

PR #818, opened 2026-07-24
fix(vae): estimate declared covariate effects reached through an intermediate variable (#801)

### LINKED ISSUE #801: `vae` when using the true model (i.e covariate effects), the covariate effects become frozen

mattfidler:
## Summary

Fixes #801 -- with `est="vae"`, a declared covariate effect could be **frozen**
(not estimated) or **crash the fit** when the covariate reaches its coefficient's
model line only through an intermediate model variable, e.g.

```r
model({
  wt70 <- WT / 70
  ka <- exp(lka + beta.ka * log(wt70) + eta.ka)   # beta.ka reaches WT via wt70
  ...
})
```

The VAE `nonMuTheta` preprocessing classified covariate coefficients with its own
detector, which only looked for the data covariate appearing *literally* on the
coefficient's line. Through an intermediate it missed `beta.ka`, so `beta.ka` was
treated as a plain non-mu-referenced structural theta. The consequence depended
on `nonMuTheta`:

- `nonMuTheta="none"`: the coefficient was left **frozen** at its `ini()` value.
- `nonMuTheta="eta"`/`"fix"`: an eta was injected. The injection is a flat text
  substitution (`beta.ka` -> `beta.ka + eta.beta.ka`), correct for a bare
  structural theta (`exp(tv + eta)`) but corrupting a covariate coefficient:
  `beta.ka * log(wt70)` becomes `beta.ka + eta.beta.ka * log(wt70)`, i.e. a bare
  additive population parameter beside `lka`. rxode2 then correctly rejects it
  with `"2+ single population parameters in a single mu-referenced expression"`.
- only the default `nonMuTheta="regress"`/`"grad"` happened to estimate it (via
  the generic non-mu regression), masking the bug.

The documented guarantee is that declared covariate coefficients are estimated
**regardless of `nonMuTheta`**.

## Fix

rxode2 already recognizes `beta.ka` as a covariate coefficient in this model --
`ui$mu2RefCovariateReplaceDataFrame` lists it (the same table `.uiModifyForCovs`
folds into an `nlmixrMuDerCov#` column via the mu3/symengine path, rewriting the
line to `ka <- exp(lka + nlmixrMuDerCov1 * beta.ka + eta.ka)`). The VAE detector
was simply not consulting it.

`.vaeCovariateCoefThetas()` (`R/preProcessVaeNonMuTheta.R`) now reads rxode2's own
authoritative classification -- the union of

- `muRefCovariateDataFrame$covariateParameter` (linear `beta*WT`), and
- `mu2RefCovariateReplaceDataFrame$covariateParameter` (algebraic mu2/mu3/mu4,
  including the intermediate-variable form)

intersected with the non-fixed structural thetas. The coefficient is then
excluded from eta injection and estimated by the regress M-step in every
`nonMuTheta` mode. This also means a structural (non-random-effect) theta sharing
or following a covariate line is never mis-detected, since rxode2 does not list
it as a coefficient.

## Note (not an rxode2 bug)

The original expression `wt70 <- WT/70; ka <- exp(lka + beta.ka*log(wt70) + eta.ka)`
parses fine in rxode2 and is correctly classified. The `"2+ single population
parameters"` error came only from nlmixr2est's flat eta-injection producing a
malformed expression; rxode2's rejection of that malformed input is correct.

## Reviews

Independent local reviews (Claude, Antigravity/Gemini, Copilot/GPT). An earlier
iteration used a hand-rolled AST heuristic; reviewers flagged over-propagation
(a structural theta reclassified as a coefficient), which led to reading rxode2's
authoritative table instead -- simpler and with no over-detection.

## Tests

- `tests/testthat/test-vae-covariate-estimate.R` (fast, essential): detection
  through an intermediate var (asserting on `mu2RefCovariateReplaceDataFrame`),
  exclusion from `.vaeNonMuThetas`, regressed in every `nonMuTheta` mode, and a
  negative test (a structural intercept on a covariate line and a downstream
  structural multiplier stay non-mu structural). 70/70 pass.
- `tests/testthat/test-vae-covariate.R` (slow): end-to-end `est="vae"` with
  `nonMuTheta="eta"` on the indirect model no longer errors and estimates the
  declared coefficient (~2.4, off its 0.1 init).


mattfidler:
### mu-aware method audit (was the same bug class present elsewhere?)

Fitted the intermediate-covariate model (`wt70 <- WT/70; cl <- exp(lcl + beta.cl*log(wt70) + eta.cl)`) across the mu-referenced-aware methods to check whether any froze or crashed the coefficient:

| method | result |
|---|---|
| saem | estimated |
| saem (nonMuTheta="eta") | estimated |
| focei / foce / fo | estimated |
| nlme | estimated |
| npag / npb | estimated |
| vae / vae (nonMuTheta="eta") | estimated (this fix) |
| posthoc | 0.1 -- expected (posthoc does no estimation) |

**Only `est="vae"` had the bug.** Every other method relies on rxode2's mu2/mu3 rewrite (`.uiModifyForCovs`), which folds the intermediate covariate into an `nlmixrMuDerCov#` column via a symengine derivative check -- it handles arbitrary transforms correctly. VAE was the only method doing its own covariate-coefficient classification (in its `nonMuTheta` preprocessing) independent of rxode2's tables; the fix makes it read those tables.

### Note on the review findings (rejected)

Independent Copilot and Antigravity reviews both flagged that non-mu forms such as `cl <- exp(lcl+eta.cl)*(1 + b*WT)` are no longer detected. These are not mu-reference expressions (the coefficient sits outside the linear predictor / the derivative re-includes the coefficient), so the mu-based regress M-step could not use them anyway -- excluding them is correct, not a regression. Verified that the mu2/mu3 derivative check does classify every valid inside-predictor transform (`sqrt`, `(WT/70)^2`, `exp(WT/100)`, centered, interaction, indirect), each building its `nlmixrMuDerCov#` column; a regression test locks this in.

PR #819, opened 2026-07-24
fix: parFixed SE for residual-error parameters read uninitialized memory (#816)

### LINKED ISSUE #816: parFixed SE for residual-error parameters is uninitialized memory; $cov already holds the correct value
## Summary

For a SAEM fit, `fit$parFixed` reports the residual-error parameter's `SE` as a denormal (e.g. `9.39e-323`), while the **correct** standard error is already sitting in `fit$cov`. The value is not an underflow — it varies between identical runs with a fixed seed, which points at uninitialized memory rather than a numerical problem.

Observed on nlmixr2est 7.0.0 (rxode2 5.1.3, nlmixr2 6.0.0); `.updateParFixed()` in `R/nlmixr2output.R` is unchanged in 7.0.1.

## Reproducible example

```r
library(nlmixr2)
one.cmt <- function() {
  ini({ tka <- log(1.57); tcl <- log(2.72); tv <- log(31.5)
  model({ ka <- exp(tka+eta.ka); cl <- exp(tcl+eta.cl); v <- exp(tv+eta.v)
    d/dt(depot)  <- -ka*depot
    d/dt(center) <-  ka*depot - cl/v*center
    cp <- center/v
    cp ~ add(add.sd) })
}
fit <- nlmixr2(one.cmt, theo_sd, est = "saem", saemControl(print = 0))
fit$parFixed
```

```
       Parameter  Est.        SE      %RSE Back-transformed(95%CI) BSV(CV%)
tka           Ka 0.456     0.189      41.5       1.58 (1.09, 2.29)     70.4
tcl           Cl  1.01    0.0822      8.11       2.76 (2.35, 3.24)     27.2
tv             V  3.45    0.0432      1.25       31.6 (29.0, 34.4)     13.3
add.sd           0.697 9.39e-323 1.33e-320    0.697 (0.697, 0.697)
```

## The correct value is already computed

`fit$cov` is 7x7 and *does* include `add.sd`. Taking the square root of its diagonal gives a perfectly reasonable SE, which never reaches `parFixed`:

```r
data.frame(sqrt_diag_cov = sqrt(diag(fit$cov))[rownames(fit$parFixed)],
           parFixed_SE   = fit$parFixed[["SE"]],
           row.names     = rownames(fit$parFixed))
```

```
       sqrt_diag_cov parFixed_SE
tka       0.18947573       0.189
tcl       0.08220407      0.0822
tv        0.04323244      0.0432
add.sd    0.04708051   9.39e-323   <-- 0.0471 is available, garbage is reported
```

So this looks like a plumbing bug in how `popDf$SE` is populated for error-parameter rows, not a missing computation.

## Evidence that it is uninitialized memory, not underflow

Three consecutive runs with `set.seed(42)` and `rxode2::rxSetSeed(42)` before each fit, identical model and data:

| run | `add.sd` SE | `tka` SE |
|---|---|---|
| 1 | `9.39e-323` | 0.189 |
| 2 | `6.95e-310` | 0.189 |
| 3 | `6.95e-310` | 0.189 |

Every THETA SE is bit-stable across runs; only the error-parameter SE changes. `6.95e-310` is the classic signature of a read from an uninitialized double. A genuine underflow would be deterministic given identical inputs.

For scale: the smallest normal double is `2.2251e-308`, and `9.39e-323` is 19x the smallest representable nonzero double, retaining about 5 bits of mantissa.

## Affects more than the default covMethod

Same model, varying only the covariance method:

| `est` | `covMethod` | `add.sd` SE |
|---|---|---|
| `saem` | `"sa"` (default) | `9.39e-323` |
| `saem` | `"linFim"` | `6.95e-310` |
| `focei` | `"r,s"` | **`0.1076`** |

FOCEi produces a sane SE for the same parameter on the same data, so the estimand is well defined and computable — consistent with your point that a better estimate should be obtainable here. Both SAEM covariance paths are affected.

## Why it matters

The failure mode is quiet and easy to misread. `9.39e-323` prints as a number, not as `NA`, so `%RSE` becomes `1.33e-320` and the parameter looks as though it was estimated with absurd precision. Downstream code that filters on `%RSE` or checks `is.na(SE)` will not catch it. I ran into it while checking documentation claims and initially recorded it as "SAEM does not compute SEs" — which is wrong, and which this output actively encourages.

## Suggested handling

Ideally `parFixed` should carry the `sqrt(diag(cov))` value that already exists. Failing that, `NA_real_` would at least be honest and would make the gap visible rather than plausible.

Happy to test a patch.

mattfidler:
## Summary

Fixes #816: for SAEM fits, `fit$parFixed` reported the residual-error
parameter's `SE` as an uninitialized-memory denormal (e.g. `9.39e-323`,
varying between identical seeded runs) while the correct value was already
sitting in `fit$cov`.

## Root cause

Two layers:

1. **Out-of-bounds read in `foceiFinalizeTables` (`src/inner.cpp`).**  The
   per-theta SE vector was filled positionally from `sqrt(diag(cov))` over the
   non-skipped thetas.  SAEM installs a *structural-theta* covariance block for
   finalization (error thetas excluded; the full theta+Omega+residual matrix is
   swapped in after the fit object is built), while `foceiSkipCov` does not skip
   error thetas -- so the fill loop indexed one past the end of the diagonal and
   surfaced uninitialized memory as the residual SE (and `%RSE`).
2. **Stale formatted table.**  The post-fit covariance installs
   (`.saemInstallFullCov`, `.saemInstallAnalyticCov`, `.covInstallResult`)
   patched only the numeric `$parFixedDf`; the displayed `$parFixed` was never
   refreshed, so the garbage stayed visible even after the correct SE had been
   computed.

## Fix

- `src/inner.cpp`: the SE fill now maps by the covariance dimnames when they
  are present (an R-side cov is always named), and the positional fallback for
  the native unnamed C++ cov is bounds-guarded.  A theta with no covariance row
  gets `NA` (blank in the table) instead of garbage.  The dimnames stamping no
  longer overwrites an already-named covariance.
- `R/nlmixr2output.R`: new shared `.updateParFixedRefreshSeFromCov()` updates
  `$parFixedDf` *and* regenerates the formatted `$parFixed` from it, so the
  displayed residual `SE`, `%RSE`, and CI carry `sqrt(diag(fit$cov))`.  The CI
  is recomputed through the parameter's default back-transform
  (identity/exp/expit/probitInv); manual `backTransform` rows keep their
  existing CI.  The three install sites now share this helper instead of three
  divergent hand-patches.

## Result

```
       Parameter  Est.     SE %RSE Back-transformed(95%CI) BSV(CV%)
tka           Ka 0.456  0.189 41.5       1.58 (1.09, 2.29)     70.4
tcl           Cl  1.01 0.0822 8.11       2.76 (2.35, 3.24)     27.2
tv             V  3.45 0.0432 1.25       31.6 (29.0, 34.4)     13.3
add.sd           0.697 0.0471 6.75    0.697 (0.605, 0.789)
```

(previously `add.sd` printed `SE 9.39e-323`, `%RSE 1.33e-320`, and a
degenerate CI)

## Tests

- `test-parFixedDf.R`: unit test of the refresh helper (fill-missing,
  full-overwrite, denormal-counts-as-missing, log-scale CI recompute, FIXED
  decoration preserved).
- `test-saem-cov-sa.R`: the formatted `$parFixed` residual SE/%RSE must match
  `sqrt(diag($cov))` for both the `sa` and `linFim` covariances, and must be
  blank (never garbage) when the covariance has no row for the theta.
- `test-cov-decouple-saimp.R`, `test-cov-analytic.R`: the displayed table
  tracks the installed covariance after `setCov(fit, "sa"/"imp"/"analytic")`.

Independent reviews run before submission: GitHub Copilot (GPT family; two
test-coverage suggestions, both adopted) and Antigravity (Gemini family; one
real finding -- stale CIs for back-transformed thetas on covariance refresh --
fixed, one finding rejected as unreachable because `.foceiSetupSkipCov`
normalizes `skipCov` to theta length).

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -54,6 +54,11 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          # cache-version 2: drop caches holding qs2/stringfish binaries built
+          # against RcppParallel < 6.0.0 (pre-oneTBB); those fail to load with
+          # undefined tbb::internal symbols.  qs2/stringfish are forced to
+          # source so they always link the installed RcppParallel.
+          cache-version: 2
           extra-packages: |
             any::rcmdcheck
             nlmixr2/lbfgsb3c
@@ -64,6 +69,8 @@ jobs:
             nlmixr2/rxode2
             nlmixr2/PreciseSums
             qs=?ignore
+            qs2=?source
+            stringfish=?source
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -33,6 +33,8 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          # See R-CMD-check.yaml: qs2/stringfish must link RcppParallel >= 6.0.0
+          cache-version: 2
           extra-packages: |
             any::pkgdown
             nlmixr2/n1qn1c
@@ -43,6 +45,8 @@ jobs:
             nlmixr2/rxode2
             nlmixr2/PreciseSums
             qs=?ignore
+            qs2=?source
+            stringfish=?source
             local::.
           needs: website
 
--- a/.github/workflows/rhub.yaml
+++ b/.github/workflows/rhub.yaml
@@ -60,7 +60,11 @@ jobs:
         with:
           token: ${{ secrets.RHUB_TOKEN }}
           job-config: ${{ matrix.config.job-config }}
-          extra-packages: nlmixr2/rxode2
+          # See R-CMD-check.yaml: qs2/stringfish must link RcppParallel >= 6.0.0
+          extra-packages: |
+            nlmixr2/rxode2
+            qs2=?source
+            stringfish=?source
       - uses: r-hub/actions/run-check@v1
         with:
           token: ${{ secrets.RHUB_TOKEN }}
@@ -97,7 +101,11 @@ jobs:
         with:
           job-config: ${{ matrix.config.job-config }}
           token: ${{ secrets.RHUB_TOKEN }}
-          extra-packages: nlmixr2/rxode2
+          # See R-CMD-check.yaml: qs2/stringfish must link RcppParallel >= 6.0.0
+          extra-packages: |
+            nlmixr2/rxode2
+            qs2=?source
+            stringfish=?source
       - uses: r-hub/actions/run-check@v1
         with:
           job-config: ${{ matrix.config.job-config }}
--- a/.github/workflows/runtime-benchmarks.yaml
+++ b/.github/workflows/runtime-benchmarks.yaml
@@ -36,6 +36,8 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          # See R-CMD-check.yaml: qs2/stringfish must link RcppParallel >= 6.0.0
+          cache-version: 2
           extra-packages: |
             any::rcmdcheck
             nlmixr2/lbfgsb3c
@@ -46,6 +48,8 @@ jobs:
             nlmixr2/rxode2
             nlmixr2/PreciseSums
             qs=?ignore
+            qs2=?source
+            stringfish=?source
           needs: check
 
       - name: Install package
--- a/.github/workflows/slow-tests.yaml
+++ b/.github/workflows/slow-tests.yaml
@@ -49,6 +49,8 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
+          # See R-CMD-check.yaml: qs2/stringfish must link RcppParallel >= 6.0.0
+          cache-version: 2
           extra-packages: |
             any::rcmdcheck
             nlmixr2/lbfgsb3c
@@ -59,6 +61,8 @@ jobs:
             nlmixr2/rxode2
             nlmixr2/PreciseSums
             qs=?ignore
+            qs2=?source
+            stringfish=?source
           needs: check
 
       - uses: r-lib/actions/check-r-package@v2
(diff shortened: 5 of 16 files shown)


mattfidler:
Due-diligence note: while validating this PR, running the full weekly-batch file `test-cov-analytic.R` end-to-end stalled (100% CPU, no output, >90 min) deep into the file. Investigated whether this PR caused it:

- On this branch the full file stalled at the `covFull` fallback test (~L527); the same fit **passes in minutes when run in isolation**, and the 3-test subsequence around it passes on both trees.
- On **unmodified origin/main** the full file also stalls -- at a *different* test (~L585, non-mu-ref covariate reuse), after passing the test where the branch stalled.

So the stall is pre-existing on main, position-nondeterministic, and appears only after dozens of fits/model compiles accumulate in one R session (rxode2 model-cache pressure and/or core oversubscription are the leading suspects). It is not introduced by this PR.

Related observation: the weekly `slow-tests.yaml` run on main (2026-07-18) failed in all 8 batches at the `R CMD check` examples step ("argument is of length zero" in `nlmixr2est-Ex.R`) before any batch tests ran, so `cov-analytic` has had no green CI signal recently; both issues are worth tracking separately from this fix.

All tests directly exercising this PR pass: `parFixedDf` 19/19, `saem-cov-sa` 34/34, `cov-decouple-saimp` 20/20, and the `setCov(fit,'analytic')` + censored + covFull tests in `cov-analytic` passed before the unrelated stall point in every run.

mattfidler:
The check failures were CI infrastructure, not this PR: RcppParallel 6.0.0 (oneTBB) broke ABI for prebuilt qs2/stringfish binaries (Linux/macOS: `stringfish.so: undefined symbol ...tbb8internal...concurrent_vector_base_v3...`) and broke rxode2.dll loading on Windows (fixed upstream in nlmixr2/rxode2#1162). Merged the workflow fix from #820 (source-build qs2/stringfish + dependency-cache bump) into this branch to re-run green.

PR #832, opened 2026-07-28
Fix/focei eta reset monotone

### LINKED ISSUE #831: FOCEi objective is not a function of `theta`: inner eta-reset/nudge cascade discards converged EBEs and inflates the objective
## Summary

The FOCEi objective that the outer optimizer sees depends on hidden per-subject state (the warm-started EBEs) that `likInner()` mutates between outer iterations. The same `theta` can therefore return values far apart. With the default derivative-free outer optimizer (`bobyqa`), which builds a quadratic model by *interpolating function values*, this corrupts the interpolation model: the fit stalls, the objective rises and oscillates, and it exits with `Normal exit from bobyqa` at a point worse than one it already evaluated.

On a real (non-shareable) TMDD model this showed up as: smooth descent to OFV 527.7 by iteration 50, then a jump to ~700 with oscillation, ending at **880.47** — worse than the starting value of 847.13. Evaluating that run's own iteration-68 `theta` in a clean session gives **527.70**, i.e. the engine reported a value **+350 inflated** at a parameter vector that agreed with the iteration-50 minimum to 3–4 significant digits in every coordinate.

A fully public, self-contained reproduction is below.

## Reproducible example

Simulated data only; the ingredients are (1) four random effects, (2) true BSV above `diagOmegaBoundUpper * initial omega` so omega pins at its bound and standardized etas are inflated, (3) sparse/censored data so some subjects carry little eta information.

```r
library(nlmixr2)

simMod <- rxode2::rxode2({
  ka   <- exp(lka   + eta.ka)
  vc   <- exp(lvc   + eta.vc)
  vmax <- exp(lvmax + eta.vmax)
  km   <- exp(lkm   + eta.km)
  cp   <- centr / vc
  d/dt(depot) <- -ka * depot
  d/dt(centr) <-  ka * depot - vmax * cp / (km + cp)
})

nsub  <- 32L
doses <- rep(c(10, 50, 200), length.out = nsub)
tobs  <- c(0.5, 1, 2, 4, 8, 12, 24, 48)
ev <- do.call(rbind, lapply(seq_len(nsub), function(i) {
  rbind(data.frame(id = i, time = 0, amt = doses[i], evid = 1),
        data.frame(id = i, time = tobs, amt = 0, evid = 0))
}))
omTrue <- lotri::lotri(eta.ka + eta.vc + eta.vmax + eta.km ~
                         c(0.6, 0, 0.6, 0, 0, 0.6, 0, 0, 0, 0.6))
sim <- rxode2::rxSolve(simMod, ev,
                       params = c(lka = log(0.8), lvc = log(30),
                                  lvmax = log(15), lkm = log(2)),
                       omega = omTrue, returnType = "data.frame",
                       addDosing = TRUE, seed = 20260727)

## rxSolve(addDosing=TRUE) marks observation rows evid = 2
d <- sim[, c("id", "time", "amt", "evid", "cp")]
names(d) <- c("ID", "TIME", "AMT", "EVID", "DV")
d$EVID <- ifelse(d$EVID == 1, 1L, 0L)
set.seed(11); n <- nrow(d)
d$DV <- d$DV * (1 + rnorm(n, 0, 0.2)) + rnorm(n, 0, 0.05)
d$DV[d$EVID == 1] <- NA_real_
LLOQ <- 0.05
d$CENS <- ifelse(d$EVID == 0 & !is.na(d$DV) & d$DV < LLOQ, 1L, 0L)
d$CENS[d$EVID == 1] <- NA_integer_
d$DV[which(d$CENS == 1)] <- LLOQ

fitMod <- function() {
  ini({
    lka <- log(0.4); lvc <- log(50); lvmax <- log(8); lkm <- log(1)
    eta.ka ~ 0.1; eta.vc ~ 0.1; eta.vmax ~ 0.1; eta.km ~ 0.1
    propSd <- 0.2; addSd <- 0.05
  })
  model({
    ka   <- exp(lka   + eta.ka)
    vc   <- exp(lvc   + eta.vc)
    vmax <- exp(lvmax + eta.vmax)
    km   <- exp(lkm   + eta.km)
    cp   <- centr / vc
    d/dt(depot) <- -ka * depot
    d/dt(centr) <-  ka * depot - vmax * cp / (km + cp)
    cp ~ prop(propSd) + add(addSd)
  })
}

fit <- nlmixr2(fitMod, d, est = "focei")

## (1) the optimizer returns a point worse than one it evaluated
trace <- fit$parHistData$objf[fit$parHistData$type == "Unscaled"]
c(final = fit$objDf$OBJF[1], best = min(trace), diff = fit$objDf$OBJF[1] - min(trace))

## (2) cold-start evaluation at the fit's OWN final estimates:
##     the objective as a pure function of theta
pf <- fit$parFixedDf; th <- setNames(pf$Est, rownames(pf))
om <- diag(fit$omega); om <- om[om > 0]
cold <- nlmixr2(
  fitMod |> rxode2::ini(lka = th[["lka"]], lvc = th[["lvc"]],
                        lvmax = th[["lvmax"]], lkm = th[["lkm"]],
                        propSd = th[["propSd"]], addSd = th[["addSd"]],
                        eta.ka = om[[1]], eta.vc = om[[2]],
                        eta.vmax = om[[3]], eta.km = om[[4]]),
  d, est = "focei",
  control = list(maxOuterIterations = 0L, covMethod = "", calcTables = FALSE))
fit$objDf$OBJF[1] - cold$objDf$OBJF[1]   # inflation
```

### Observed (nlmixr2est 7.0.1)

```
iterations              : 92
best OFV seen in trace  : -33.29667
FINAL reported OFV      : -25.07155      <-- 8.2 worse than a point it visited
message                 : Normal exit from bobyqa
COLD-START OFV at theta : -33.29655
INFLATION               : +8.22          <-- objective is path dependent
```

### Expected

`FINAL` should equal `best` up to convergence tolerance, and `INFLATION` should be ~0: the objective must be a function of `theta` alone.

## Root cause

Four separate defects in `src/inner.cpp`, all in `likInner()`.

**1. The restart cascade never keeps the best (the main one).** The `doEtaNudge` ladder runs up to six `n1qn1_` restarts (warm/reset → `+etaNudge` → `-etaNudge` → `-etaNudge2` → `+etaNudge2` → `0` with `var = 0.2`). Each call **overwrites** `f` and `fInd->x`; there is no `f_best`/`eta_best` tracking. Since `LikInner2()` recomputes the individual objective at whatever eta the cascade leaves behind, a restart that lands somewhere worse silently becomes the answer.

**2. The `if (!tryAgain)` re-check guards in that cascade are unreachable.** Every one of them sits inside `if (tryAgain) { ... }`, so `!tryAgain` is always false and the re-check never runs. `tryAgain` therefore stays `true` and **all remaining restarts fire unconditionally**, ending in `if (tryAgain) { std::fill_n(fInd->x, fop->neta, 0); }` — the eta is zeroed regardless of what those five optimizations found, and `f` is not recomputed to match.

**3. `resetEtaSelective()` zeros the subject's entire eta vector.** Its own comment says "Zero every eta except protected (mu-ref-covariate) ones", but with the default `muModel = "none"` the protection array is `NULL` and it is a plain `std::fill` over all etas. One eta in its tail discards every converged EBE that subject has. Combined with a threshold of `resetEtaSize = qnorm(1 - resetEtaP/2)` applied to a *standardized* eta — which `p` = 15% of etas exceed **by construction** when `eta ~ N(0, Omega)` — the per-subject reset rate is `1 - (1-p)^k`: 39% at k=3, 48% at k=4, 73% at k=8. It gets worse the more random effects a model has.

**4. `eta1SD` is off by `sqrt(n-1)`.**

```cpp
op_focei.etaS = op_focei.etaS + (etaMat - op_focei.etaM) % (etaMat - oldM);  // Welford M2 = SS
...
op_focei.eta1SD = 1/sqrt(op_focei.etaS);                                     // needs etaS/(n-1)
```

`etaS` is the sum of squared deviations, never divided by `n` or `n-1` (its only two consumers are both `1/sqrt(etaS)`). With 27 subjects `eta1SD` is ~5.1x too small, so the `|eta_j/SD_j|` criterion effectively tests 7.3 SD instead of 1.44 SD and is inert; and when `etaS == 0` (every subject sharing an eta value, which is the state at start-up and right after a mass reset) it is `Inf` and the criterion fires for *every* nonzero eta.

Also worth noting: the reset block is gated on `!op_focei.calcGrad`, so the objective and its finite-difference gradient are computed under different rules.

## Proposed fix

Branch: `fix/focei-eta-reset-monotone` (off `main` at `163a11f8a`).

1. **Monotone restart cascade** — track `(fBest, etaBest)` and apply the best candidate at the end; on a restart returning `NA`, restore the best rather than aborting.
2. **Repair the dead `if (!tryAgain)` guards** so the cascade stops as soon as the optimizer actually moves, and drop the unconditional zero-fill (the best-of restore now owns the final eta).
3. **Per-component reset** — a per-component bound triggers a per-component reset (`resetEtaComponent()`), not a whole-vector wipe.
4. **Reset as a candidate, not an override** — when the bound fires, compare the inner objective at the reset eta against the warm one and keep the better. This preserves the safety-net purpose (escaping a genuinely bad warm start) while guaranteeing the reset can never make a subject's contribution worse. This is the change that actually restores the invariant.
5. **`eta1SD = etaSdInv(etaS, n)`** — divide by `n-1`; a zero/non-finite variance disables that criterion for the component instead of firing it unconditionally.

Regression test added at `tests/testthat/test-focei-eta-reset-path-dependence.R`, asserting both invariants above on the simulated data (a prose rule would not hold).

### Results

Public reproduction above:

| | iterations | final OFV | best in trace | final − best | cold-start at same theta | inflation |
|---|---|---|---|---|---|---|
| 7.0.1 | 92 | −25.07 | −33.30 | **+8.23** | −33.30 | **+8.22** |
| patched | 196 | **−64.21** | −64.21 | +0.0004 | −61.61 | −2.60 (PASS) |

The real TMDD model that surfaced this:

| | default | `resetEtaP = 0` workaround | patched |
|---|---|---|---|
| 4 random effects | 880.47 (ping-pongs, ends above its 847.13 start) | 232.56 | **226.41** |
| 3 random effects | 491.61 (trace looks healthy) | 383.16 | **213.69** |

Note the second row: the objective was degraded by ~278 units on a run whose iteration trace showed no sign of trouble at all, which is why this is easy to miss.

### Test suite (pristine `main` vs patched, same build, same subset)

Subset: `focei-1`, `focei-inner`, `focei-cens`, `focei-eta-only`, `fix-eta`, `etaMat`, `focei-mceta`, `focei-foce-plus`, plus the new test.

| | files | tests | pass | fail | error |
|---|---|---|---|---|---|
| `main` @ `163a11f8a` | 8 | 23 | 71 | 0 | 2 |
| patched | 9 | 24 | 72 | **2** | 2 |

The 2 errors are identical in both and pre-existing (`focei mixed effects works on matExp` / `focei matExp event sensitivities build HdEta` call `rxUiGet.foceiEtaS` / `rxUiGet.foceiHdEta` by name; these are S3 methods and are not visible under `testthat::test_dir()` — an artifact of how I invoked the suite, not of the patch).

**One open regression**, which I have not resolved:

```
test-focei-foce-plus.R:78/79 -- focep/mfocep/ifocep equal foce with foce="foce+"
  Expected `fM$objective` <= `ref$objective + 0.5`
  Actual: 118.2 > 117.1
```

`mfocep`/`ifocep` land ~1.1 units above the plain `focep` reference instead of at or below it. That fixture's own comments describe it as knife-edge (it pins `warm = "save"` because the default `warm = "calc"` already steers its multi-modal inner problem into a shallower basin at ~122.5), and it depends specifically on the mu-group regression warm-starting the inner optimizer into a *deeper* conditional mode — which is exactly the warm-start path this patch changes. I tried scoping the reset-as-candidate change to `muModel == 0` to leave the mu estimators untouched; that did **not** change the outcome, so the cause is one of the other changes (most likely the per-component reset, or the cascade now exiting early once the optimizer moves) altering the mu warm-start. Someone who knows the intent of the mfocei/ifocei re-centring loop should decide whether the deeper-mode property should be preserved explicitly rather than emerging from a whole-vector eta wipe.

## Environment

- nlmixr2est 7.0.1 (reproduced) / `main` @ `163a11f8a` (source reviewed and patched)
- rxode2 5.1.5, nlmixr2 6.0.0
- R 4.6.1, Windows 11

billdenney:
Fix #831 

@mattfidler, this needs human eyes because it makes some changes that will likely affect all focei estimations. In general, it questions eta resets and the way that retries mid-run occur. I think that this could be part of why bobyqa often performed better than normal FOCEi estimation.

- FOCEi: the inner eta-reset / eta-nudge machinery could make the objective
  function depend on the optimizer's history rather than on `theta` alone, so
  the same `theta` could return values hundreds of objective-function units
  apart.  With a derivative-free outer optimizer (the default `bobyqa`) this
  corrupts the interpolation model and the fit stalls, oscillates, and can exit
  "normally" at a point worse than one it already visited.  Fixed by:

    - Making the inner restart cascade **monotone**: each nudge restart is now a
      candidate and the best eta found is the one kept.  Previously every
      `n1qn1` restart overwrote the previous result, so the last restart won even
      when it was worse.
    - Repairing the `if (!tryAgain)` re-check guards in that cascade, which were
      unreachable (always evaluated inside `if (tryAgain)`).  Once the first
      nudge fired, every remaining restart ran unconditionally and the eta was
      then zeroed regardless of the result.
    - Making the standardized-eta reset **per component**.  A single eta in its
      tail previously zeroed the subject's entire eta vector, discarding every
      converged EBE that subject had.
    - Fixing `eta1SD`, which was computed as `1/sqrt(etaS)` where `etaS` is
      Welford's *sum of squared deviations* rather than the variance.  It is now
      divided by `n - 1`, and a zero/non-finite variance disables that criterion
      for the component instead of producing `Inf` (which made it fire for every
      nonzero eta).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,31 @@
 # nlmixr2est 7.0.2
 
+## Bug fixes
+
+- FOCEi: the inner eta-reset / eta-nudge machinery could make the objective
+  function depend on the optimizer's history rather than on `theta` alone, so
+  the same `theta` could return values hundreds of objective-function units
+  apart.  With a derivative-free outer optimizer (the default `bobyqa`) this
+  corrupts the interpolation model and the fit stalls, oscillates, and can exit
+  "normally" at a point worse than one it already visited.  Fixed by:
+
+    - Making the inner restart cascade **monotone**: each nudge restart is now a
+      candidate and the best eta found is the one kept.  Previously every
+      `n1qn1` restart overwrote the previous result, so the last restart won even
+      when it was worse.
+    - Repairing the `if (!tryAgain)` re-check guards in that cascade, which were
+      unreachable (always evaluated inside `if (tryAgain)`).  Once the first
+      nudge fired, every remaining restart ran unconditionally and the eta was
+      then zeroed regardless of the result.
+    - Making the standardized-eta reset **per component**.  A single eta in its
+      tail previously zeroed the subject's entire eta vector, discarding every
+      converged EBE that subject had.
+    - Fixing `eta1SD`, which was computed as `1/sqrt(etaS)` where `etaS` is
+      Welford's *sum of squared deviations* rather than the variance.  It is now
+      divided by `n - 1`, and a zero/non-finite variance disables that criterion
+      for the component instead of producing `Inf` (which made it fire for every
+      nonzero eta).
+
 ## Breaking changes
 
 - `est="vae"`: naming a covariate in `vaeControl(shapes=)` now also **limits
--- a/src/inner.cpp
+++ b/src/inner.cpp
@@ -277,9 +277,8 @@ struct focei_options {
   // the per-subject reset criterion).
   mat eta1SD;
   // Historical 1/sqrt(etaS) scaling, kept for thetaReset(), which standardizes
-  // the MEAN eta rather than an individual one.  resetThetaP was calibrated
-  // against this scaling, so it is preserved verbatim; see the note in
-  // thetaReset().
+  // the MEAN eta rather than an individual one (thetaReset() supplies the
+  // sqrt(n) of the standard error of the mean); see the note there.
   mat eta1SDmean;
   double n;
   double logDetOmegaInv5;
@@ -3066,17 +3065,17 @@ static inline bool thetaReset0(bool forceReset = false) {
   return true;
 }
 
-void thetaReset(double size){
+void thetaReset(double size, double n){
   if (op_focei.isSaem) return;
   if (op_focei.maxOuterIterations <= 0) return;
   if (std::isinf(size)) return;
-  // NOTE: this criterion is applied to the MEAN eta, not to an individual one,
-  // so it needs a different scale from the per-subject reset.  (The statistically
-  // correct z for "is mean(eta) != 0" would be etaM * sqrt(n) / SD.)  resetThetaP
-  // was tuned against the historical 1/sqrt(etaS) scaling, so eta1SDmean keeps
-  // that verbatim: correcting eta1SD to 1/SD without this would silently make
-  // theta resets ~sqrt(n-1) times more likely and destabilize converging fits.
-  mat etaRes =  op_focei.eta1SDmean % op_focei.etaM; //op_focei.cholOmegaInv * etaMat;
+  // NOTE: this criterion is applied to the MEAN eta, not to an
+  // individual one, so it needs a different scale from the
+  // per-subject reset: the z for "is mean(eta) != 0" carries a
+  // sqrt(n) from the standard error of the mean.  eta1SDmean keeps
+  // the 1/sqrt(etaS) scaling (etaS is the Welford sum of squares, not
+  // the variance).
+  mat etaRes = std::sqrt(n) * (op_focei.eta1SDmean % op_focei.etaM); //op_focei.cholOmegaInv * etaMat;
   double res=0;
   for (unsigned int j = etaRes.n_rows; j--;) {
     if (isMuRefCovProtected(j)) continue; // mu-ref-covariate etas never trigger a theta reset
@@ -3559,7 +3558,7 @@ void innerOpt() {
           (!op_focei.initObj || op_focei.checkTheta==1) &&
           R_FINITE(op_focei.resetThetaSize)){
         // Not thread safe...
-        thetaReset(op_focei.resetThetaSize);
+        thetaReset(op_focei.resetThetaSize, op_focei.n);
       }
       std::fill(op_focei.etaM.begin(),op_focei.etaM.end(), 0.0);
       std::fill(op_focei.etaS.begin(),op_focei.etaS.end(), 0.0);
@@ -10226,7 +10225,7 @@ Environment foceiFitCpp_(Environment e){
       // is (n - 1); etaSdInv() divides the Welford sum of squares by count-1.
       op_focei.eta1SD = etaSdInv(op_focei.etaS, n - 1.0);
       op_focei.eta1SDmean = 1/sqrt(op_focei.etaS); // thetaReset scale; see note there
-      thetaReset(op_focei.resetThetaFinalSize);
+      thetaReset(op_focei.resetThetaFinalSize, n - 1.0);
     }
   }
   e["optimTime"] = foceiElapsedSeconds(wallT0);
--- a/src/inner.cpp
+++ b/src/inner.cpp
@@ -273,7 +273,14 @@ struct focei_options {
   mat cholOmegaInv;
   mat etaM;
   mat etaS;
+  // 1/SD of the eta distribution; standardizes an INDIVIDUAL eta (etaInBound(),
+  // the per-subject reset criterion).
   mat eta1SD;
+  // Historical 1/sqrt(etaS) scaling, kept for thetaReset(), which standardizes
+  // the MEAN eta rather than an individual one.  resetThetaP was calibrated
+  // against this scaling, so it is preserved verbatim; see the note in
+  // thetaReset().
+  mat eta1SDmean;
   double n;
   double logDetOmegaInv5;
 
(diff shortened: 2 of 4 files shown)


mattfidler:
The AI agent claims that the objective function does not depend on the inner objectives' history.  This is still false.  It still depends on the inner objectives' history since it restarts at the last best known eta.  It simply will keep the better objective function per inner problem instead of stating the change

This was historically to remove the large eta deviations, but this is OK overall, though the comments are a bit wrong in places.  

I will strip the comments and make a few more changes then merge.

mattfidler:
It isn't clear that this will fix either the bobya or nlminb optimizers, though.

billdenney:
It does make the initial problem that I was working on not ping pong up and down in worse OFVs; it now estimates near monotonically to a much better OFV. That is a definite improvement. The rest was a guess on my part.

mattfidler:
The documentation will need to be updated since it is now outdated (though in all fairness, I doubt anyone reads it all that much).

PR #834, opened 2026-07-30
feat(imp): per-subject proposal adaptation for the importance-sampling EM (t proposal, AUTO, Pareto k-hat)

mattfidler:
## What this is

Per-subject adaptation of the importance-sampling proposal in `imp`/`impmap`/`qrpem`, plus the diagnostic that shows why it is needed.

## The finding that drove it

`xi` (NONMEM's `IACCEPT` quantity) and the Kish effective sample size are both **means over samples drawn from the proposal**, so neither can detect a tail the proposal rarely visits. On plain theophylline — transformably normal, 11 observations per subject — two of twelve subjects have importance weights with **infinite variance**:

| subject | Pareto k-hat | `xi` | ESS frac | verdict |
|---|---|---|---|---|
| 1 | **2.60** | 0.98 | 1.00 | unreliable |
| 5 | 0.68 | 0.96 | 0.99 | degrading |
| 10 | **1.28** | 0.96 | 0.99 | unreliable |

Not a small-sample artifact: max k-hat *grows* with more draws (2.49 → 3.31 → 3.76 at `isample` 300 → 1000 → 4000).

Three remedies, same fixture:

| approach | max k-hat | subjects k>0.7 | cost |
|---|---|---|---|
| Gaussian, `isample = 300` | 1.61 | 4 | baseline |
| Gaussian, failing subjects at `isample = 3000` | **3.28** | 1 | 10× samples |
| **t proposal, `df = 20`** | **−0.58** | **0** | 0.25% of ESS |

More samples treats the symptom and can make it worse — extra draws reach further into the tail the proposal misses. Only a heavier tail treats the cause.

## What landed

- [x] `fit$env$impXi` — NONMEM-style `xi`, exposed (no behaviour change)
- [x] Per-subject proposal-scale state (bit-identical refactor)
- [x] `gammaMethod` — NONMEM's two-sided per-subject `gamma` controller, with dimension-aware damping (`p = 2/neta`; a fixed exponent gives a period-2 limit cycle at `neta ≥ 8`)
- [x] IS covariance evaluated at the proposal the fit actually converged on
- [x] `gammaMethod="auto"` — `individual` only for non-transformably-normal models
- [x] `$runInfo` names which efficiency statistic is in force, and that the two are not comparable
- [x] `fit$env$impPsisK` — Pareto k-hat (validated against `loo::psis`)
- [x] `df` — multivariate-t proposal (NONMEM `DF`)
- [x] Per-subject `isample` (NONMEM per-subject `ISAMPLE`)
- [x] `auto` — per-subject `df`/`isample`/`iaccept` (NONMEM `AUTO=1`), **on by default**
- [x] Acceptance gate, tuned and re-run
- [x] Vignette (nlmixr2#404)

## Default change

`auto = TRUE` is the new default. Measured on theophylline against an `isample = 20000` reference (Gaussian, `df=8` and `df=20` references agree to 0.0003), 20 seeds:

| | objective RMSE | max k-hat | subjects k>0.7 | `Omega` RMSE |
|---|---|---|---|---|
| `auto = FALSE` | 0.0315 | 2.44 | 2.20 | 0.00247 |
| `auto = TRUE` | 0.0376 | **0.49** | **0.05** | **0.00198** |

Rationale: infinite-variance weights are a correctness problem whose error is unbounded in the worst case; the extra objective noise is bounded and measurable. `auto = FALSE` recovers the previous behaviour **bit-identically**.

## What is NONMEM's and what is ours

The NM7 Technical Guide's importance-sampling derivation is Gaussian throughout and never mentions `DF`. Bauer's *Tutorial Part II* documents the t sampler, `IACCEPT ≈ 0.2`, per-subject `ISAMPLE` and `AUTO=1`.

**Neither publishes the numbers.** Only the `nobs < neta` trigger and `IACCEPT ≈ 0.2` appear anywhere. So `df = 30`, the k-hat thresholds, the budget rule, the `2/neta` exponent and the 1.25× cap are **nlmixr2's own choices**, tuned on the measurements above — not reproductions of NONMEM internals. Unlike NONMEM's `AUTO`, which its tutorial warns "may result in lack of stochastic reproducibility", this stays seeded and thread-count independent.

## Verification

Every commit bit-identity checked against `main` on the `auto=FALSE`/`gammaMethod="global"` path — `identical()` on objf, theta, omega, impObj, impNeff, impCondMean, impLi and the full traces, not tolerances. Every commit independently reviewed (Antigravity/Gemini); that review caught four real defects, including the covariance step drawing Gaussian samples while weighting them as t.

**Not yet ready to merge:** the full `test-impmap.R` regression run under the new default is still outstanding. See the checklist comment.

https://claude.ai/code/session_01AX55hh6MDrfoMQHGk1pbwh

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -84,7 +84,8 @@ Suggests:
     rmarkdown,
     rlang,
     numDeriv,
-    rstan
+    rstan,
+    loo
 LinkingTo:
     BH,
     n1qn1 (>= 6.0.1-12),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,65 @@
+# nlmixr2est (development version)
+
+## New features
+
+- Importance-sampling EM (`est="imp"` / `"impmap"` / `"qrpem"`): the proposal
+  density is now adapted **per subject** rather than by one global setting, and
+  a diagnostic is reported that can tell when it matters.
+
+    - `fit$env$impPsisK` gives a **Pareto k-hat** per subject -- the tail index
+      of that subject's importance weights.  `k > 0.7` means those weights have
+      infinite variance and that subject's contribution is untrustworthy.  This
+      is worth checking because the two statistics already reported cannot
+      detect the problem: `xi` (NONMEM's `IACCEPT` quantity) and the Kish
+      effective sample size are both means over samples drawn *from* the
+      proposal, so neither sees a tail the proposal rarely visits.  On plain
+      theophylline, two of twelve subjects have k-hat of 2.60 and 1.28 while
+      `xi` reads ~0.97 and the effective-sample fraction ~0.99 for those same
+      subjects.
+
+    - `impmapControl(df=)` switches the proposal from a multivariate normal to
+      a multivariate **t** (NONMEM `DF`).  This is the remedy for a bad k-hat,
+      because it changes the proposal's *tails* rather than its width, and tail
+      weight is what decides whether the weights are well behaved.  More
+      samples does not help -- boosting a failing subject tenfold moved its
+      k-hat from 0.76 to 3.28 -- whereas `df = 20` cleared every failing
+      subject for 0.25% of the effective sample size.
+
+    - `impmapControl(isample=)` additionally accepts one count **per subject**.
+
+    - `impmapControl(gammaMethod=)` selects how the proposal scale is adapted:
+      one shared value, or per subject two-sided toward `iaccept` on that
+      subject's own `xi` (NONMEM's rule).  `"auto"`, the default, uses the
+      per-subject law only for models that are not transformably normal, since
+      `gamma = 1` is already efficient when the individual posterior is close
+      to Gaussian.
+
+    - `impmapControl(auto=)` is NONMEM's `AUTO=1`: choose `df`, `isample` and
+      `iaccept` per subject.  **It defaults to `TRUE`.**  It escalates `df`
+      only for subjects whose k-hat says they need it, leaving the rest on the
+      cheaper Gaussian, and shifts sample budget from data-rich subjects to
+      difficult ones.  Measured on theophylline against a high-accuracy
+      reference, it takes the worst k-hat from 2.44 to 0.49 and improves
+      `Omega` accuracy about 20%, for about 19% more Monte-Carlo noise on the
+      objective; infinite-variance weights are a correctness problem whose
+      error is unbounded in the worst case, while the added noise is bounded
+      and measurable.  Set `auto = FALSE` for the un-adapted behaviour, which
+      is the better choice when `fit$env$impPsisK` is already comfortably below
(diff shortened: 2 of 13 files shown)


mattfidler:
## Verification complete

**`R CMD check`: 1 NOTE**, and it is not from this branch —

```
checking compilation flags used ... NOTE
  Compilation used the following non-portable flag(s): '-mno-omit-leaf-frame-pointer'
```

That flag comes from the system `Makeconf`, not from anything here. The earlier `unstated dependencies in 'tests'` WARNING was mine (`loo::psis` used to cross-validate the k-hat implementation) and is fixed by declaring `loo` in `Suggests`.

**Test suites, all under the new `auto = TRUE` default:**

| suite | result |
|---|---|
| `test-impmap.R` (25 tests) | 163 pass, 0 fail |
| `test-qrpem.R` | 72 pass, 0 fail |
| `test-qrpem-slow.R` | 66 pass, **1 pre-existing fail** |
| `test-imp-parallel.R` | 8 pass, 0 fail |
| `test-cov-decouple-saimp.R` | 20 pass, 0 fail |
| `test-imp-xi-gamma.R` | 74 pass, 0 fail |
| `test-imp-auto.R` | 20 pass, 0 fail |
| `test-imp-tprop.R` | 27 pass, 0 fail |
| `test-imp-psis.R` | 27 pass, 0 fail |

### The one failure is pre-existing on `main`

`Q1: default (qr/sir off) fit is unchanged vs the pre-QRPEM baseline` compares against a stored baseline at `1e-8`. Only `impObj` misses. `origin/main`, built from scratch in a clean worktree, reproduces the same numbers to the digit:

| | this branch | `origin/main` |
|---|---|---|
| fixef | 2.907e-10 | 2.907e-10 |
| omega | 2.314e-10 | 2.314e-10 |
| **impObj** | **2.669e-08** | **2.669e-08** |

So this branch is bit-for-bit identical to `main` on that fit. The baseline predates five `imp.cpp` commits on `main`, three of them parallelization work that reorders summation. **Refreshing it belongs on `main`, not here** — regenerating another branch's baseline from this PR would hide drift the test is correctly reporting.

### Three `test-qrpem-slow.R` tests now pass `auto = FALSE`

Two of them found genuine interactions with the new defaults rather than merely asserting old behaviour:

- **Q6** asserts `sirSample == isample` is bit-identical to `sir = FALSE`. SIR engages on `sirN < that subject's isample`, and `auto` reallocates `isample` per subject, so the equality stops holding uniformly.
- **Q3** pins the Sobol shift *stream*; a t proposal consumes an extra uniform per draw for the chi-square scale, shifting it.

Both test qr/sir plumbing rather than `auto`, so each isolates itself — the same opt-out the other gamma-sensitive tests use.

### Merge order

nlmixr2#404 documents `auto` as defaulting on and no longer carries a version note, so **this PR should land first or together with it**.

PR #852, opened 2026-08-05
feat(saem): restore est="fsaem" (f-SAEM MAP-centered IMH kernel)

### LINKED ISSUE #845: Add an f-SAEM curvature-MAP kernel to SAEM (continuous + general-likelihood endpoints)
## Proposal

Add an optional curvature-MAP kernel to the SAEM E-step: the f-SAEM / Monolix "nlme-IMH" Laplace
independence proposal, as a 4th Metropolis kernel activated with `saemControl(nu = c(2, 2, 2, K))`. It
proposes each subject's random effects from a per-subject Laplace approximation of the individual
posterior, `N(mode_i, Gamma_i)`, and covers **both** continuous endpoints (Proposition 2) and general
log-likelihood / discrete endpoints (Proposition 1) of the f-SAEM paper.

## Current state

`main`'s SAEM E-step is the classic three-kernel gradient-free Metropolis sampler and `saemControl(nu = ...)`
is a length-3 vector (asserted `len = 3`). It mixes slowly on correlated / high-curvature posteriors, and
general log-likelihood endpoints (`ll(name) ~ <log-density>`: Poisson / binomial / negative-binomial /
time-to-event) sample with random-walk Metropolis only -- the worst-mixing option.

## Background

Karimi, Lavielle & Moulines, *f-SAEM: A fast Stochastic Approximation of the EM algorithm for nonlinear
mixed effects models*, CSDA 141:123-138 (2020), arXiv:1910.12222, gives an independent-MH proposal
`N(psi_hat_i, Gamma_i)` in two forms:

- **Proposition 2 (eq. 17), continuous data:** `Gamma_i = (J' diag(1/g^2) J + Omega^-1)^-1`, from
  linearizing the structural model (`J` = eta-sensitivity Jacobian, `g` = residual error). This is the
  Gauss-Newton / expected-information form.
- **Proposition 1 (eq. 13), general data:** `Gamma_i = (-H_{psi_hat_i} + Omega^-1)^-1`, where
  `H_{psi_hat_i}` is the **observed Hessian** of `log p(y_i | psi_i)` at the individual mode. Valid for any
  data model (continuous, categorical, count, TTE).

Eq. (18) links them: Proposition 2 is Proposition 1 with the expected information replacing the observed
Hessian (`E[-H] = J'J/sigma^2` for Gaussian data). The paper reports large mixing gains for both -- e.g.
its repeated-time-to-event (Weibull) example shows roughly an order-of-magnitude ESS improvement over RWM.

## Proposed implementation

Add the 4th kernel, with the proposal covariance built per endpoint type:

- **Continuous endpoints -> Proposition 2.** `Gamma_i = (J' diag(1/g^2) J + Omega^-1)^-1`, with `J` from
  the eta-sensitivity model (closed-form for `linCmt()`).
- **General / discrete endpoints -> Proposition 1.** The PRED solve's output column is the per-observation
  log-likelihood, so `L_s(eta) = sum_{j in s} ll_j(eta)`; build the observed Hessian
  `H_s = d^2 L_s / d eta^2` per subject (central finite differences of `L_s`, `~1 + 2*nk^2` PRED solves per
  refresh) and propose `N(mode_s, (-H_s + Omega^-1)^-1)`. Fall back to the prior precision `Omega^-1` for
  any subject whose `-H_s + Omega^-1` is not positive-definite.
- **Shared machinery:** cross-chain-mean mode; periodic H-refresh; front-loading to the early iterations
  (as in Monolix / saemix `nbiter.map`); independent-MH accept with the proposal-density correction.
- **Activation:** relax the `nu` length assertion to allow a 4th element; `nu[4] > 0` turns the kernel on
  and the endpoint distribution selects Proposition 1 vs Proposition 2 automatically. Default off (length-3
  `nu` behaves exactly as today).

## Prototype validation

theta RMSE to the simulation truth; "off" is the current RWM E-step (`nu = c(2,2,2)`), "MAP" is the new
kernel (`nu = c(2,2,2,2)`).

**Continuous:**

| model | off (RWM) | MAP |
|---|---|---|
| easy 1-cmt, N=100 (well-identified) | 0.071 | 0.071 (no regression) |
| collinear 2-cmt, N=60, reduced iterations | 0.412 | **0.356** |

On the collinear model MAP at reduced iterations also beats RWM at full iterations (~0.449), i.e. better
accuracy in fewer iterations; `linCmt()` models get the same via closed-form sensitivities.

**General / discrete** (`ll(...)` syntax, e.g. Poisson `ll(cp) ~ DV*log(lam) - lam - lgamma(DV + 1)`):

| model | off (RWM) | MAP (Prop 1) |
|---|---|---|
| random-intercept Poisson, N=80 (scalar H) | 0.023 | 0.021 |
| count-PK, N=60, 2 mu-referenced etas via ODE (multivariate H) | 0.045 / 0.080 | 0.041 / 0.078 |

Unbiased (theta ~= RWM ~= truth), scalar and multivariate observed Hessians, prior fallback exercised.
Mixing on a harder posterior (count-PK, N=60, 3 mu-referenced etas, sparse ~5 obs/subject), reduced vs full
iterations:

| | RWM, reduced | MAP, reduced | RWM, full |
|---|---|---|---|
| mean theta RMSE (3 seeds) | 0.187 | **0.102** | 0.184 |

MAP at reduced iterations matches RWM at full iterations -- the same accuracy in roughly half the
iterations.

## Follow-ups / open points

- **Analytic second-order Hessian** for the discrete branch: a `hess2`-style peer model (analytic
  `d^2 ll / d eta^2`) would give an exact `H` and remove the `~2*nk^2` extra PRED solves per refresh. FD is
  adequate for a first cut.
- **Heavy-tailed / TTE:** the paper's Remark 6 / Appendix E.1 suggest a Student-t proposal for heavy-tailed
  targets; worth evaluating on RTTE.
- **Model-syntax notes for tests/docs:** general-likelihood SAEM models must use additive mu-referencing
  (`theta + eta`); a random *slope* (`eta * covariate`) is not mu-referenceable, and time-varying rates
  should enter through an ODE state rather than an algebraic `time` term.
- Add continuous and discrete (Poisson count) SAEM examples showing `nu = c(2, 2, 2, K)`.

## References

- Karimi, Lavielle & Moulines (2020), *f-SAEM*, CSDA 141:123-138, arXiv:1910.12222 -- Prop 1 (eq. 13),
  Prop 2 (eq. 17), expected-vs-observed information (eq. 18), Remark 6 (Student proposal), RTTE example.
- Delyon, Lavielle & Moulines (1999), Ann. Statist. 27(1):94-128 -- SAEM convergence.

mattfidler:
Restores `est="fsaem"` / `saemControl(fast=TRUE)`, removed in 76a539b8f. Closes #845.

Since the first draft this has also gained the tunables #845 asked for, and
multi-endpoint support -- and one of the two "defects" the original description
claimed has been measured and turned out not to be one. See "Correction" below.

## What came back

Reverted 76a539b8f as-is: `est="fsaem"`, `fsaemControl()`,
`saemControl(fast/fastKernel/fastCov/fastIter/fastLik/nRetry)`, the C++ kernel
(`fsaemInnerMap_`, `fsaemImhKernel_`, `fsaemStepCpp_`, `fsaemMapImhCpp_` in
`inner.cpp`; the SAEM-loop IMH branch in `saem.cpp`) and the three fsaem test
files.

Two things deliberately did **not** come back, because `main` moved on:

* `.saemGeneralLik` keeps its post-removal name and home (`R/saemRxUiGetModel.R`).
  Plain `saem` fits `ll()` endpoints now, so the detector is genuinely shared and
  fsaem calls it instead of carrying a duplicate `.fsaemGeneralLik`.
* The NEWS entries stay where `main` has them. 7.0.2 is the in-flight CRAN
  submission, so this lands under 7.0.3.

## Correction -- the event-sensitivity ("jump") shape

An earlier revision of this description called this "Defect 1" and said the
inner's jump shape leaked into the SAEM solve "on every fast iteration, and from
the very first SAEM solve". **That is not what happens.** Measured by removal:
commenting out all four halves of the fix -- the three `odeSwapEsOff()` calls
(`saem_fit`, `saem_do_pred`, `fsaemRestoreSaemSolve`) *and*
`OdeSwapEsBatch(odeSlotInner)` in `fsaemMapImh` -- leaves an fsaem fit of a
`lag(depot)` model **bit-identical**: same fixef and objf to 10 digits, same
0.871389 acceptance.

Nothing on the fsaem path installs a shape in the first place.
`.fsaemInnerSetup` goes through `vaeInnerSetup_`, which does not install one
(the live active flag stays 0); focei's fit-wide `rxEventSensLoadModel` lives in
the focei *fit driver*, which fsaem never reaches. And a fit's FIRST load of the
SAEM model rebinds rxode2's event globals itself, so a shape installed
beforehand is already gone by the time `saem_fit` runs.

The guard is still worth keeping, but for a different reason than claimed: it
protects **re-entry** into an already-loaded SAEM solve while another
component's shape is live (focei's fit-wide install, or the nlm theta-gradient
path). There `rxLoad` is a no-op, so nothing else takes the shape down.

Consequences for the tests:

* The modeled-lag fit test does **not** pin this and never could. Its comment now
  says so instead of implying otherwise.
* `tests/testthat/test-fsaem-es.R` (new, and in the push/PR subset rather than
  the weekly batch) constructs the re-entry case deliberately: install the FOCEi
  inner's shape, go back through `fit$saemDopredIpred`, and assert the live flag
  goes 1 -> 0, `esDroppedN` increments by exactly 1, and the predictions are
  unchanged. With `odeSwapEsOff()` removed from `saem_do_pred` the flag stays 1
  and the counter does not move.
* It asserts on the guard's *behaviour*, not on a numeric difference: even with a
  shape live the predictions are unchanged, because the jump injection is
  compartment-count guarded.
* `.odeSwapInfo()` gained the fields this needs -- `esLive`, `esLiveNstate`,
  `esLiveNparam`, `esSlot`, `esSlotIdx`, `esOffN`, `esDroppedN`, and a per-slot
  `esActive` column. Note `rxode2EventSensShapeSize()` is the save-BUFFER size, a
  constant; the active flag from `rxode2EventSensGetDims()` is the liveness
  signal.

## Defect 2 -- the chain was rescored against the wrong state

This one is real. The IMH kernel moves `phiM`, but only the general-likelihood
branch recomputed the predictions, the observation loss and the per-subject
llik afterwards. For a normal endpoint -- where the kernel is *additive* and the
random walks still run -- `do_mcmc` was handed `U_y`/`DYF`/`fsave` from the
pre-IMH state, so every random-walk proposal was scored against a different
chain state's likelihood. The DYF rebuild is now a lambda and is re-run after
the move for every distribution.

The symptom is a **shifted answer, not an over-accepting kernel**. `do_mcmc`
writes `U_y(ind)=Uc_y(ind)` on every acceptance, so a stale entry self-heals the
first time that subject accepts anything, and the damage is diluted below noise:
with the rescore removed and the kernel firing on 40 of 45 iterations, method-2
acceptance measured 0.2369 against 0.2344 for plain saem. So the regression test
asserts `saemDiag$nRescore == fsaemDiag$nStep` and `saemDiag$uYStaleMax > 0`
(~22 nats), **not** an acceptance bound.

## Options (addressing #845's control surface)

* `saemControl(nu=)` takes an optional 4th element: the IMH sweeps per iteration
  (default 5, previously hardcoded in four places). Sweep count only -- the
  kernel is still switched on by `fast=TRUE`.
* `fastFallback="prior"` proposes from `N(centre, Omega)` for a subject with no
  usable proposal covariance, rather than holding it still. This is #845's
  Proposition-1 fallback. It is **inert in every test so far** (`nBadGamma` is 0
  throughout: `H = J' Sigma^-1 J + Omega^-1` is PD by construction), and the test
  asserts that inertness rather than pretending to exercise it.
* `fastMode="chainMean"` centres the proposal on the subject's mean over the
  chains rather than its conditional mode.
* `fastHRefresh=k` recomputes the mode and proposal covariance every `k` active
  iterations, saving an inner optimization per subject in between.

The last three trade acceptance for work or robustness, never correctness -- an
independent Metropolis-Hastings proposal need not match the target, and the
acceptance ratio is computed from the proposal actually used.

## Multiple endpoints

The fast kernel now covers any number of `add`/`prop` normal endpoints, any
number of `ll()` ones, and any **mix** of the two. It previously degraded to
standard SAEM with a message.

Almost nothing in the numerics needed changing: `likInner0` reads each
observation's own transform and variance, so the FOCEi inner information
`H = Gamma_i^-1` was already the sum over that subject's observations whatever
endpoints they belong to, and the SAEM chain already carried per-endpoint
`ares/bres/yj/lambda` indexed by `ix_endpnt`. What was single-endpoint:

* the `.fsaemSupported` gate;
* the covariate path, which rebuilt the inner THETA with `ares[1]`/`bres[1]` for
  **every** residual, so endpoint 2 was preconditioned with endpoint 1's error;
* and, for a mixed model, `distribution` itself.

That last one also fixes plain `saem`. `distribution` is ONE scalar in the
kernel and R set it to `"general"` whenever **any** endpoint was `ll()` -- which
would have scored a normal endpoint's prediction as a log density. It is now the
DEFAULT loss (`"normal"` unless every endpoint is `ll()`), and a new
per-endpoint `cfg$distEp` drives an override of the `ll()` observations' rows
after the normal block at every DYF site. The override is a no-op unless the
model actually mixes the two, so an unmixed fit stays bit-identical.

Two things worth knowing, both recorded in `plans/fsaem.md`:

* Gating `skipStochPhi0` / `localTrust` on "is mixed" is **backwards**. I tried
  it on the reasoning that an `ll()` parameter is not a `nonMuTheta="regress"`
  regressor; those guards are what keeps a phi0 in range, and disabling them let
  the `ll()` endpoint's SD reach 360 for a truth of 4.
* An SD written inside an `ll()` has no positivity constraint the way a normal
  endpoint's does, so it hits `log(negative)` and diverges. Carry it on the log
  scale.

Neither multi-endpoint fix is pinnable at fit level -- a wrong IMH proposal costs
acceptance, not correctness (measured 0.95 against 0.96 for the `ares[1]` bug) --
so `tests/testthat/test-fsaem-multi.R` pins the inner THETA directly and checks
`H` against a `numDeriv` Eq-17 with a **block-diagonal** Sigma. The two
endpoints' residual SDs differ by 10x on purpose, so using one for both is off by
~100x. The mixed model's reference is its own Gaussian twin: writing the PD
endpoint as its exact normal log-density recovers the same parameters, including
the SD that now lives inside the `ll()`.

## How the ODE pool data was used to check the solve

`.odeSwapInfo()` immediately after an fsaem fit:

```
  slot  name neq nlhs loaded sizesPool deny
1    0 inner   2   14   TRUE      TRUE    0
2    1  <NA>   0    0  FALSE     FALSE    1     (pred, thetaSens, hess2, outer...)
...
poolSlot=0 poolName=inner poolNeq=2 poolNlhs=14 opNeq=2 opNlhs=14
overrideNeeded=FALSE needsScratch=FALSE scratchNlhs=0
lhsWidthMismatchN=0 scratchUsedN=0 overrideArmedN=0 overrideNeutralizedN=0
activeOverride (want all -1): -1
```

Read across:

* `opNeq`/`opNlhs` equal the plan's `poolNeq`/`poolNlhs`, so the live solve is
  the model the registry thinks it is.
* fsaem registers exactly **one** peer, so `overrideNeeded=FALSE` and
  `scratchNlhs=0`: no `ind->neqOverride` compaction and no private lhs buffer are
  in play, and `lhsWidthMismatchN=0` confirms nothing read past a thread's lhs
  slice.
* `activeOverride` is -1 for every subject after the fit -- no leaked override.
* `scratchUsedN`/`overrideArmedN`/`overrideNeutralizedN`/`pooledSolveN` are all
  0: **the pooled-swap path is inert for fsaem.** (The shared ES shape turns out
  to be inert on this path too -- see "Correction" above. Neither shared resource
  was what fsaem was getting wrong.)

That the solve returns the right numbers is asserted directly, not inferred:

* `test-fsaem-inner.R` -- the returned MAP is a stationary point of the
  individual objective, and `H` matches an independently `numDeriv`-computed
  `J' J / sigma^2 + Omega^-1` (paper Eq 17), i.e. the eta-sensitivity columns
  read out of the solve are the right columns.
* `test-fsaem-multi.R` -- the same identity across two endpoints with a
  block-diagonal Sigma.
* `test-fsaem-imh.R` -- on a linear-Gaussian model the Laplace proposal is exact,
  so every proposal is accepted from any start (paper Remark 3); on a nonlinear
  model the sampler recovers posterior mean ~ MAP and posterior variance ~ Gamma.
* In a live fit, `fsaemDiag_()` (new) reports 20 steps / 3600 proposals /
  **87.6% accepted** / 0 MAP failures / 0 non-PD proposal covariances on theo_sd
  -- the ~0.9 the paper reports for a well-mixing independent sampler.

## About the benchmark that got it removed

The removal rationale was "no early-convergence advantage." That was measured on
theo_sd, which cannot show one: it is a well-identified 1-cmt model, 20 burn-in
iterations already put **both** methods at the MLE (RMSE ~0.01 each over six
seeds, which is SAEM's own Monte-Carlo noise floor), and the f-SAEM paper reports
exactly "no regression" for a well-identified model. The old
"converges faster ... (the point)" test was comparing two draws from the same
noise.

On the model class issue #845 actually cites -- collinear 2-cmt, N=60, sparse,
reduced iterations, RMSE to the simulation truth:

| seed | fsaem | saem |
|---|---|---|
| 1 | **0.121** | 0.194 |
| 7 | **0.089** | 0.189 |

The unit test is rewritten to assert convergence (both methods reach the MLE from
a poor start, which the kernel counters show is done *with* the IMH kernel) rather
than a stochastic inequality that flips with the seed.

## Still outside the envelope

* Mixture models: `.fsaemSupported` rejects them, and the two mixture arms of the
  E-step keep the scalar `distribution`.
* A declared off-diagonal `omega` block: the inner rebuilds Omega from the
  DIAGONAL of `Gamma2_phi1`, so the IMH would score against a different prior
  than the chain. Degrades rather than mis-targets.
* Residual kinds outside `add`/`prop` (`pow`, transform-normal, ...).
* IOV.
* The FIM's residual slot is now **dropped rather than mis-filled** -- see below.
  The remaining piece is that it is dropped rather than made per-endpoint, so
  `covMethod="fim"`/`"sa"` still report no residual SE for a multi-endpoint
  model (they already did not) and their fixed-effect SEs are conditional on the
  residual rather than marginal over it.

## The FIM's single residual slot

`nb_param = nphi1 + nlambda + 1`, so the SAEM information matrix has exactly ONE
residual slot, and `d1_logsigma2 = 0.5*resy/sigma2 - 0.5*ntotal` mixes endpoint
0's variance with whichever endpoint `resy` last held while counting EVERY
observation. That only describes a model with a single additive residual.
Anything else -- several endpoints, a proportional residual, a general
log-likelihood -- filled it anyway and inverted it alongside the parameters that
ARE reported, because `.saemFimToCov()` drops the residual column only AFTER
`solve()`.

It is now filled only in that one case, and a zero trailing row is dropped
BEFORE inverting. Dropping *before* is the load-bearing half: leaving the zero
row in makes the matrix exactly singular, so `solve()` fails and
`covMethod="fim"`/`"sa"` is lost entirely -- which is what happened to every
`ll()` model, whose slot was already zeroed. Those models gain the method
outright.

Measured rather than assumed, because the first framing of this was wrong: on a
2-endpoint PK/PD model the bogus row was worth **~4%** on the second endpoint's
theta SE (0.0561 -> 0.0539). It is **not** why that SE differs from the
linearized FIM's 0.106 -- that is a genuine difference between the
stochastic-approximation FIM and the linearization on a weakly identified phi0,
and a single additive endpoint shows the same character on its own residual
(0.0478 vs 0.0354).

`covMethod="linFim"` (the default) is untouched: `calc.COV()` indexes
`ares[ix_endpnt]` and friends and was always per-endpoint correct.

## Test batching

`test-fsaem.R` (the fit-level file) stays in weekly `.slowBatches` batch 2.
`test-fsaem-es.R`, `test-fsaem-multi.R`, `test-fsaem-inner.R` and
`test-fsaem-imh.R` are unit-level and run in the essential push/PR subset -- the
two regression pins in particular needed push/PR coverage rather than weekly.


mattfidler:
Pushed the issue #845 follow-ups: `nu[4]`, a fallback estimator, cross-chain-mean mode, periodic H-refresh, and the acceptance instrumentation. All four options are opt-in and the default path is bit-identical (theo_sd `saem`/`fsaem` `fixef()` equal to the pre-change branch at zero tolerance).

**New options** (all `saemControl()`):

| option | default | effect |
|---|---|---|
| `nu = c(2,2,2,K)` | length 3 | `nu[4]` is the IMH sweeps per iteration (was hardcoded 5 in four places). Sweep count only -- it does not switch the kernel on |
| `fastFallback` | `"skip"` | `"prior"` proposes from `N(centre, Omega)` for a subject whose own proposal covariance is unusable |
| `fastMode` | `"map"` | `"chainMean"` centres the proposal on the subject's mean over the chains |
| `fastHRefresh` | `1L` | recompute the mode + covariance every `k` active iterations, reusing them in between |

Measured on theo_sd (20 fast iterations, all reaching the same MLE):

```
default          nStep 20  reuse  0  accRate 0.882
chainMean        nStep 20  reuse  0  accRate 0.594
fastHRefresh=5   nStep 20  reuse 16  accRate 0.671
both             nStep 20  reuse 16  accRate 0.458
```

The last three trade acceptance for work or robustness, never correctness: an independent MH proposal need not equal the target, and the ratio is computed from the proposal actually used. What `fastHRefresh` must **not** skip -- and does not -- is the per-iteration re-parameterization, since `likInner0` has to score at the current theta/omega.

**Two places the work corrected its own premise, both worth a look:**

1. **The over-acceptance bug does not present as over-acceptance.** `do_mcmc` writes `U_y(ind)=Uc_y(ind)` on every acceptance, so a stale entry self-heals the first time that subject accepts anything. With the rescore removed and the kernel firing on 40 of 45 iterations, method-2 acceptance was 0.2369 against 0.2344 for plain saem -- noise. The real symptom is a shifted answer (tka 0.4362 vs 0.4620). So the regression test asserts `nRescore == nStep` and `uYStaleMax > 0` (measured ~22 nats), not an acceptance bound. The reasoning is recorded next to the counters so this is not re-derived wrongly later.

2. **The prior fallback is not reachable end to end.** `H = J' Sigma^-1 J + Omega^-1` is PD by construction, so no healthy fit produces a bad Gamma -- 0 on the exponential-TTE general likelihood (which forces the exact FD Hessian), on 2-obs-per-subject theo_sd, and on the same with `fastCov="hessian"`. Driving the prior term to zero fails earlier in SAEM's own nearPD. The test asserts the option is **inert** rather than pretending to exercise the rescue.

**Also fixed along the way:** the `saemControl(mcmc=)` passthrough validated `mcmc$nu` but never assigned it (unlike `niter`/`nmc` beside it); the covariate fast path could not honor any sweep count at all; and a fourth way to lose a subject -- a finite Cholesky that will not itself invert -- was uncounted.

**Diagnostics** now ride on the fit rather than a `:::` global a later fit overwrites: `fit$saemDiag` (per-kernel Metropolis proposals/acceptances, plus `nRescore`/`uYStaleMax`) and `fit$fsaemDiag` (steps, proposals, acceptances, MAP failures, non-PD covariances, cache reuse, prior rescues).

Tests: `test-fsaem` (15 blocks), `test-fsaem-inner`, `test-fsaem-imh`, `test-saem-loglik`, `test-cov-analytic-defaults`, `test-saem-cov-analytic`, `test-mu-timevarying`, `test-single-subject` -- all pass.

mattfidler:
Ran antigravity (Gemini) and Copilot (GPT) over this branch three times each, fixing findings between rounds. The loop converged: **round 3 antigravity found nothing new and verified all seven prior fixes**; Copilot went 4 -> 5 -> 2 findings.

**Ten real bugs found and fixed**, plus an eleventh neither reviewer reported.

Round 1: a zero-valued time-varying covariate beta was treated as "Plambda not populated" (inner kept ini values while the chain used 0); `fastLik` was a complete no-op on the Hessian path (`foceiControl(foce=)` is ignored when interaction is on -- the inverse of its own roxygen); the `fastHRefresh` cache guard only checked the cache was non-empty, permitting a silent out-of-bounds read and write. **That new guard immediately caught a bug neither reviewer saw**: `.fsaemImh` built `cholGamma` with `t(vapply(...))`, which returns a vector when `neta == 1`, so the matrix was transposed -- invisible because a single column indexes identically either way.

Round 2: a **correlated omega** was silently mis-targeted -- the inner rebuilds Omega from the diagonal only, so the IMH scored against a different prior than the SAEM chain, i.e. the composed kernel did not target the chain's distribution (now rejected by `.fsaemSupported`); `ll()` endpoints ran with interaction on where `focei.R` turns it off for any `needOptimHess` model; on retry exhaustion a bounded proposal was **clamped but scored with the unclamped Gaussian draw**, so the acceptance ratio described a point that was not the one proposed (now abandoned -- an exact Metropolis outcome -- and counted as `nBoundFail`); `lower`/`upper` were indexed without a size check.

Round 3: a **`fix()`ed structural theta** broke the entire inner theta vector. A fixed theta is not in phi1/phi0, so counting it made the structural positions exceed `nphi1 + nphi0`, tripped the C++ length guard, and filled *every* structural slot from the fallback -- while the fixed parameter itself sat at 0 rather than its own value. Measured after the fix on theo_sd with `tcl <- fix(1)`: acceptance 0.935, `tcl` held at exactly 1.0000, 0.0024 from the SAEM estimates. Also, a partial `nbd` silently disabled the bound/retry path entirely.

**One finding rejected, and worth recording.** Antigravity argued `phiPop` is read positionally while the phi vector follows the ETA order, with a concrete trigger -- and `ui$muRefDataFrame` genuinely does disagree with `ini()`. I implemented it and measured: acceptance **collapsed from 0.882 to 0.016**. The phi vector follows the **ini (ntheta)** order and the positional read was correct all along. Reverted, pinned with a permuted-`ini` regression test, and the reasoning recorded at both sites so it does not get "fixed" again.

Tests: 22 blocks / 149 assertions across `test-fsaem`, `test-fsaem-imh`, `test-fsaem-inner`, `test-saem-loglik` -- **0 failures, 0 errors**. New regression tests cover the time-varying covariate fit, each `fastLik` value, the permuted-`ini` ordering, the correlated-omega gate, and the fixed structural theta.

mattfidler:
@hiddevandebeek I have also added the differences in our algorithms as a optional opt-in or opt-out.

However, I still don't know what you have so perhaps it would be good to check and/or cross check between the two.

hiddevandebeek:
I benchmarked this branch on the paper's own warfarin data
(`belhal/saemixmcmc/data/warfarin_data.txt`, same ka/V/k parameterisation and `psi0 = (1, 7, 1)`).
Checked on f7fba0fe4; a re-run of one cell against 90f326b12 was bit-identical, so this describes
the current defaults.

Two things to watch out for before any of the numbers mean anything.

**A FOCEi fit is not a usable reference here.** The two estimators converge to different points --
FOCEi gives `tka=-0.616`, `saem` at 500 iterations gives `tka=-0.477`. That 0.139 gap is bigger
than any effect worth measuring, and it sits where you would expect (`omega.ka` around 0.7, where
Laplace is weakest). Against FOCEi, plain `saem` appears to get worse between 200 and 500
iterations when it is simply converging to its own answer, and `fastIter` appears to have a sharp
optimum that it does not have. I lost most of a day to this.

**`handleUninformativeEtas` (on by default) confounds `saem`-vs-`fsaem`.** Its test runs once at
the initial estimates against an absolute tolerance, so poor starting values that underflow the
predictions freeze etas for the whole fit -- from the paper's `psi0`, the volume eta for 19 of 32
subjects. The kernel then gets credit for compensating. #865 fixes it; worth rebasing before
benchmarking.

Distance of `(tka, tv, tk, add.sd)` to the converged `saem` solution, 4 seeds, `fastKernel="firstN"`:

| total iterations | `saem` | k=5 | k=10 | k=20 |
|---|---|---|---|---|
| 14 | 0.571 | 0.164 | 0.125 | 0.129 |
| 25 | 0.256 | 0.095 | 0.092 | 0.094 |
| 50 | 0.156 | 0.138 | 0.116 | -- |
| 100 | 0.117 | 0.119 | 0.119 | 0.098 |
| 200 | 0.120 | 0.070 | 0.082 | 0.094 |
| 500 | 0.036 | 0.030 | 0.043 | 0.061 |

Paired on matched seeds: t = 3.1-3.4 at 14 iterations, ~1.8 at 25, nothing at 50-100. The payoff
is large but concentrated in the first ~25 iterations. `fastIter` barely matters -- 5, 10 and 20
are within noise at every budget. And every setting lands on the same point by 500, within the
reference's own 0.036 of seed scatter, so the kernel is an accelerator rather than a different
estimator. That last point seems worth stating in the docs, since it is what makes it safe to
leave on.

## fastMode="chainMean"

I ran your `chainMean` against `map` on the same seeds and budgets, `fastIter=20`:

| total iterations | `map` | `chainMean` | paired diff | t |
|---|---|---|---|---|
| 14 | 0.129 | 0.211 | -0.082 +- 0.047 | -3.50 |
| 25 | 0.094 | 0.104 | -0.010 +- 0.031 | -0.66 |
| 50 | 0.136 | 0.092 | +0.044 +- 0.063 | +1.38 |
| 100 | 0.098 | 0.112 | -0.014 +- 0.029 | -0.94 |

So it is worse at the shortest budget and indistinguishable elsewhere. Acceptance drops from
0.67-0.74 to 0.50-0.57, which I think is the clue: the centre moves to the chain mean but Gamma
still comes from the MAP's H, so the proposal's location and its shape describe different points.

Before you had this in, I had hacked the same idea into an older build but *also* evaluated the
curvature at the mean (`maxInnerIterations = 0`), and that went the other way -- 0.084 at 14
iterations against 0.129 for `map`, on these same seeds and reference. Small sample, and not your
code, so treat it as a hint rather than a result. But it does suggest the centre and the curvature
point have to move together for this to help, and that moving only the centre is the worst of the
three.

## The MAP search cold-starts every iteration

`inner.cpp:16371` sets `env["etaMat"]` to zeros, so `innerOpt1` restarts from the prior mean each
fast iteration. On a simulated Michaelis-Menten model (4 etas, N=60, known truth) that search
fails for a large minority of subjects, and `nBadGamma` tracks it exactly:

| seed | `mapFail` as shipped | seeded from the chain |
|---|---|---|
| 1 | 31 / 60 | 0 |
| 2 | 21 / 60 | 2 |
| 3 | 44 / 60 | 0 |
| 4 | 46 / 60 | 0 |

Both reference implementations warm-start (`belhal/saemixmcmc` `R/mcmc.R:91-93`, saemix
`estep:113-115`), and Algorithm 2 specifies what to compute rather than where the optimizer
starts. But warm-starting everywhere is not free -- on a collinear 2-cmt model it was slightly
worse on 4/4 seeds where `mapFail` was already 0. So the narrow suggestion is to keep the cold
start and retry from the chain state only when the cold optimization fails.

Two small ones while I was in there: `fastKernel="throughout"` and `"additive"` still evaluate the
same expression (`saem.cpp:1637-1638`), and `saem.cpp:1632` still says the live kernel is not yet
wired with a dead `(void)fsaemActive` below it.

All of this is one dataset with 4 seeds per cell, so anything below ~0.05 in the long-budget rows
is not resolvable. Happy to run anything else that would be useful.

hiddevandebeek:
So I guess its advantageous in extremely short runs. But yeah like you said, its not magic. Not really sure whether its the most useful thing, but its sort of nice to have (?). 

Il look into the code a bit what the differences are

mattfidler:
Maybe; @billdenney  requested it years ago; it isn't found commonly (or at all) in the literature so it may be useful... but maybe not.

billdenney:
Out of the list here, I'm not exactly sure which part I requested. (I also realize that I request a lot of different features because I do a lot of ~weird~ unique modeling.)

If it is about generalized likelihood for SAEM, I use generalized likelihoods relatively often for things like Markov models (relatively often means several times a year) or making combined distributions (e.g. a zero-inflated exponential distribution for a Poisson process). I tend to prefer SAEM over focei for confidence interval accuracy (when HMC/NUTS is not available).

If not that, yes, I probably still asked for it and can provide rationale. 😄

mattfidler:
likelihood in `saem` is avaialble, but experimental.  

Here is a walk down memory lane https://github.com/nlmixrdevelopment/nlmixr/issues/246

PR #853, opened 2026-08-05
fix(cov): analytic covariance ignored covSolveTol and ran single-threaded

hiddevandebeek:
Two independent bugs in the analytic covariance's pooled solve. They were found
together because the first one's benchmark exposed the second.

## 1. The pooled route ignored `covSolveTol`

The augmented solves behind the analytic R matrix are central-differenced twice and
Richardson extrapolated to recover the 3rd-order tensor, so the solve error **is** the
answer's error. They are run at `solveTol` -- `foceiControl(covSolveTol=)` if set, else
tightened from `sigdig` by `.foceiAnalyticSolveTol` (1e-9 at the default `sigdig = 3`).
`.foceiAnalyticSolveAll` honoured that only on its `rxode2::rxSolve` fallback; the pooled
route dropped it, and `vaeOuterSolve_`'s `OdeFitTolGuard` **reset** the live tolerances to
the fit's -- 1e-3 at the default.

`OdeFitTolGuard` is right for a *gradient*: it must be the gradient of the objective the
optimizer is minimising. But that caller no longer exists -- the outer gradient has been
all-C++ since the R core was deleted, so every caller of `.foceiAnalyticSolveAll` is in
`R/foceiCovAnalytic.R`, and `est="vae"`'s M-step goes through `foceiGradPooledDirect_`.
`vaeOuterSolve_` is the covariance's entry point and nothing else's.

So it takes a tolerance and picks the guard -- `OdeSolveTolGuard` when supplied,
`OdeFitTolGuard` otherwise (pass `NA`), so a future gradient caller keeps today's
behaviour. `OdeSolveTolGuard` already existed for exactly this and was unused.

**Two user-visible consequences.** SEs carried the fit's solve error rather than the
covariance's; and because the pool is only available with `fast = TRUE`,
**`foceiControl(fast=)` changed the standard errors**. A user who explicitly set
`covSolveTol` had it ignored.

Holding the fit fixed (`maxOuterIterations = 0`) so only the solve tolerance moves, max
relative SE difference between the pooled route and the rxSolve route that always
honoured `solveTol`:

| sigdig | fit rtol | before | after |
|---|---|---|---|
| 3 | 1e-3 | 1.73e-4 | **0** |
| 4 | 1e-4 | 1.15e-5 | **0** |
| 6 | 1e-6 | 8.38e-8 | **0** |

The gap tracking the fit tolerance across three decades is what identified the cause; it
is now identically zero at every one. With only `foceiControl(fast=)` differing at a fixed
theta on a 5-ETA 2-compartment model: **1.71e-2 -> 3.88e-6**, and `fast=TRUE` moved *onto*
the tight values rather than merely nearer them (`tv2` 0.31483354 against 0.31483476,
previously 0.32022658).

## 2. The pooled solve ran single-threaded

Found while benchmarking the above -- the pooled route was measurably *slower* than the
rxSolve route it replaced, which it structurally should never be.

```r
pooled    if (is.null(.c) || is.na(.c) || .c < 1L) 1L else as.integer(.c)
rxSolve   if (is.null(am$cores)) 0L else am$cores
```

`0` is `rxControl`'s default and means "use rxode2's thread setting". rxSolve passes it
through and gets a threaded solve; the pooled branch turned it into a literal `1`, and

```cpp
const bool doParallel = (cores > 1) && solveMethodThreadSafe(op);
```

so `outerSolveFill`'s OpenMP-over-subjects loop never ran. The pool's parallelism -- the
reason that function exists -- was off by default here.

## Measurements

All timings call `.foceiCovAnalyticCalc()` repeatedly on **one** fitted object, 8
alternating reps. Differencing whole fits (my first approach) cannot measure this: fit
variance is seconds against a sub-second step, and it produced *negative* covariance
costs.

| | pooled | rxSolve | ratio |
|---|---|---|---|
| 3 ETA, 1-cmt | 0.330 -> **0.260s** | 0.340s | 1.00 -> **0.76** |
| 5 ETA, 2-cmt | 1.040 -> **0.710s** | 0.845s | 1.20 -> **0.84** |

The 5-ETA "before" was non-overlapping, i.e. the pool really was slower. It is now faster
on both.

Cost of honouring `solveTol`, isolated the same way: **+0.010s** (3 ETA) and **+0.030s**
(5 ETA), 1.04x with overlapping samples -- at the noise floor, and far smaller than the
threading fix returns.

The threading fix is numerically inert: per-subject solves are independent. The
pooled-vs-rxSolve SE gap stays `0.000e+00` at sigdig 3/4/6 and the `fast=` difference
stays `3.877e-06`, bit for bit, before and after.

## Tests

New regression test in `test-cov-analytic.R` pins `fast=`-invariance of the analytic SEs
directly. Also corrects a stale comment in `test-vae-nonmutheta-grad.R` describing
`.foceiAnalyticSolveAll` as shared with the fast gradient.

**Not run here:** the full `test-cov-analytic.R` file. It is weekly-batched and
compile-bound (a distinct augmented model per direction set), and did not complete in this
environment. It should be run before merge -- its own header states the analytic augmented
solve runs at `10^-(sigdig+6)`, which is exactly the premise the pooled path was violating,
so it is the file most likely to have something to say.

## Registration

`vaeOuterSolve_` gains an argument, so `src/init.c`'s forward prototype and its
`callMethods[]` arity go 4 -> 5 alongside `Rcpp::compileAttributes()`, per the manual
registration convention in CLAUDE.md.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,13 @@
   as a result -- measured at about 4% of that step, which is inside its
   run-to-run noise; set `covSolveTol` to trade accuracy back for speed.
 
+- `covMethod="analytic"` solves in parallel again.  The pooled covariance solve
+  coerced `rxControl(cores = 0)` -- the default, which means "use rxode2's thread
+  setting" -- to a literal 1, so it ran **single-threaded over subjects** while the
+  `rxSolve` route it replaced honoured the 0 and ran threaded.  The shared pool's
+  parallelism was therefore off by default here.  A 5-ETA 2-compartment covariance
+  goes from 1.04s to 0.71s, and from slower than the unpooled route to faster.
+
 - Fixed the objective function for a model that has a **general-likelihood
   endpoint (`ll()`, `pois()`, `binom()`, ...) alongside any other endpoint**.
   Each observation's distribution was read one row before the model had been
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,13 @@
   at the default `sigdig`).  Both routes now agree exactly.  Set `covSolveTol` to
   trade accuracy back for the slightly larger covariance step.
 
+- Fixed a subject whose pooled augmented solve failed being scored into
+  `covMethod="analytic"` as zeros -- no prediction and no sensitivity -- instead
+  of sending the covariance to its fallback.  The zero fill was written for the
+  R outer gradient, which replaced such a subject's column by a finite
+  difference; that gradient is gone, and zeros are finite, so nothing downstream
+  noticed.  Such a population now falls back to the unpooled solve.
+
 - Fixed the pooled `covMethod="analytic"` solve running single-threaded.  It
   coerced `rxControl(cores = 0)` -- the default, meaning "use rxode2's thread
   setting" -- to a literal 1, so its loop over subjects never went parallel,
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,20 @@
 
 ### Estimation
 
+- `covMethod="analytic"` now solves at the tolerance it asks for.  The augmented
+  solves behind the analytic R matrix are central-differenced twice and Richardson
+  extrapolated to recover a 3rd-order tensor, so they are run at
+  `foceiControl(covSolveTol=)` -- or, unset, a value tightened from `sigdig`.
+  Whenever the shared ODE solve pool was available that tolerance was dropped and
+  the fit's own, much looser, tolerance was used instead.  Two consequences, both
+  gone: standard errors carried the fit's solve error rather than the
+  covariance's (1.7e-2 relative on a 5-ETA 2-compartment model at the default
+  `sigdig`), and since the pool is only available with `fast = TRUE`,
+  **`foceiControl(fast=)` changed the standard errors**.  The pooled and
+  unpooled routes now agree exactly.  The covariance step does slightly more work
+  as a result -- measured at about 4% of that step, which is inside its
+  run-to-run noise; set `covSolveTol` to trade accuracy back for speed.
+
 - Fixed the objective function for a model that has a **general-likelihood
   endpoint (`ll()`, `pois()`, `binom()`, ...) alongside any other endpoint**.
   Each observation's distribution was read one row before the model had been
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,26 +12,21 @@
 
 ### Estimation
 
-- `covMethod="analytic"` now solves at the tolerance it asks for.  The augmented
-  solves behind the analytic R matrix are central-differenced twice and Richardson
-  extrapolated to recover a 3rd-order tensor, so they are run at
-  `foceiControl(covSolveTol=)` -- or, unset, a value tightened from `sigdig`.
-  Whenever the shared ODE solve pool was available that tolerance was dropped and
-  the fit's own, much looser, tolerance was used instead.  Two consequences, both
-  gone: standard errors carried the fit's solve error rather than the
-  covariance's (1.7e-2 relative on a 5-ETA 2-compartment model at the default
-  `sigdig`), and since the pool is only available with `fast = TRUE`,
-  **`foceiControl(fast=)` changed the standard errors**.  The pooled and
-  unpooled routes now agree exactly.  The covariance step does slightly more work
-  as a result -- measured at about 4% of that step, which is inside its
-  run-to-run noise; set `covSolveTol` to trade accuracy back for speed.
+- Fixed `covMethod="analytic"` ignoring its own solve tolerance whenever the
+  shared ODE solve pool was available, solving at the fit's much looser tolerance
+  instead of `foceiControl(covSolveTol=)` (or, unset, a value tightened from
+  `sigdig`).  The augmented solves are differenced twice to recover a 3rd-order
+  tensor, so their error is the standard errors' error: they carried the fit's
+  instead, and because the pool needs `fast = TRUE`, **`foceiControl(fast=)`
+  changed the standard errors** (1.7e-2 relative on a 5-ETA 2-compartment model
+  at the default `sigdig`).  Both routes now agree exactly.  Set `covSolveTol` to
+  trade accuracy back for the slightly larger covariance step.
 
-- `covMethod="analytic"` solves in parallel again.  The pooled covariance solve
-  coerced `rxControl(cores = 0)` -- the default, which means "use rxode2's thread
-  setting" -- to a literal 1, so it ran **single-threaded over subjects** while the
-  `rxSolve` route it replaced honoured the 0 and ran threaded.  The shared pool's
-  parallelism was therefore off by default here.  A 5-ETA 2-compartment covariance
-  goes from 1.04s to 0.71s, and from slower than the unpooled route to faster.
+- Fixed the pooled `covMethod="analytic"` solve running single-threaded.  It
+  coerced `rxControl(cores = 0)` -- the default, meaning "use rxode2's thread
+  setting" -- to a literal 1, so its loop over subjects never went parallel,
+  while the `rxSolve` route it replaced passed the 0 through and did.  A 5-ETA
+  2-compartment covariance goes from 1.04s to 0.71s.
 
 - Fixed the objective function for a model that has a **general-likelihood
   endpoint (`ll()`, `pois()`, `binom()`, ...) alongside any other endpoint**.
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,12 @@
   at the default `sigdig`).  Both routes now agree exactly.  Set `covSolveTol` to
   trade accuracy back for the slightly larger covariance step.
 
+- Fixed `foceiControl(covSolveTol=)` being dropped part-way through the
+  covariance step.  Once the analytic route had restored the fit's ODE solve --
+  which it does whether it succeeded or declined -- the finite-difference
+  covariance work after that point ran at the fit's tolerance again, because
+  rebuilding the solve resets the tolerances along with it.
+
 - Fixed a subject whose pooled augmented solve failed being scored into
   `covMethod="analytic"` as zeros -- no prediction and no sensitivity -- instead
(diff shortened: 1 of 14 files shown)


mattfidler:
Went through the Copilot comments and then re-reviewed the branch, with independent
second-opinion passes from Antigravity (Gemini) and the Copilot CLI (GPT).

### Copilot's three comments

Both NEWS bullets now lead with "Fixed" like their neighbours, and lost the
root-cause narration.  The comment block that asserted which callers exist was
replaced by the tolerance *contract* -- a caller that differences these solves passes
its own tolerance, a caller that differentiates the objective passes `NA` -- and the
stale "shared with focei's own fast gradient" line above it was corrected at the same
time.

### The new test did not actually fail without the fix

Its `1e-4` tolerance was taken from the 1.7e-2 gap measured on a 5-ETA 2-compartment
model, but that gap **tracks the fit tolerance**, and on the 3-ETA model the test
actually runs, at `sigdig = 4`, the defect is only **1.16e-5** -- under the threshold.
Measured both sides on the model under test (1.16e-5 with the bug, 5.6e-8 without)
and set `1e-6`, with the numbers recorded in the test so it is not loosened back.

It also compared `fast=TRUE` against `fast=FALSE` without checking they took
different routes, so a `fast=TRUE` run that stopped reaching the pool would have
compared the unpooled route against itself and agreed at any tolerance.  It now
asserts `pooledSolveN` advances under `fast=TRUE` and not under `fast=FALSE` -- and,
because `pooledSolveN` counts the *attempt*, that no subject was flagged.

### Two further defects found on the same path

**An unsolvable subject was scored into the covariance as zeros.**  A subject whose
pooled augmented solve failed was given a zero-filled `E` and returned with the rest.
That was written for the R outer gradient, which replaced such a subject's column
wholesale with a finite difference; the R gradient is gone and the covariance is what
reads these structures now.  Nothing downstream caught it -- a zero `E` is *finite*,
so the callers' `is.finite` guards pass.  It now falls through to the unpooled route,
which is what the comment on that branch already claimed happened.

**The fit's tolerance could be captured from an already-tightened state.**
`op_focei.fitAtol/fitRtol` are reset per fit and captured lazily by whichever
`OdeFitTolGuard` runs first.  Selecting the covariance guard *instead of* the fit
guard meant a fit whose first guard was this one never recorded its own tolerance, so
the next guard captured the tight covariance value and pinned it as the fit's --
reachable through the per-individual finite difference at the end of that same
function, which takes an `OdeFitTolGuard` while the tightened guard is still live.
The fit guard is now taken unconditionally and first, with the tighter one layered on
top; they nest and unwind in the right order.

### From the second-opinion passes

- `cores = 0` had no test and could not get a useful one through a fit: on a host
  where `getRxThreads()` is 1, the old literal `1` and the new value are the same
  number.  Resolution moved to `.foceiPoolCores()` and is pinned in a new essential
  (non-batched) test.  Writing it turned up that a non-numeric `cores` returned
  `NA_integer_` with a coercion warning rather than falling back -- and a warning
  raised inside a fit lands in `$runInfo`.
- `.foceiAnalyticSolveAll`'s `tol` defaulted to `1e-10`, which contradicts what the
  argument now means: a caller omitting it *because* it wants the fit's tolerance
  would silently get `1e-10`.  There is no value right for both, so the default is
  gone and callers state it.

Rejected: that an explicit `covSolveTol` needs its own case separate from the
sigdig-derived default -- both reach the pooled solve through the same `solveTol`
variable, with no branch between them that could regress one and not the other.

Still not covered, and worth saying plainly: there is no test that a *failed* subject
falls back rather than zero-filling.  Provoking one deterministically needs a model
that fails for exactly one subject, which I did not build.

mattfidler:
Second review round (Antigravity/Gemini and Copilot CLI/GPT, independently).  One of
them found a real hole in my own earlier fix; the other's headline finding turned out
to be wrong, but only measurement could show that.

### The fit tolerance could still be captured from a tightened state

My earlier commit took the fit-tolerance guard first *inside* `vaeOuterSolve_`, on the
grounds that `op_focei.fitAtol/fitRtol` are captured lazily by whichever guard runs
first in a fit.  That was not enough.  `CovSolveTolGuard` wraps the **whole** covariance
step, so on a fit that never ran an analytic gradient -- the only other thing that takes
a tolerance guard -- it is the first guard of the fit, and `vaeOuterSolve_` is entered
with the tolerances already tightened to `covSolveTol`.  The capture inside then recorded
`covSolveTol` as the fit's and pinned it for everything afterwards, including the
per-individual finite difference.

The invariant is now stated once, in `foceiNoteFitTol()`: a guard about to overwrite the
live tolerances records what it read first.  All three guards call it, and they are the
only three places in the package that touch `rxSetSolveAtolRtol`.

### The threading claim, settled by measurement

The other reviewer read `outerSolveFill`'s `min2(cores, getOpCores(op))` as clamping the
R-side fix straight back to a serial `1` under the default `rxControl(cores = 0)` --
which would have made this half of the PR a no-op.  It does not: rxode2 resolves
`op->cores` during solve setup, so `getOpCores(op)` is the thread count rather than the
raw control value.

Neither number was observable from R, which is why this could not be settled by reading.
`odeSwapInfo_()` now reports `pooledSolveCores` -- what the subject loop ran with after
every clamp.  On an 11-thread host: **11 with the fix, 1 with `.foceiPoolCores()` forced
back to its old literal `1`**.  The covariance test asserts it goes above 1 wherever the
host has threads to give, which covers the whole chain the unit test cannot reach.

### Also

`covSolveTol` vs the sigdig-derived default now has a no-fit test.  I had rejected this
twice on the grounds that both reach the pooled solve through the same `solveTol`
variable with no branch between them -- that still holds, so what the test pins is which
of the two wins, and that the default tracks `sigdig` rather than a frozen literal.

### Knowingly left undone

There is still no test that provokes a genuine per-subject solve failure and checks the
fallback.  Making one subject fail deterministically and portably means relying on
solver behaviour that is platform-sensitive, and a flaky test here would be worse than
none.  Both reviewers raised it; it is a real gap, not an oversight.

`foceiOuterFdInd_`'s work is discarded when a flagged subject sends the population to
the rxSolve fallback.  That is wasted effort on a rare path, not a wrong answer, and the
machinery is scaffolding for planned work -- left alone deliberately.

mattfidler:
Third round.  Antigravity came back clean on everything it was pointed at (the
`foceiNoteFitTol` invariant, the `pooledSolveCores` atomic, and whether the new
assertions can actually fail).  Copilot found one more, and it is real.

### `covSolveTol` was dropped part-way through the covariance step

`restoreFitSolve_()` rebuilds the fit's global rxode2 solve from the fit's saved
`rxControl` -- and rebuilding the solve resets `atol`/`rtol` along with it.  Both of
its call sites sit inside `CovSolveTolGuard`, so from the restore onward the
finite-difference covariance work ran at the fit's tolerance again.  It is reached
whether the analytic route **succeeded** or **declined**, so this is not just a
fallback-path corner.

Instrumented with the guard active at `covSolveTol = 1e-12`:

```
[probe] before rxSolve_: atol=1e-12 rtol=1e-12
[probe] after  rxSolve_: atol=1e-07 rtol=0.0001
```

The live tolerance is now carried across the rebuild.  The rebuilt solve also reads
its tolerances straight off the fit's own `rxControl`, which makes it a better source
for the recorded fit tolerance than any guard's capture of whatever happened to be
live, so it records that too.

Worth noting from the probe: the fit's `atol` and `rtol` genuinely differ (`1e-7` vs
`1e-4`).  The guards set a single value for both, and `.foceiAnalyticSolveTol()`
returns one number that the unpooled route also applies to both, so the two routes
still agree -- but it is why "the fit's tolerance" is a pair and not a number.

This one is a pre-existing defect rather than something the PR introduced.  It is in
scope because it is the same failure as the PR's title -- `covSolveTol` asked for and
not used -- reached a different way, and a reader who saw the first fix land would
reasonably assume the second was covered.

mattfidler:
This looks good to me @hiddevandebeek (I did look through the code).  There still is a bunch of AI commentary, but the code looks ok.

mattfidler:
You should make sure you agree and then once the tests pass we can merge this.

mattfidler:
Round 4 came back clean from both reviewers, so the review loop has converged.

### Test results

Run against the final build, on this branch:

| file | tests | passed | failed | errors | warnings |
|---|---|---|---|---|---|
| `test-focei-pooled-solve-args.R` (new) | 2 | 14 | 0 | 0 | 0 |
| `test-cov-analytic.R` | 39 | 195 | 0 | 0 | 0 |
| `test-agq-cov.R` | 3 | 12 | 0 | 0 | 0 |
| `test-vae-nonmutheta-grad.R` | 15 | 51 | 0 | 0 | 0 |
| `test-focei-fast-grad.R` | 11 | 46 | 0 | 0 | 0 |

**70 tests, 318 assertions, no failures and no warnings.**  One skip, in
`test-cov-analytic.R`: the pre-existing deliberate `skip()` at line 971 ("FOCEI
analytic R is wrong ... gold harness retained as the acceptance test").  Unrelated to
this branch.

`test-cov-analytic.R` was the file the PR description flagged as not having been run
-- its header states the analytic augmented solve runs at `10^-(sigdig+6)`, which is
exactly the premise the pooled path was violating.  It passes.  `test-agq-cov.R` was
worth running for the same reason (it reads `.foceiAnalyticSolveTol` directly), and
the two gradient files because the later commits touch shared C++: `restoreFitSolve_`,
`foceiNoteFitTol`, and the `odeSwapNotePooledSolve` signature.

`test-focei-pooled-solve-args.R` is deliberately NOT in `.slowBatches` -- it is
fit-free (0.12s) and runs on every push, unlike everything else here, which is weekly.

### Summary of the branch

Beyond the two original fixes, this now also fixes: `covSolveTol` being dropped after
the solve restore; a failed subject scored into the covariance as zeros; and the fit
tolerance being capturable from an already-tightened state (twice -- the first fix for
it was incomplete).

Two things are knowingly left undone and are described in the comments above: no test
provokes a genuine per-subject solve failure, and `foceiOuterFdInd_`'s work is
discarded on that fallback path.

PR #857, opened 2026-08-05
feat(focei): analytic gradient and covariance for a theta shared by two random effects

hiddevandebeek:
## Summary

The analytic FOCEI outer gradient (`foceiControl(fast=TRUE)`) and the analytic covariance (`covMethod="analytic"`) declined to finite differences whenever a structural theta was the mu-reference of **more than one** random effect -- one typical value entering two parameters, each with its own eta:

```r
cl <- exp(tx + eta.cl)
v  <- exp(tx + eta.v)   # tx is the mu of both eta.cl and eta.v
```

`.foceiAnalyticDirections()` bailed with *"a theta shared by two random effects"*, because the direction map reuses a **single** eta's sensitivity column per theta and could not express `df/dtx = df/d(eta.cl) + df/d(eta.v)`.

## Change

Such a theta now takes its **own true-sensitivity direction** -- the same path any non-mu structural theta already uses -- where the augmented model differentiates `df/dtx` symbolically. That derivative *is* the required sum over the etas the theta mu-references, so no special "summing" machinery is needed. The decline is removed; nothing else in the assembly changes.

Because `.foceiAnalyticDirections()` is shared, this enables **both** the analytic gradient and the analytic covariance for these models in one change; they previously both fell back to FD.

## Validation

- **Gradient** matches central-difference FD exactly (the shared theta to 4 decimals).
- **Covariance** matches the observed-information oracle -- the FD Jacobian of the (independently FD-validated) analytic gradient -- over the theta block to within FD noise (~2% SE, dominated by the oracle's own noise), cross-terms included.
- The analytic-vs-`r,s` SE difference seen in passing is the analytic-R vs FD-**sandwich** estimator difference: a matched *normal* model shows the same pattern, so it is not a shared-theta effect.

The existing *"falls back to FD when a theta is shared by two etas"* test is replaced by one that fits a well-specified (simulated) shared-theta model and asserts the analytic covariance engages (not the FD fallback) and its SEs match the finite-difference R-matrix.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,14 @@
   did not verify was the objective below rather than the gradient; against
   central differences of the corrected objective it agrees to 8e-3 relative.
 
+- The analytic outer gradient and `covMethod = "analytic"` covariance now handle
+  a structural parameter that is the mu-reference of **more than one random
+  effect** (one typical value entering two parameters, each with its own eta),
+  which previously fell back to finite differences.  The parameter takes its own
+  sensitivity direction, where the augmented model differentiates it symbolically
+  -- the correct sum over the etas it references.  The gradient matches central
+  differences exactly and the covariance matches the finite-difference R-matrix.
+
 ## Bug fixes
 
 ### Estimation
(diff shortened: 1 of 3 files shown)


mattfidler:
This may take some refactoring for saem and other places.  It may also take some refactoring in rxode2 which identifies the mu-referenced thetas.

hiddevandebeek:
ill leave it open for now.

mattfidler:
Maybe we can discuss if/how it can be implemented elsewhere.  It is a good idea.  Flagging it for analytical gradients for instance may be helpful

PR #865, opened 2026-08-06
fix(saem): do not freeze an eta when the informativeness test underflows

hiddevandebeek:
`saemControl(handleUninformativeEtas=TRUE)` is the default.  It decides, per subject and per eta,
whether that subject's data informs that eta, and the ones that fail are frozen: the mask is
multiplied into the MCMC proposal noise (`saem.cpp:3991`, `:4384`), so a flagged eta stays at its
mu for the whole fit.

The test perturbs each eta by `+/- qnorm(1 - alpha/2) * sd * sqrt(3/5)` and sums
`pred(-) + pred(+) - 2 * pred(0)` over the subject's observations, calling the eta informative
when that exceeds `tol` (`1e-7`, absolute).

It runs **once, at the initial estimates**.  If those are poor enough that the prediction
underflows at the observed times, the sum is tiny for a reason that has nothing to do with the
eta, and the eta is frozen for the rest of the fit -- including long after `theta` has moved
somewhere sensible.

## The change

Accumulate the size of the predictions the statistic is built from, and only take an
"uninformative" verdict when they are above the tolerance:

```c
retL(i, j) = (std::abs(ret(i, j)) > tol) || !(scale(i, j) > tol);
```

This is a monotone relaxation -- `retL_new >= retL_old` everywhere -- so it can only ever
*unfreeze* an eta, never freeze one that is currently free.  No existing fit gains a frozen eta.

I deliberately did not switch to a relative tolerance.  Predictions in the failing regime sit
around `4e-11`, under rxode2's default `atol=1e-8`, so they are solver noise and a relative test
would be comparing noise with noise; it would also change the threshold for every user and could
newly freeze etas with small but real curvature.

## Effect

Warfarin (32 subjects), started from `k = exp(0) = 1/h` where the true value is near `0.02/h`, so
the predicted concentration at `t >= 24 h` underflows.  Every subject whose first sample is at
`t = 24` -- 19 of 32 -- had its volume eta frozen.  Distance is the Euclidean distance of
`(tka, tv, tk, add.sd)` to the FOCEi estimate, mean of 4 seeds, `nBurn=60`/`nEm=40`:

| | frozen etas | omega.v | distance to FOCEi |
|---|---|---|---|
| FOCEi reference | -- | 0.034 | -- |
| before | 19 / 32 | 0.009 | 0.320 |
| after | 1 / 32 | 0.034 | 0.067 |
| before, `handleUninformativeEtas=FALSE` | 0 / 32 | 0.041 | 0.099 |

The last row is the control: the fix gets there without the user having to know the flag exists.

That the initial estimates are the trigger, rather than the data, is settled by running the same
model and data from estimates near the solution -- 0 of 32 frozen, before and after.

## Not fixed here

One subject (the largest dose in the dataset) is still frozen: its predictions sit just above
`tol` while its curvature sits just below, so the original test applies and still gets it wrong.
The underlying design issue is that the mask is computed once and never revisited, and the
targeted follow-up is to re-evaluate it at the end of burn-in, once `theta` has moved.  That is a
larger change -- it touches the SAEM loop rather than one leaf function -- so it is deliberately
out of scope here.

## Tests

`tests/testthat/test-uninformative-etas.R` gains three assertions that call the test directly with
a synthetic environment, so they need no fit:

- real predictions, no response to the eta -- still uninformative (behaviour preserved)
- real predictions that respond to the eta -- informative
- predictions that underflow -- must not be frozen (fails before this change)

The existing assertions in that file are the real regression guard: they check that IV-only
subjects still come back with `eta.ka == 0`, i.e. that the feature still does its job where the
predictions are meaningful.  They pass unchanged.


mattfidler:
## Review

Reviewed this three times over, each round cross-checked by two independent reviewers -- GitHub Copilot (`gpt-5.4`) and Antigravity (`gemini-3.1-pro-high`) -- and iterated until both came back clean. Every claim below was reproduced by calling `_nlmixr2est_uninformativeEta` directly rather than reasoning about the source.

The diagnosis is right and the direction is right. Two defects in the implementation, both of which let the original bug back in.

### 1. `scale` is a sum, so the verdict depends on the sample count, not the predictions

`scale(curid, cureta) += std::abs(val[i])` accumulates over all `3 * nobs` rows for the subject/eta and is then compared against the same absolute `tol`. So the guard crosses `tol` on observation count alone. With every prediction at `4e-11` -- squarely in the "too small to judge" regime the guard exists for -- and `ret` exactly `0`:

| observations | `sum abs(pred)` | mask |
|---|---|---|
| 1 | 1.2e-10 | 1 (free) |
| 833 | 1.0e-7 | 1 (free) |
| 1000 | 1.2e-7 | **0 (frozen)** |
| 5000 | 6.0e-7 | **0 (frozen)** |

The guard silently stops working from about `n = 834` observations, so it helps sparsely-sampled subjects and abandons richly-sampled ones. I suspect this is also the "one subject still frozen" in the PR description -- it is the largest dose in the dataset, i.e. the one whose predictions clear `tol` soonest in aggregate.

Taking the largest prediction instead of the accumulated one makes the verdict depend on the predictions, which is what the comment says it does:

```c
double curAbs = std::abs(val[i]);
if (curAbs > scale(curid, cureta)) scale(curid, cureta) = curAbs;
```

### 2. A failed solve is read as "no curvature" and freezes the eta

`NaN > tol` is `false`, so if the ODE solve fails on **one** perturbation arm, `ret` is `NaN` while the other arms keep `scale` finite and over `tol`. `havePred` is then true, `abs(NaN) > tol` is false, and the eta is frozen for the whole fit:

```r
.ue(c(10, 10, NaN))   # 0 -- frozen
.ue(c(NaN, 10, 10))   # 0 -- frozen
.ue(c(10, NA, 10))    # 0 -- frozen
```

This is pre-existing rather than introduced, but it is the same class of bug the PR is fixing -- an eta frozen for a reason that has nothing to do with the eta -- and it is worth closing in the same change. Checking `ret` for finiteness alongside `scale` covers both the partial and the total failure:

```c
bool havePred = R_finite(ret(i, j)) && R_finite(scale(i, j)) && scale(i, j) > tol;
```

Both changes preserve the PR's stated invariant. `retL = (abs(ret) > tol) || !havePred` can only ever return `1` where the original test returned `1`, and each change only makes `havePred` false more often, so they can only unfreeze further. No fit gains a frozen eta.

### Checked and rejected

Reported by one reviewer or the other, verified as not real:

- **The absolute `tol` disables detection for small-valued models.** True, but the failure direction is safe (the feature turns itself off) where the old behaviour was to freeze everything, which is strictly worse. `tol` is fixed at `1e-7` inside `.uninformativeEtas()` and is not reachable from `saemControl()`, so it cannot be driven to a pathological value. Agreed with the PR's reasoning for not going relative.
- **`.Call()` with a string routine name fails under `load_all` / `R CMD check`.** It does not. `src/init.c` calls `R_useDynamicSymbols(dll, TRUE)` and `R_forceSymbols(dll, FALSE)`, and the file runs clean both ways.
- **`id[i] - 1` / `w[i] - 1` index out of bounds for non-sequential subject IDs.** Unreachable -- nlmixr2 renumbers IDs to `1..n` before `.uninformativeEtas()` is called. Traced a real fit with `ID = 7, 42, 100`; the function sees `1, 2, 3`. (Worth knowing that the merge in `R/uninformativeEtas.R:230` would return zero rows rather than index out of bounds if that ever stopped holding.)
- **The test's `pm` layout `c(-1, 1, 0)` computes a different statistic than production's `c(-1, 0, 1)`.** The kernel accumulates by `pm` value, not row position, so it is order-agnostic. The pairing in the test is valid.
- **Non-positive `tol` produces a nonsense mask.** `tol` is a default argument of `.uninformativeEtas()` fixed at `1e-7`; `R/saem.R:202` does not pass it and `saemControl()` does not expose it. Not reachable.
- **Naive accumulation of `ret` can cancel a real signal below `tol`.** Reproduced -- with predictions at `1e6` and a per-observation second difference of `1.86e-9`, the C++ running sum lands at `2.3e-8` against an exact `1.86e-7`, so the eta freezes. But `ret`'s accumulation is untouched by this PR, and the signal being lost is `1.9e-15` *relative* to the prediction, i.e. within an order of magnitude of double eps. An eta with that little effect is uninformative on any reading; the verdict is defensible and the real culprit is again the absolute `tol` at a prediction magnitude of `1e6`. Left alone.

### Test coverage

The three new assertions all exercise the 1-observation shape, so none of them would have caught issue 1. Added:

- a multi-observation helper covering `n = 1, 10, 833, 1000, 5000` -- verified to **fail at `n = 1000` and `n = 5000`** against the summed guard, and to leave the "eta does nothing" / "eta moves the predictions" cases at every `n` untouched;
- non-finite predictions on every arm and on a single arm.

Full file passes: 25 assertions, 0 failures. The pre-existing IV-only regression test (`eta.ka == 0` for subjects 1-3) still passes, so the feature still freezes where the predictions are real.

### Not addressed

The mask is still computed once and never revisited, so a subject whose predictions are just above `tol` while its curvature is just below is still frozen on initial estimates alone. Re-evaluating at the end of burn-in remains the right follow-up, and I agree it is out of scope here.

Optionally: when the guard fires for a subject, that is a usable signal that the initial estimates are bad. Surfacing it in `$runInfo` would tell the user something the fit currently swallows.

### Patch

Committed on top of 652972e as `bf33f73`. Happy to push to this branch if you would like.

<details>
<summary><code>src/uninformativeEtas.cpp</code></summary>

```diff
-  // size of the predictions `ret` is built from, to tell "the eta does not move the
-  // prediction" apart from "there is no prediction to move"
+  // size of the predictions `ret` is built from, to tell "the eta does not move the
+  // prediction" apart from "there is no prediction to move".  Largest, not a sum: a sum
+  // grows with the observation count, so the same underflowing subject would change
+  // verdict on nothing but how often it happened to be sampled.  NaN never wins the
+  // comparison, so a failed solve leaves this at zero.
   NumericMatrix scale=NumericMatrix(nid, neta);
   std::fill(scale.begin(), scale.end(), 0.0);
   IntegerMatrix retL=IntegerMatrix(nid, neta);
   for (int i = 0; i < simId.size(); i++) {
     int curid = id[i] - 1;
     int curpm = pm[i];
     int cureta = w[i] - 1;
-    scale(curid, cureta) += std::abs(val[i]);
+    double curAbs = std::abs(val[i]);
+    if (curAbs > scale(curid, cureta)) scale(curid, cureta) = curAbs;
     switch (curpm) {
@@
       // This runs once, at the initial estimates.  If those are poor enough that the
-      // prediction underflows at every observed time, perturbing the eta moves nothing
-      // for a reason unrelated to the eta -- do not freeze it for the whole fit on that.
-      retL(i, j) = (std::abs(ret(i, j)) > tol) || !(scale(i, j) > tol);
+      // prediction underflows (or fails outright) at every observed time, perturbing the
+      // eta moves nothing for a reason unrelated to the eta -- do not freeze it for the
+      // whole fit on that.  `ret` is checked too: a solve that failed on ONE arm leaves it
+      // NaN while the other arms keep `scale` finite and over tol, and NaN > tol is false.
+      bool havePred = R_finite(ret(i, j)) && R_finite(scale(i, j)) && scale(i, j) > tol;
+      retL(i, j) = (std::abs(ret(i, j)) > tol) || !havePred;
```

</details>

mattfidler:
Pushed the two fixes to this branch as `bf33f73` (fast-forward from `652972e`, nothing upstream rebased away).

Recap of what changed on top of your commit:

- `scale` now tracks the **largest** prediction rather than the accumulated one, so the verdict no longer crosses `tol` on observation count alone.
- `havePred` also requires `ret` to be finite, so a solve that fails on a single perturbation arm is no longer read as "no curvature".
- Tests: a multi-observation case (`n = 1, 10, 833, 1000, 5000`) that fails against the summed guard at `n = 1000` and `n = 5000`, plus non-finite predictions on every arm and on one arm. 25 assertions in the file, all passing, including the pre-existing IV-only regression test.
- `NEWS.md`: dropped a trailing-whitespace-only line, restored the blank line after `### Estimation`, and reworded to match the new guard.

Since the C++ signature is unchanged, `src/init.c` needs no update.

mattfidler:
## Follow-up: the end-of-burn-in revisit

Added `saemControl(revisitUninformativeEtas=)` as the targeted follow-up named in "Not fixed here" -- re-deciding the mask once `theta` has moved, rather than living with the verdict the initial estimates produced.

**It ships off by default.** I measured cost and benefit before choosing that, and the benefit did not survive measurement.

### How it works

`SAEM::revisitUninformativeEtas()` (`src/saem.cpp`) re-runs the informativeness test at the end of burn-in. Predictions come from `user_fn` -- the evaluation the SAEM loop already performs every iteration -- so it adds no `rxSolve` of its own and no R callback: the cached `_rx` and rxode2's solve state are untouched, and since `user_fn` draws no random numbers the RNG stream is unchanged. It builds a phi matrix at `eta = 0` from `mprior_phi1`/`mprior_phi0`, perturbs one column by `+/- delta`, and applies the same verdict as `_nlmixr2est_uninformativeEta`.

The probe half-width comes from the **initial** Omega, exactly as the first test's does, so the two evaluations differ only in `theta`. Reading the current Omega instead makes the revisit fight itself: a frozen eta contributes nothing to its own variance, so Omega shrinks, the probe narrows, and the eta looks *less* informative the longer it has been frozen.

Skipped for mixture models (their first test runs through a separate pruned model) and when `handleUninformativeEtas=FALSE`. `saem` now returns `ueRevisitInfo` (`ran`/`unfroze`/`froze`) so tests can assert the mechanism ran rather than infer it from an unchanged mask.

### Cost: nil

theo_sd, `nBurn=200`/`nEm=300`, 6 reps each:

| | wall time | objf |
|---|---|---|
| off | 4.42s (sd 0.53) | -- |
| on | 4.22s (sd 0.10) | bit-identical |

Inside the noise. The probe is `neta + 1` extra evaluations at **one** iteration against ~500 iterations x ~6 evaluations each.

### Benefit: not demonstrated

1. **The scenario that motivated this PR** (warfarin from `k = 1/h`): 0 of 32 etas frozen either way; the mask never changes. The underflow guard already fixes it.
2. **Well-behaved fits**: the revisit runs, changes no verdict, identical results.
3. **The only case where the mask changed** -- `tv=20`, so every prediction underflows at the initial estimates:

| | objf | ran/unfroze/froze | `eta.ka==0` IV \| PO |
|---|---|---|---|
| off | 540.04 | 0/0/0 | 0 of 3 \| 0 of 3 |
| on | **750.37** | 1/0/**12** | 3 of 3 \| **3 of 3** |

It fired, froze 12 verdicts, and made the objective **worse** -- and froze `eta.ka` for the PO subjects, where `ka` *is* identifiable. Both fits are unrecoverable from that start, but the revisit did not help.

The structural reason: the two verdicts only disagree when `theta` moved a long way during burn-in, which usually means it has not settled -- so the second verdict is no more trustworthy than the first, and unlike the first it can freeze an eta permanently. Meanwhile `tol` is absolute at `1e-7` while predictions are O(1)-O(1000), where solver noise at `rtol=1e-4` already swamps it, so little is frozen to begin with.

Hence: available, documented, off. Anyone who hits the narrow residual case can turn it on; nobody pays for it otherwise. If you would rather it were unfreeze-only (`ue = max(ue_initial, ue_revisit)`), that is a one-line change -- it makes the `tv=20` regression impossible, at the cost of it doing nothing in every case I measured.

### Review and tests

Copilot (`gpt-5.4`) and Antigravity (`gemini-3.1-pro-high`) both returned **no issues** on the probe indexing, memory safety, solve-state save/restore, and verdict parity. I separately hand-checked the chain-to-level mapping for `nmc = 1, 2, 3, >3`: each of the three levels is read exactly once, and padded duplicate chains are dropped by the `l0 + c < 3` guard.

New file `tests/testthat/test-uninformative-etas-revisit.R` (19 assertions, weekly batch 6): control default and validation; the revisit runs and changes a verdict in the `tv=20` case; it agrees with the first test from good initial estimates; and it does not run when disabled or when `handleUninformativeEtas=FALSE`. Both test files pass -- 44 assertions, 0 failures.

`NAMESPACE`/`DESCRIPTION` are deliberately untouched: my roxygen is 8.1.0 against the repo's 8.0.0 and reflows every `importFrom` block, so I reverted that churn and kept only `man/saemControl.Rd`.

mattfidler:
Looks good to me.

PR #866, opened 2026-08-06
fix(focei): clamp the lower gradTrim bound at -gradTrim, not +gradTrim

hiddevandebeek:
## The bug

`numericGrad()`'s lower gradient clamp tests `g[cpar] < op_focei.gradTrim` where it means `< -op_focei.gradTrim`:

```c
if (g[cpar] > op_focei.gradTrim){
  g[cpar]=op_focei.gradTrim;
} else if (g[cpar] < op_focei.gradTrim){   // should be < -op_focei.gradTrim
  g[cpar]=-op_focei.gradTrim;
}
```

The branch above has already caught everything over `+gradTrim`, so by the time the `else if` runs we know `g <= gradTrim` -- and `g < gradTrim` is then true for every value except exactly `gradTrim`. So **every remaining component is replaced by `-gradTrim`**, rather than only those below `-gradTrim`.

Concretely, on the `|g| > gradTrim` path: the forward difference comes back at `2e4` with `gradTrim = 1e4`, the component is re-differenced centrally and returns a legitimate `+50`, and that `+50` is then written as `-1e4`. Wrong sign, magnitude out by 200x, handed straight to the optimizer.

## Scope

Four sites, all on paths that re-difference a component before clamping it: the backward and central rescues after a non-finite forward difference, and the two `|g| > gradTrim` branches that recompute centrally first. The *outer* tests in the same function already use the correct `-gradTrim` form (as do the non-recomputed arms), which is what makes this look like a copy-paste slip rather than intent.

**Latent at the defaults.** `gradTrim` defaults to `Inf`, so `R_FINITE(op_focei.gradTrim)` is false and none of these branches execute. It becomes reachable when `foceiControl(gradTrim=)` is set to a finite value.

## Testing

None added. Triggering it needs a finite `gradTrim` *and* a forward difference that goes non-finite or exceeds the trim at a specific parameter; a test arranging that would be fragile enough to be worse than no test. The change is a sign correction on four otherwise identical clamps, and the existing suite covers the default `gradTrim = Inf` path where these branches are skipped.


mattfidler:
## Review: Claude + Copilot (gpt-5.4) + Antigravity (gemini-3.1-pro-high)

Five rounds, iterating until both external reviewers came back clean. Round 5: Copilot "No remaining actionable defects found", Antigravity "There are no defects remaining."

### The original fix is correct and complete

All three reviewers independently confirmed it. All four inner clamp sites are patched; the two without their own `R_FINITE(gradTrim)` guard are covered by their outer guards; a repo-wide scan finds no other site with the same slip.

One correction to the PR description: these branches are **not** reachable at the defaults, and `gradTrim = Inf` is not the only reason. `outerOpt` defaults to `bobyqa`, which is derivative-free -- `numericGrad()` is never called. Reaching any of this needs a gradient-based `outerOpt` (`nlminb`, `n1qn1`, `lbfgsb3c`, ...).

### Four more defects in the same function

Review turned these up in `numericGrad()`. All pre-existing, all fixed here, since leaving one arm of a set of identical clamps unfixed is worse than fixing none.

**1. The `gradCalcCentralSmall` confirmation returned the objective, not a derivative.** `tmp = g[cpar]` overwrote the objective at `theta+delta` *before* the next line differenced against it, so the branch computed `(gradient - objective)/(2h)`. The magnitude guard below it could never fire, because it compared against the same clobbered value.

Measured on `theo_sd` with a gradient-based `outerOpt`, forcing every component through the branch:

| | gradient components |
|---|---|
| before | -2.6e5 ... -1.1e6, every one negative, fit dead after one iteration |
| after | -20 ... +22, same scale as the plain forward difference, 8 iterations |

The `-objf/(2h)` signature is unmistakable. Unlike the sign slip this needs no unusual control setting -- just a component whose forward difference falls under `gradCalcCentralSmall` (default `1e-4`), which happens as the outer problem converges.

**2. That result was the one arm of the chain with no `gradTrim` clamp** -- an `else if` sitting after both clamp branches.

**3. The non-finite fallback assumed the pure-central path.** Reached through the `gradCalcCentralLarge` switch it read a stale `tmp0` -- `NA_REAL` on the first parameter, the *previous parameter's* perturbed objective after that -- yielding minus the forward gradient in the first case and a cross-parameter difference in the second. On the central path it also differenced against an `f` that is only ever filled in on the forward path, so `f == 0`. Both halves are broken, in complementary ways. Fixed by putting `tmp0`/`tmp` into the central convention at the switch, plus a `haveF` flag so the fallback computes the unperturbed objective when it needs it.

**4. A failed confirmation solve discarded the whole gradient.** All three confirmation branches (`gradCalcCentralSmall` and both `gradTrim` recomputations) replaced `g[cpar]` with their central difference unconditionally. When the `cur-delta` solve fails that difference is `NaN`; every comparison below it is then false, so neither the magnitude selection nor the clamps recover, and the gradient is thrown away via `zeroGrad`. The branch conditions guarantee the pre-recompute value is finite, so all three now fall back to it.

Both external reviewers found this one independently, in rounds 3 and 4. Worth flagging that fix 1 made one route to it *newly* reachable: with `tmp` no longer clobbered by a finite value, a `tmp` that the NA rescue above had already left non-finite could now propagate into the confirmation.

### Tests

New `tests/testthat/test-focei-fd-fallbacks.R` (18 assertions, ~14s, kept in the essential push/PR subset).

The forced-confirmation test reproduces defects 1 and 2 directly: it drives every component through the branch and verifies it ran, via the `Mixed Gradient` rows in `parHistData`, with `gradCalcCentralLarge` pushed out of reach so that label is unambiguous. It then pins the gradient scale. Confirmed it fails against the unfixed code.

**The `gradTrim` sign slip itself still has no test, and I agree with the reasoning in the PR description.** I tried: rebuilding with the unfixed clamps leaves the fit bit-identical. Triggering it needs a component whose forward difference clears `gradTrim` while its recomputed central difference lands back inside the band, and that is not arrangeable from a control setting. The third test is therefore a bounds guard only -- it catches a clamp that goes missing or stops being symmetric, not a re-broken sign -- and its aggregate assertions are documented in the file as unable to isolate a single mis-clamped component.

### Verification

500 assertions across 13 FOCEi/AGQ test files, 0 failures, on a clean rebuild after merging `main`:

`focei-1`, `focei-fd-fallbacks`, `focei-fast-grad`, `focei-shi21-bounds`, `focei-zero-theta`, `focei-theta-reset-bounds`, `focei-warm`, `focei-wang2007-basic`, `focei-eta-only`, `focei-foce-plus`, `cov-focei`, `mfocei`, `agq-fast-grad`.

Defects 1, 3 and 4 change the finite-difference outer gradient for anyone using a gradient-based `outerOpt`, so that is the blast radius to weigh -- results there will move, in the direction of being right.

mattfidler:
Looks good to me.

Issue #867, opened 2026-08-06
foceiControl(fast=) changes the converged fit when a log-transform endpoint has a non-positive prediction


hiddevandebeek:
## Summary

On stock `nlmixr2data::theo_sd` with a `lnorm` endpoint, `fast=TRUE` and `fast=FALSE` converge to different parameter estimates and different objective values (686.19 vs 1240.93). Dropping the `TIME==0` observations makes them agree to 0.2%.

Cause: the objective **clamps** the transformed prediction at rows where the untransformed prediction is <= 0, so it is locally constant in theta there, while the augmented sensitivity model differentiates the **unclamped** expression. The analytic outer gradient is therefore not the gradient of the objective it is minimising.

## Reprex

```r
library(nlmixr2)

mod <- function() {
  ini({
    tka <- 0.45; tcl <- 1; tv <- 3.45; prop.sd <- 0.7
    eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1
  })
  model({
    ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
    d/dt(depot)  <- -ka * depot
    d/dt(center) <-  ka * depot - cl / v * center
    cp <- center / v
    cp ~ lnorm(prop.sd)                       # log transform-both-sides
  })
}

ctl <- function(fast) foceiControl(print = 0L, covMethod = "", calcTables = FALSE,
                                   sigdig = 3, fast = fast, maxOuterIterations = 100L)
d <- nlmixr2data::theo_sd

fT <- nlmixr2(mod, d, "focei", ctl(TRUE))
fF <- nlmixr2(mod, d, "focei", ctl(FALSE))
rbind(fast_TRUE  = c(objf = fT$objf, unlist(fixef(fT))),
      fast_FALSE = c(objf = fF$objf, unlist(fixef(fF))))
#>                 objf     tka     tcl     tv  prop.sd
#> fast_TRUE   686.1876 -5.2524 -0.9481 -5.573   7.2201
#> fast_FALSE 1240.9290  2.6542 -0.7374  7.567   2.8442

d2 <- d[!(d$EVID == 0 & d$TIME == 0), ]      # drop the TIME==0 observations
gT <- nlmixr2(mod, d2, "focei", ctl(TRUE))
gF <- nlmixr2(mod, d2, "focei", ctl(FALSE))
[...]

PR #868, opened 2026-08-06
WIP: make the per-subject FD fallback actually apply, and cover omega (fast=TRUE)

hiddevandebeek:
**Draft -- opened for visibility, not review-ready.**

## The bug

`foceiControl(fast=TRUE)` sends the **whole** gradient evaluation to finite differences whenever any subject's augmented sensitivity solve fails -- even though the per-subject machinery to avoid that (`foceiOuterFdInd_`, the flagged-subject collection, the ES-batch swap, the eta pin, the outlier pass) is already there. It can never apply, for two reasons that compound:

```c
// outerSolveFill, on main: the bail comes BEFORE nobs is assigned
if (getOpNeq(op) > 0 && ISNA(solve0[0])) { E.ok = false; continue; }
int nobs = 0; ... E.nobs = nobs;                    // never reached for a failed subject
```

so a flagged subject arrives with `nobs == 0`, and then

```c
nobsAll[i] = Es[i].nobs;                            // 0
if (nobsAll[i] <= 0) return declineHere(13);        // declines the whole evaluation
```

which fires *before* the substitution at the end of `gradPooledCore`. The finite difference is computed and thrown away.

**Assigning `E.nobs` before the bail is the whole of what makes it reachable.**

## Then: omega

With it reachable, the existing substitution only ever covered theta -- `gradPooledCore` writes `gmat` rows by KERNEL slot from full-theta-indexed FD columns, and its own documentation records the gap:

```
//' Omega directions are NOT covered yet
```

The fold-in now goes through `fixedTrans` and covers the whole row, so a flagged subject contributes to omega as well as theta and sigma.

Omega reaches an individual likelihood only through `curOmegaInv()`/`curLogDetOmegaInv5()`, and building a perturbed `Omega^-1` needs `setOmegaTheta()`, which calls R and cannot run inside a parallel region. The perturbed pair is built once serially per evaluation and applied per subject through a `thread_local OmegaScope` while the subjects are differenced in parallel.

## Theta/sigma restructured to the non-fast path's shape

Replaces a per-subject step search with `numericGrad`'s arrangement: a sequential loop over the parameters the optimizer actually moves, **one** `shi21Central` step per parameter searched on the summed -2LL over the flagged subjects, then explicit +-h legs, with every likelihood evaluation parallel over subjects and a serial retry for any subject that failed (as `foceiS` does). Per-subject slopes survive the pooling, so the outlier test and its TV refinement are unaffected; `tvDerivFromSamples` is now shared by both passes, and the TV grid is walked once for the whole outlier set instead of once per outlier subject.

A forward-difference variant (one leg per parameter, escalating to central on a large slope -- exactly `numericGrad`'s arrangement) was implemented and **rejected on measurement**: it degraded 19 of 20 graded cases by 2x-81x. Recorded in a comment so it is not retried.

## Differences taken at the fit's ODE tolerance

A subject only reaches this path after rung 2 loosened its tolerance trying to rescue the failed solve, so it would otherwise be differenced on a *looser* function than the analytic terms it is summed with. Worse, a retry part-way through a phase could leave the two legs of one central difference at different tolerances. Both the global atol/rtol (reusing `OdeFitTolGuard`) and the per-individual factor are pinned before every evaluation.

## Validation

Graded against the **objective's own** gradient -- a `maxOuterIterations=0` refit differenced either side of theta, i.e. what the non-fast path would hand the optimizer -- not against the analytic gradient. Grading against the analytic is circular, and it concealed #867: on `lnorm` the analytic `tv` component is off by ~8.7x including a sign flip, and the FD substitution is the accurate one.

The substituted gradient tracks the analytic one to within the probe's resolution on 1-cmt add, add+prop, lnorm, single-eta, covariate, 2-cmt neta=5 and two-endpoint warfarin. Thread invariance holds (no OpenMP reduction anywhere; sums taken after the region in fixed subject order), and `covMethod="r,s"` is untouched.

## Also fixed

- The outer FD step store was allocated per optimizer parameter but indexed by full-theta position, so any `fix()`ed parameter wrote past its end.
- A step that terminated on its bound was cached *before* being rejected, so every later evaluation skipped the search, failed the same bound test and declined -- for the rest of the fit. Clamped steps are never cached now, and `nFdOutlier` gains a `stepClamped` counter.
- The loops walked every full-theta column rather than the free parameters, so a fixed parameter was differenced and discarded, and could trip the outlier pass into a 41-point TV grid.
- `mixDeriv` is set when a difference degrades to one-sided, matching `numericGrad`.
- `gradTrim`/`zeroGrad` now apply to the analytic route, which returned past both. (`gradTrim` defaults to `Inf`, so only the `zeroGrad` half changes default behaviour.)

## Test hooks

`NLMIXR2EST_OUTER_FD_H`, `_FD_SPAN`, `_FD_MAXINNER` added alongside the existing `NLMIXR2EST_OUTER_FAIL_ID`/`_FD_SKIP`; all tagged `TEST HOOK (removable)`. No model-level knob reaches the flagged state (the augmented and inner solves fail at the same `maxsteps` threshold), and without a forced step the accuracy of a difference is not observable from R -- which is how the step/span/cap hypotheses for #867 were eliminated. Inert unless set (verified). `tests/testthat/test-focei-outer-fd-fallback.R`, 26 assertions, essential push/PR subset.

## Known gaps

- **Not run through `devtools::test()`.** rxode2 5.1.7 (now required) will not install on my machine; under 5.1.4 the fixture helper cannot recompute fits and `test-focei-fast-grad.R` segfaults intermittently *on `main` too*. All validation above is direct numeric comparison. CI is the arbiter.
- The "bit-identical when nothing is flagged" no-regression baseline was captured against `76664cd90` and is stale after 25 commits of covariance/npag work; it needs regenerating from `b564ee833`.
- One run in four of the new test file saw `nAnalyticGradDirect` come back 0. Not reproducible (three clean re-runs, and `= 4` in every isolated scenario tried); consistent with this box's intermittent instability under rxode2 5.1.4, but unproven, and would need hardening rather than excusing if CI sees it.
- No progress reporting, unlike `numericGrad`/`foceiS`; cost is visible after the fact via `nOuterFdInd`, `stepClamped` and `nFdChartrand`.


hiddevandebeek:
Rebased onto `b564ee833` (current `main`); force-pushed. `src/inner.cpp` auto-merged, so the conflicts were only in `NEWS.md` and `R/foceiGradAnalytic.R`.

**One resolution changed behaviour and is worth calling out.** `main` had already fixed the flagged-subject zero-fill in `.foceiGradAnalytic`, and fixed it better than this branch did: it records the flagged ids, bumps the counter and falls through to the `rxSolve` route, so an analytic gradient is still produced by an alternative solve path. This branch returned `NULL` and declined to finite differences. I took `main`'s version wholesale and dropped the corresponding NEWS bullet -- claiming that fix here would have been wrong, and the branch no longer has the decline behaviour that bullet described. `NEWS.md` was purely additive otherwise.

**Post-rebase state on this machine** (rxode2 5.1.4, so read with the caveat below):

- builds clean against `main`'s covariance/npag changes
- thread invariance: every objf/parameter/gradient comparison identical at 1 vs 4 threads
- fallback still substitutes on all five probe models (`nOuterFdInd` 4/8/8/12/8, zero clamped steps, zero declines)
- `tests/testthat/test-focei-outer-fd-fallback.R`: 26/26 assertions, three consecutive clean runs

**Two things a reviewer should not take on trust:**

1. The "bit-identical to `main` when nothing is flagged" claim in the description was measured against `76664cd90`. `main` has since gained 25 commits across the covariance and npag paths, so that 96-line baseline is stale and the claim does not transfer until it is regenerated from `b564ee833`. I have not done that yet.
2. In one of four runs of the new test file, `expect_gt(nAnalyticGradDirect, 0)` failed (the analytic route declined for that fit). It did not reproduce -- three subsequent runs were clean, and the same model/flags/control gives `nAnalyticGradDirect = 4` in every isolated scenario I tried, including after the six `.foceiGradDirect` calls the preceding test makes. It is consistent with the intermittent instability this box already shows on the fast-gradient path under rxode2 5.1.4 (where `test-focei-fast-grad.R` segfaults intermittently on `main` as well), but I cannot prove that, and a flaky assertion would need hardening rather than excusing if CI shows it too.

hiddevandebeek:
Corrected the title, description and commit message. The earlier version framed this as adding the per-subject finite-difference fallback, saying the whole gradient evaluation previously fell back to finite differences. That was wrong: `foceiOuterFdInd_` and the fallback are already on `main` for the theta/sigma directions, and `main`'s own roxygen records the actual gap ("Omega directions are NOT covered yet"). What this branch does is cover omega, restructure the theta/sigma block to `numericGrad`'s shape, pin the differences to the fit's ODE tolerance, and fix the bugs listed. The NEWS entry had the same mis-framing and is fixed too.

hiddevandebeek:
Correcting my own correction, with the verification I should have done first.

I earlier said the per-subject fallback was already working on `main` and that this branch only extended it to omega. That was wrong. The machinery is on `main`, but it can never run: `outerSolveFill` bails on a failed solve *before* assigning `E.nobs`, so a flagged subject arrives with `nobs == 0`, and the stacking loop's `nobsAll[i] <= 0` guard declines the whole evaluation before the substitution is reached -- discarding the finite difference it just computed. Verified by diffing `76664cd90` and `b564ee833` directly rather than reading the substitution site in isolation.

So the original framing was right: on `main` a flagged subject sends the whole gradient to finite differences. Assigning `E.nobs` before the bail is what makes the fallback apply at all; covering omega is the second half. Title, description, NEWS and the commit message now say both.

hiddevandebeek:
@mattfidler I left in some hooks for testing, since normally the fallback isnt reachable (and thus cannot be tested). Perhaps you want to have a look at how FD fallback now handles compared to non-fast FD path?

mattfidler:
This is a large change, but it seem to read like I thought it should be applied.  While the commentary is sycophantic, I still believe that if there is an outlier slope the Chartrand TV is likely the best call.  

The commentary says it is the best call for just that failing outlier, but when I was testing it it required the whole group to be able to reproduce the right function value.  In that case it was in a poorly setup outer gradient pool so sometimes it worked and sometimes it didn't. 

I can imagine that a very pathological surface can imitate this phenomena, so I am still on the fence on if the Chartrand should be applied to all FD failing subjects or just one.

Of course, the analytic gradient subject (even relaxed) should be good enough

mattfidler:
After looking further:

- Solve that failed is `NA` is now pooled in with infinite solve.  I don't know this is the right behavior, what we measured was that if the solve should be relaxed instead.  Now it skips that and goes to FD -- this accidently pulls that step right out from under the user.

- The fintte difference is now moved out of `vaeOuterSolve_` which means both methods are no longer DRY and the implementation for the outer gradient which should have been applied for the vae auto-encoder is actually not as robust as it should have been.

- The theta/sigma **lost** the individual shi difference `h` values that I believe are so important in the solve.  These are things I struggled to get the AI to keep.


Also AI flagged the following gap

- the `analyticOuterGrad() == 0` is not an issue when the analytic gradient is actually zero, and is likely not to be zero when using FD of any sort.  So this simply skipping an actual possible gradient on an individual where it could be actually zero.  Then it is flagged when it shouldn't be.

mattfidler:
I am running an agent to address these findings and then maybe you can look at it after that.  I am also having it look at how the MAD outlier is handled to see if there is something that needs to change there.

mattfidler:
For the MAD outliers there are some things that need to be adjusted.  I will also add that at the end of the agent work (probably for some tunable options so we can run some numerical experiments in pathological likelihood surfaces)

Currently only FD outliers are tested but the statistic comes from the whole group.  Perhaps the right call, though MAD outliers inside the analytic gradient may call for some things.

mattfidler:
Merged `origin/main` (clean; main's `gradTrim` sign fix from #866 survives alongside the new clamp) and worked through the Copilot thread, my own review, an Antigravity review and a Copilot CLI review. Pushed as four commits.

## The tests now run

rxode2 5.1.7 installs on my box, so the description's largest "known gap" is closed:

- `test-focei-outer-fd-fallback.R` **39/39**
- `focei-fast-grad`, `focei-fd-fallbacks`, `focei-fast-methods`, `focei-ll-fast-grad`, `agq-fast-grad`, `cov-analytic`, `vae-nonmutheta-grad` -- **0 failures**
- **Thread invariance verified**: with two subjects flagged, the substituted gradient is *bit-identical* at 1 vs 4 threads, objf included. The claim holds.

I did not see the intermittent `nAnalyticGradDirect == 0` reported in the description across ~15 runs of the file.

## Real defects found and fixed

**1. Silent wrong gradient for `est="vae", nonMuTheta="grad"`** (found by the Copilot CLI, confirmed by reading). Since the substitution moved out of `gradPooledCore`, a flagged subject leaves a **zero column**, and `foceiGradPooledDirect_` returned `gv` straight to R. Zeros are finite, so the `is_finite()` check passed a gradient silently missing an entire subject. `main` declined here (its `nobsAll` guard fired first), so this was a regression introduced by making the path reachable. Now declines. Note the fold-in cannot simply be repeated at that entry: it is handed a theta/eta/omega that are deliberately *not* the inner problem's point.

**2. Relaxation was taken away from the subjects it was tuned on** (@mattfidler's review). rxode2's bad-solve test asks whether the *integration* failed; a solve can pass it and still return non-finite lhs, and the new `is_finite` flag sent that subject **straight to FD**, skipping rung 2. A non-finite read now earns the same relaxation ladder and the same per-subject counter, and only falls to FD if that fails. `fit$env$nOuterSolveRelaxed` counts the rescues.

I first tried folding the read into the retry *predicate*. **Reverted** -- it changes when `odeSwapIndBadSolve` is sampled relative to `iniSubjectE`/`calc_lhs`, which perturbs which subjects are retried on solves that are fine today: 10 failures across `test-focei-fast-grad`, one gradient declining outright. The shipped form is a rescue *after* the normal path, so it cannot touch a subject that already reads finite.

**3. Exact analytic zero no longer sets `zeroGrad`** (@mattfidler's review; I had flagged the same thing). The FD branches set it on `g == 0.0` because a finite difference essentially cannot land on exact zero, so there it is a reliable "dead direction" tell. An analytic gradient reaches exact zero by ordinary algebra, so copying that test fired `thetaResetZero()` on `fast=TRUE` fits with nothing wrong. The **non-finite** half is kept -- an FD substitution really can put a NaN into an otherwise analytic sum.

**4. Per-subject shi `h` restored** (@mattfidler's review). The shared step (one search on the summed -2LL) is replaced by a per-subject search again; the searches parallelize like the legs. This also brings back the **median-step repair** of a clamped step, which a single shared step made impossible.

Measured, theo_sd, max rel err of the substituted gradient vs the all-analytic one:

| flagged | 1/12 | 2/12 | 6/12 | 11/12 | 12/12 |
|---|---|---|---|---|---|
| shared step | 3.688% | 3.637% | 4.012% | 3.569% | 3.577% |
| per-subject | 3.688% | 3.658% | 3.981% | 3.562% | 3.572% |

**No measurable difference on this dataset.** Recording that rather than claiming a win: it says the ~3.6% residual (concentrated in `tka`) is *not* a step-size effect, so it lives in the profile likelihood's noise floor -- the inner re-optimization. Per-subject `h` is restored for the badly conditioned case and the clamp repair, not on the strength of these numbers.

**5. `foceiOuterFdInd_` early-returned on `nth == 0`**, making the omega block unreachable (Copilot's inline comment). Guarded on `ntheta + omegan`.

**6. Uninitialized read.** `E.f`/`E.R` are `set_size` (not zeroed) and the `is_finite()` scan ran before `ko == nobs` was checked. Result was unaffected; the read was undefined. Gated.

**7. Step-cache keying was an accident.** The omega phase re-derived its subject set from `ids0` with a second copy of the theta filter, while `fdStepCacheGet` keys the store on that set -- a divergence would have made each phase invalidate the other's steps. Now shares the vector.

**8. `.foceiGradDirect` discarded correct gradients.** `.foceiAnalyticGradSetup` returns NULL for any `fix()`ed theta, and the helper then returned NULL *after* the gradient had been computed -- reporting "no analytic gradient" for one that existed. This is what made the analytic path untestable on exactly the models where full-theta and free-parameter indexing differ. Returns the unnamed values now, which let me add the **`fix()`ed-theta test** -- the case the step-store indexing fix targets and that every other case here misses.

**9. Test hardening.** `all(is.finite(NULL))` is `TRUE`, so several assertions passed vacuously when the route declined; lengths are asserted first. The clamped-step test only checked `is.finite(stepClamped)` (true of 0) -- it now asserts the clamp fired *and kept* firing.

**10. Median/MAD.** Both outlier passes took `v[n/2]` -- the upper of the two middle order statistics for even `n`, the common case here -- while citing the Iglewicz-Hoaglin formula. One shared helper, stating the estimator the comment already claimed.

## Open, not addressed

- **The FD fold-in is still not DRY, and the autoencoder route still only declines.** Making it shared needs `foceiOuterFdInd_` parameterized by its point (theta, per-subject reference etas, omega) instead of reading `op_focei` globals. That is the remaining piece.
- **The outlier test's reference distribution** includes the FD slopes being judged, and `outerFdRef` fails *open* (a `gMap` arity mismatch silently degrades it to FD-only, with no counter). More fundamentally, the test assumes per-subject slopes are exchangeable; they scale with each subject's information content, so on unbalanced data a legitimately large slope can be flagged and TV-refined.
- Analytic subjects are never tested for outlier status -- correct for the *repair* (their slopes are exact) but it means an extreme analytic slope inflates the MAD and desensitizes the test for the FD subjects, with nothing able to remove or record it.

mattfidler:
Phase 2 pushed (`3d4f90ac7`): the FD fold-in now takes its point explicitly, so **one implementation serves both entries**.

`foceiOuterFdIndCore()` takes an `FdIndPoint` (theta, per-subject reference etas, an optional Omega override, whether omega is in scope). `foceiOuterFdInd_` passes all defaults, so the FOCEi path is byte-for-byte unchanged -- I verified that as a no-op *before* wiring anything else. `foceiGradPooledDirect_` now **substitutes instead of declining**, with the caller's Omega installed per worker through the same `thread_local` `OmegaScope` the omega phase already used (`_omegaOv` is thread_local, so a master-thread scope would not reach the OpenMP workers).

Omega directions are deliberately *not* differenced at that entry: a perturbed `Omega^-1` has to go through `setOmegaTheta()`, which needs the estimation-scale omega vector, and the entry is handed `Oi` as a matrix. Those kernel slots are set **NA** rather than left at an incomplete analytic sum, so a consumer that reads them fails visibly. The VAE M-step regresses over thetas (`regNames` is "the thetas the M-step regresses") and drops the omega names before its own finite check, so it is unaffected.

**One thing to be aware of, stated plainly: the pooled fold-in is not covered by a test.**

I wrote one driving `.vaeInnerSetup` + `.vaeGradEval`, and it **segfaults inside `foceiGradPooledDirect_`** -- `memory not mapped` at `.Call(_nlmixr2est_foceiGradPooledDirect_)`. It does so with **no subject flagged**, i.e. with none of the new code reached, and it reproduces on the previous commit (`6e7a581d4`) after a clean rebuild. So it is pre-existing in that setup sequence -- either a latent problem there or my test driving it wrongly -- and not a regression from this work. I removed the test rather than ship it red. Worth someone who knows the VAE setup path taking a look: a segfault rather than an error is not a good failure mode for an internal entry.

Everything else is green: the FOCEi fallback file (39 assertions), `focei-fast-grad`, `focei-fd-fallbacks`, `focei-fast-methods`, `agq-fast-grad`, `vae-nonmutheta-grad`, `vae-grad-fit` -- 0 failures -- and thread invariance still bit-identical at 1 vs 4 threads with two subjects flagged.

mattfidler:
Phase 5 pushed (`8a5ec513b`): the outlier test's reference distribution.

It pooled the analytic slopes with the FD slopes it was **judging**, so a flagged subject contributed to its own centre and scale. That is fine while the flagged set is a small minority -- median/MAD only break down near 50% -- but this path exists for fits where subjects fail, which is exactly where the assumption stops holding. The reference is now the exact analytic slopes alone, falling back to including the FD slopes only when fewer than 3 analytic ones exist, and **that fallback is counted**: `fit$env$nFdOutlier[["refDegraded"]]`. It also makes `gradPooledCore`'s `gMap`-arity check visible -- when that leaves `outerFdRef` empty the reference silently degraded, and nothing said so.

**The measurement is the interesting part** (theo_sd, forced flags):

| flagged | nOuterFdInd | params | chartrandSlopes | refDegraded |
|---|---|---|---|---|
| none | 0 | 0 | 0 | 0 |
| 2/12 | 8 | 0 | 0 | 0 |
| 12/12 | 48 | 14 | 20 | 28 |

With 10 exact slopes to judge against, **nothing is flagged**. With no analytic reference at all, 14 parameters flag and 20 slopes get TV-refined -- and the resulting gradient is **no more accurate** than the 1/12 case (3.572% vs 3.688% max relative error).

So on this dataset the Chartrand pass only ever fires in the degraded, self-referential state, and the refinements it triggers buy nothing measurable. That is a reason to look hard at whether the pass should run at all when `refDegraded > 0` -- but that is a behaviour question separate from making the state visible, so I have recorded it rather than acted on it.

Accuracy is unchanged to 5 significant figures at every flagged count, and the related suite plus thread invariance stay green.

mattfidler:
**Correction to my earlier measurements -- please disregard the theo_sd numbers as evidence about the FD refinement.**

@mattfidler pointed out that theo_sd is the wrong model to evaluate the Chartrand TV pass on, and he is right. It is a smooth, well-conditioned 1-cmt surface; the TV refinement exists for stiff/noisy profile likelihoods where the O(h^2) term is material. Running it there cannot discriminate, so "20 refinements, no accuracy gain" says nothing about whether the pass works. I withdraw that conclusion.

Two things I said that rested on it, both withdrawn:

1. **The suggestion to gate the Chartrand pass on `refDegraded`.** That was built on the theo_sd observation that the pass only fired in the degraded state and bought nothing. With the measurement invalidated there is no case for it -- and the argument plausibly runs the other way, since a fit where subjects are failing is exactly the hard surface TV regularization is meant for. `refDegraded` remains useful as a **diagnostic** (it makes a silently weakened outlier test visible, including `gradPooledCore`'s fail-open `gMap` arity path); it should not be read as a proposal to change behaviour.

2. **"Per-subject `h` makes no measurable difference"** (the 3.688%/3.658%/3.981% table). Same flaw: per-subject steps matter for badly conditioned subjects and theo_sd has none, so both routes difference the same smooth function and agree by construction. That number is not an argument against `fdIndividualStep=TRUE`, and I should not have presented it as a qualification.

What the theo_sd numbers **do** still support, because they are about plumbing rather than accuracy: the substitution runs, is bit-identical across thread counts, survives a `fix()`ed theta, and is unchanged by the reference-distribution change. Those are correctness-of-wiring claims, not claims about differencing quality.

No code is affected -- nothing was gated on those readings. What is missing is a model that can actually discriminate: a stiff or otherwise difficult likelihood surface where the flagged subjects are genuinely badly conditioned. `fdOutlierZ` and `fdIndividualStep` are in place to make that comparison runnable once there is a fair test case.

mattfidler:
Scoping note, so my previous comment does not read as an open item against this PR: evaluating the Chartrand TV refinement and the per-subject step on a **stiff / genuinely difficult likelihood surface** is a separate **nlmixr2est** follow-up, not work for #868.

(Corrected: an earlier version of this comment said rxode2. It is nlmixr2est -- all of this code lives in `nlmixr2est/src/inner.cpp`.)

Nothing on this branch depends on that evaluation. `fdChartrand` keeps its existing default, `fdIndividualStep` defaults to per-subject, and `refDegraded` is a diagnostic counter only -- no behaviour is gated on any of the readings I withdrew. `fdOutlierZ` and `fdIndividualStep` exist so that comparison is *runnable* when someone takes it up; they are not waiting on it.

So the open items on this PR proper are just the two already listed: the `odeSwapCheckLhsWidth` probe can still fault on a setup whose parameter layout does not match the pool (it calls `calc_lhs` to probe, so it can crash before it can answer -- the lhs and state buffers are now checked, `par_ptr` is not), and the outlier test's exchangeability assumption on unbalanced data.

mattfidler:
## Summary of the changes I pushed to this branch

Ten commits on top of the original, all built and tested as they landed. Follow-up question split out as #869. Nothing merged.

### The one silent-wrong-answer bug

`foceiGradPooledDirect_` (the entry `est="vae", nonMuTheta="grad"` uses) returned the kernel sum with a flagged subject's column still **zero**. Zeros are finite, so the `is_finite()` guard passed a gradient silently missing an entire subject to the M-step. `main` declined here, so this was a regression introduced by making the fallback reachable. Found by a Copilot CLI review.

### @mattfidler's review points

1. **Relaxation was taken away from the subjects it was tuned on.** rxode2's bad-solve test asks whether the *integration* failed; a solve could pass it and still return non-finite lhs, and that subject went **straight to FD**, skipping rung 2. It now loosens and re-reads on the same ladder and per-subject counter. `nOuterSolveRelaxed` counts the rescues. (I first folded the read into the retry *predicate* -- **reverted**: it changes when `odeSwapIndBadSolve` is sampled relative to `iniSubjectE`/`calc_lhs` and broke 10 tests.)
2. **The FD fold-in is DRY again.** `foceiOuterFdIndCore()` takes an explicit point (theta, per-subject reference etas, optional Omega override, omega-in-scope). One implementation serves both entries; the pooled entry substitutes instead of declining. The FOCEi path was verified unchanged as a no-op first.
3. **Per-subject shi `h` restored**, with the median-step repair a shared step made impossible.
4. **An exact analytic zero no longer sets `zeroGrad`.** A finite difference essentially cannot land on exact zero, so there it means "dead direction"; an analytic gradient reaches it by ordinary algebra, and copying the test fired `thetaResetZero()` on healthy `fast=TRUE` fits. The non-finite half is kept.

### New options (all default to current behaviour)

`fdIndividualStep`, `fdOutlierZ`, `fdChartrandAll`, `fdOutlierAny`, `fdRefine` (`"chartrand"`/`"lanczos"`/`"richardson"`, restored from `3485a22b8`) plus `fdLanczosM`/`fdRichardsonR`/`fdRichardsonV`. All in `foceiControl()` and mirrored on `vaeControl()`.

`fdOutlierZ` is the load-bearing one: at the conventional 3.5 cut a well-behaved fit never fires the outlier pass, so **the whole refinement was unreachable and therefore untestable**. Invariant across all of them: a subject whose analytic gradient is available is never recomputed -- `isOut` is sized by the flagged set, so that holds structurally.

### Other defects fixed

- `foceiOuterFdInd_` early-returned on `nth == 0`, making the omega block unreachable (Copilot inline comment)
- `is_finite()` scan ran before `ko == nobs`, reading uninitialized `set_size` memory
- The omega phase re-derived its subject set with a second copy of the theta filter, while the step cache keys on that set -- a divergence would have made each phase invalidate the other's steps
- `.foceiGradDirect` discarded a **correctly computed** gradient whenever the naming setup declined, which `.foceiAnalyticGradSetup` does for any `fix()`ed theta -- which is what made the analytic path untestable on exactly the models where full-theta and free-parameter indexing differ
- Both outlier passes took `v[n/2]` (the upper order statistic for even `n`) while citing the Iglewicz-Hoaglin median; one shared helper now
- The outlier reference pooled the FD slopes it was *judging*; it now prefers the exact analytic slopes and counts (`refDegraded`) when it cannot
- `fdRefine` did not survive a control round-trip (`match.arg` on an integer yields `NA` **silently**), so a re-passed control quietly used the default estimator
- `odeSwapCheckLhsWidth` checked the lhs buffer it writes but not the state buffer it reads
- `foceiGradPooledDirect_` lacked the live-problem preconditions `analyticOuterGradDirect` has -- including one that mattered alone: a short `thVals` silently differenced at the **wrong theta**

### Verification

`test-focei-outer-fd-fallback.R` is 87 assertions (was 26), including a `fix()`ed theta, the VAE fold-in through a real fit (`nRegGrad > 0`, `nRegFallback == 0`), the clamp firing *and continuing* to fire, and all three estimators. `focei-fast-grad`, `focei-fd-fallbacks`, `focei-fast-methods`, `agq-fast-grad`, `focei-ll-fast-grad`, `cov-analytic`, `vae-nonmutheta-grad`, `vae-grad-fit`: **0 failures**. Thread invariance bit-identical at 1 vs 4 threads with two subjects flagged.

### Still open

- `odeSwapCheckLhsWidth` probes by *calling* `calc_lhs`, so it can fault before it can answer. Lhs and state buffers are now checked; `par_ptr` is not, and validating the parameter layout needs someone who owns that probe.
- The outlier test's exchangeability assumption on unbalanced data (carried into #869).

### On the record

Measurements I reported from theo_sd about *differencing quality* are **withdrawn** -- it is a smooth, well-conditioned surface and cannot discriminate. That covers "the Chartrand pass buys nothing" and "per-subject `h` makes no difference", and the `refDegraded` gating idea that rested on the former. The theo_sd runs still support the *plumbing* claims (the substitution runs, is thread-invariant, survives a `fix()`ed theta). No code was gated on the withdrawn readings.

mattfidler:
Both remaining open items are now tracked as issues, so this PR has none of its own outstanding:

- **#869** -- which FD refinement (`richardson` / `lanczos` / `chartrand`) is right on a stiff likelihood surface, and whether Chartrand's advantage is a property of the surface or of the pool setup at the time it was measured.
- **#870** -- `odeSwapCheckLhsWidth` can segfault: it probes by calling `calc_lhs`, which reads `par_ptr` **by index**, so the size guards added here (state buffer, and parameter *count* via the new `odeSwapNpars`) do not close it. Two models with the same parameter count but different order still mis-read. Verified the crash still reproduces after the guards -- the fix needs parameter-NAME provenance, not counting.

Last two commits on the branch:

- **outlier exchangeability** -- `foceiControl(fdOutlierScale=)`, default `TRUE`, tests the **per-observation** slope on both sides (the finite differences and the exact analytic reference). Only the test is scaled; the gradient is bit-identical either way. Safe as a default because a balanced design divides every slope by the same count, which leaves a modified z-score unchanged -- the test asserts exactly that on theo_sd (identical objf and identical outlier count on/off), then thins four subjects to 3 observations against the others' 11 so the two settings can genuinely disagree.
- **`fdRefine` round-trip fix** -- it was returned as an integer code, and a control is re-passed through `foceiControl()` where `match.arg()` on an integer yields `NA` *silently*, so a round-tripped control quietly used the default estimator.

`test-focei-outer-fd-fallback.R` is now 97 assertions (from 26). `focei-fast-grad`, `focei-fd-fallbacks`, `vae-nonmutheta-grad`, `agq-fast-grad`: 0 failures.

mattfidler:
## Review round on the branch: Copilot clean, Antigravity failed to run

Ran an adversarial review pass over the whole branch diff. **Copilot CLI (gpt-5.4): 3 rounds, now returns "no new defects"** with (a)/(b)/(c) all confirmed. **Antigravity did not produce a review** -- two attempts both ended in `Error: timeout waiting for response` after ~20 min with no output. Not reported as a clean review, because it is not one.

### Fixed this round

**Mine, both in the new `fdOutlierScale` work:**

1. `_fdThNobs` was pushed in `_fdThIds` order but read by **row of `out`**. Those coincide only while every id passes the `id < 0 || id >= nsub` filter, so one filtered id would have shifted every later subject's observation count. Row-indexed now.
2. The scaling is **theta/sigma only**. Those enter through the observation likelihood, so a slope scales with observation count. **Omega does not** -- it reaches an individual likelihood through per-subject prior terms (`eta' Omega^-1 eta`, `log|Omega|`), so dividing omega slopes by `nobs` introduced exactly the distortion the scaling exists to remove.

**Copilot, confirmed and fixed:**

3. `fdBuildOutlierRef`'s **degraded fallback** pushed raw `out(k,col)` into the reference while the candidates were scaled -- so once fewer than 3 analytic slopes were available, the modified-z test compared different units.
4. The `fdOutlierAny` and `fdOutlierScale` tests **could both have passed against dead code**. Copilot refuted my first hardening attempt too, and was right: `fdOutlierAny` is gated on `&& !any`, so driving the cut to ~0 (making the FD side fire first) guarantees the analytic branch never runs.

### An incidental finding worth flagging

Measuring the counters to write those assertions turned up this, on **balanced** theo_sd at the **default 3.5 cut**:

| | params | chartrandSlopes | analyticTrigger |
|---|---|---|---|
| `fdOutlierAny=FALSE` | 0 | 0 | 0 |
| `fdOutlierAny=TRUE` | 10 | 0 | 10 |

The **exact analytic** per-subject slopes are already outliers against each other, on a balanced dataset, at the conventional cut. That is the exchangeability problem showing up directly in a counter rather than as an argument -- and it is why `fdOutlierScale` defaults on.

It also pins an interaction: `fdOutlierAny` alone **detects but refines nothing** (`chartrandSlopes == 0`), because only outlying *finite differences* are eligible. It changes the gradient only when paired with `fdChartrandAll`. Both are now asserted -- "the option fired" and "the option changed the answer" are different claims and only the second matters.

### Rejected after checking

Copilot claimed the new controls do not round-trip (`assertLogical` rejecting the integer fields). `assertLogical(1L)` does reject in isolation, but `do.call(foceiControl, foceiControl(fdRefine="richardson", fdOutlierScale=TRUE))` **succeeds** -- `foceiControl` does not re-run those assertions on an already-built control, which is why the pre-existing `fdChartrand = as.integer(fdChartrand)` has always been fine. Verified rather than taken on trust.

### Noted, not fixed

`foceiAnalyticGradPooled_` leaves a second kernel slot at zero when one parameter maps to two (an estimated boxCox/yeoJohnson lambda) and a subject is flagged. Real, but Copilot independently confirmed that entry has **no R callers** -- only the generated wrapper in `RcppExports.R` -- so it is dead code from the R-route era.

### Branch status

**15 commits**, all pushed. `test-focei-outer-fd-fallback.R` is **106 assertions** (from 26 originally). `focei-fast-grad`, `focei-fd-fallbacks`, `focei-fast-methods`, `agq-fast-grad`, `vae-nonmutheta-grad`, `vae-grad-fit`: **0 failures**. Thread invariance bit-identical at 1 vs 4 threads.

No open items on the PR itself -- the two that remained are tracked as **#869** (which FD estimator on a stiff surface) and **#870** (`odeSwapCheckLhsWidth` can segfault; its size guards cannot close it because `calc_lhs` reads `par_ptr` by index).

mattfidler:
@hiddevandebeek  since you started this, please let me know if you think this is ready yet.

hiddevandebeek:
@mattfidler I will have a look at it today

mattfidler:
I need this code for `nn()`  It will be merged as part of updating #738 .

I'm guessing you are OK with the code.

mattfidler:
> **Note: this comment was generated by Claude (Claude Code, Opus 5), driven by @mattfidler.** The analysis and the numbers below were produced and run by the model; treat the conclusions as a starting point for discussion rather than a reviewed result.

## `foceif` stops early -- it's the outer optimizer default, not the analytic gradient

While comparing `foceif` against `focei` across several models, `foceif` showed a suspiciously *deterministic* number of outer iterations (tracking the parameter count rather than the data), and the best-ranked model differed between the two methods -- the worst model under `foceif` was the best under `focei`. Here is what that turned out to be.

### Root cause

`.foceiFastCtl()` (`R/foceiFast.R:14-27`) does not only set `fast=TRUE`; it also lets the outer optimizer re-default. `R/foceiControl.R:1290-1293`:

```r
outerOpt <- if (isTRUE(fast)) "lbfgsb3c" else "bobyqa"
```

So `focei` vs `foceif` is **bobyqa vs lbfgsb3c**, not FD-gradient vs analytic-gradient. (`focei`'s default became `bobyqa` in 6649d7a10, "Actually use bobyqa unless asked". The roxygen at `R/foceiControl.R:194` still documents the `fast=FALSE` default as `nlminb`, so the docs and the code disagree.)

The two optimizers stop on very different criteria:

* **bobyqa** stops when its trust-region radius shrinks to `rhoend = 10^-sigdig`, in *scaled parameter space*. It must keep probing until moves that small stop helping, so its iteration count depends on the data and the surface.
* **lbfgsb3c** is configured with `lbfgsPgtol = 0`, which **disables the projected-gradient test**, leaving `factr` as the only stopping rule. `lbfgsFactr = 10^-sigdig / .Machine$double.eps`, so the effective test is: quit as soon as **one iteration's** relative OFV reduction falls below `10^-sigdig`. That is a per-step *improvement* test, not a stationarity test -- it fires after a handful of iterations no matter how far the gradient still is from zero, and the count tracks the problem dimension because that is roughly how long L-BFGS-B takes to build its limited-memory curvature model.

### Measurement

`nlmixr2data::theo_sd`, one-compartment ODE model (not `linCmt()`, so `fast=TRUE` stays in scope), 7 estimated parameters. nlmixr2est 7.0.3 / rxode2 5.1.7, default `sigdig=3`. `outerEvals` counts rows of `fit$parHistData` with `type == "Unscaled"`.

| run | OFV | outer evals | time |
|---|---|---|---|
| `focei` (bobyqa, FD) | 116.8136 | 59 | 6.8s |
| `focei` + `nlminb` (FD) | 116.8696 | 57 | 2.3s |
| **`focei` + `lbfgsb3c` (FD)** | **117.2286** | **8** | 1.6s |
| **`foceif`** (lbfgsb3c, analytic) | **117.1409** | **9** | 6.3s\* |
| **`foceif` + `nlminb`** | **116.8123** | **42** | 2.3s |
| `foceif` + `lbfgsFactr=1e5` | **116.8075** | 15 | 2.4s |

\* includes the augmented-model compile; the later `foceif` rows reuse it.

**Row 3 is the control that isolates the cause:** plain `focei` with ordinary *finite-difference* gradients also stalls at 8 evaluations once you hand it `outerOpt="lbfgsb3c"`. The analytic gradient is not implicated. The `foceif` run genuinely used the analytic path -- its `parHistData` is 9/9 `"Analytic Gradient"` with no FD/relaxed/Chartrand fallback -- and its `$runInfo` reports `"gradient problems with covariance; see $scaleInfo"`, which is the expected symptom of terminating away from the optimum.

Conversely, `foceif` + `nlminb` produced a **lower OFV than any other configuration tested including `focei`'s own default**, in about a third of `focei`'s wall time. The analytic gradient does what it promises as soon as the optimizer is allowed to converge.

### Consequence for model selection

A per-model shortfall of 0.3-1.1 OFV units, varying by model, is the same magnitude as the OFV differences being used to rank candidates -- which is enough to invert the ranking of a six-model set. Any `foceif` model comparison run with the current default is comparing stopping points, not optima.

### Workaround today

```r
nlmixr2(model, data, est = "foceif",
        control = foceiControl(outerOpt = "nlminb"))
# or keep lbfgsb3c and tighten its only active stopping rule:
control = foceiControl(lbfgsFactr = 1e5)
```

### Suggested fix

1. Default `fast=TRUE` to `outerOpt="nlminb"` rather than `"lbfgsb3c"`; on this problem it was both more accurate and faster.
2. If `lbfgsb3c` stays a default anywhere, derive `lbfgsPgtol` from `sigdig` instead of leaving it at `0`, so there is a stationarity check at all rather than only a per-step improvement test.
3. Fix the `R/foceiControl.R:194` roxygen, which still says the `fast=FALSE` default is `nlminb`.
4. Add a regression test pinning the converged OFV across outer optimizers, so a future default change cannot silently move the optimum.

<details>
<summary>Reproduction script</summary>

```r
suppressMessages(library(nlmixr2est))
one.cmt <- function() {
  ini({
    tka <- 0.45; tcl <- log(c(0, 2.7, 100)); tv <- 3.45
    eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1
    add.sd <- 0.7
  })
  model({
    ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
    d/dt(depot) <- -ka * depot
    d/dt(center) <- ka * depot - cl / v * center
    cp <- center / v
    cp ~ add(add.sd)
  })
}
d <- nlmixr2data::theo_sd

run <- function(lbl, est, ctl) {
  t0 <- proc.time()[3]
  f <- suppressWarnings(suppressMessages(
    nlmixr2est::nlmixr2(one.cmt, d, est = est, control = ctl)))
  nOfv <- sum(f$parHistData$type == "Unscaled")
  cat(sprintf("%-28s ofv=%12.4f  outerEvals=%4d  time=%5.1fs\n",
              lbl, f$objf, nOfv, proc.time()[3] - t0))
  invisible(f)
}

run("focei (bobyqa, FD)",      "focei",  foceiControl(print = 0))
run("focei + nlminb (FD)",     "focei",  foceiControl(print = 0, outerOpt = "nlminb"))
run("focei + lbfgsb3c (FD)",   "focei",  foceiControl(print = 0, outerOpt = "lbfgsb3c"))
run("foceif (lbfgsb3c)",       "foceif", foceiControl(print = 0))
run("foceif + nlminb",         "foceif", foceiControl(print = 0, outerOpt = "nlminb"))
run("foceif + lbfgsb3c tight", "foceif", foceiControl(print = 0, lbfgsFactr = 1e5))
```

</details>

Issue #687, opened 2026-06-19
SAEM errors for subjects with dosing and NA observations


billdenney:
SAEM errors for subjects with dosing and NA observations, but FOCEI warns and ignores them. I think that the FOCEI behavior is preferred.

PR #899, opened 2026-08-10
fix(saem): score ll() observations by their own density in the objective and covariance (#871)

### LINKED ISSUE #871: saem: calc.COV/calc.2LL score ll() observations as Gaussian
`calc.2LL` and `calc.COV` (`R/saem_fit_aux.R:24`, `:262`) build

```r
g <- ares + bres * abs(fsave)   # :97 and :332
```

for **every** observation, and neither reads `saem.cfg$distEp` -- which is present
in the cfg (`R/saem_fit.R:706`). For an `ll()` endpoint `res_mod == 0`, so `ares`
keeps its init `10` and `bres` its `1`, and the per-observation log-density is
scored as a normal mean with SD `10 + |ll|`.

This is **not** fsaem-specific: it is wrong for every `ll()` model fit with plain
`est="saem"`, and it predates the fsaem work. `covMethod` defaults to `"sa"`
(`R/saemControl.R:411`), but `.saemSpliceLinFimVar` splices linFim's variance
block into `sa`/`fim`, and linFim is where every other method lands when
`.saemFimToCov` returns NULL -- so the defect reaches the default path anyway.

## The two functions need different fixes

* **`calc.2LL` is tractable.** Excluding `ll()` rows from `g` is not enough:
  `R/saem_fit_aux.R:99` computes `DYF[ind.io] <- -0.5*((yobs-f)/g)^2 - log(g)` for
  every row, so an `ll()` row would still be scored by a normal density, just with
  a different SD. Bypass that density for `ll()` rows and assign the model's own
  `ll()` output (the prediction column already *is* the per-observation
  log-likelihood) straight into `DYF`.
* **`calc.COV` is not fixable the same way.** It linearizes `dopred()` as if it
  returned a transformed mean -- `DF <- (f1 - f0) / dphi[id, ]`
  (`R/saem_fit_aux.R:334`) -- and builds a Gaussian `Vi` from it. For an `ll()`
  endpoint `dopred()` returns the log-likelihood, so that Jacobian is of the
  log-likelihood, not of a mean; patching `g` cannot make the covariance right.
  This needs either a genuinely different covariance route for `ll()` endpoints,
  or an explicit gate that refuses `linFim` for them and falls back to a method
  that is correct. **Decide which by measurement before writing code.**

## Verification

Against the model's own Gaussian twin: an `ll()` written as the exact normal
log-density must give **the same SEs directly** as the equivalent `add()` model.
The `0.5*log(2*pi)` per observation is a `-2LL` constant offset, so it applies to
the `calc.2LL` comparison only -- it is not an SE offset.

Run `tests/testthat/test-saem-loglik.R`, `test-saem-cov-sa.R`,
`test-saem-cov-analytic.R`.

---
Split out of the f-SAEM follow-up backlog because it is a plain-`saem` defect.
Context: #845.

mattfidler:
Closes #871.

`calc.2LL` (`R/saem_fit_aux.R`) and `calc.COV` build a Gaussian residual SD

```r
g <- ares + bres * abs(fsave)
```

for **every** observation.  A general log-likelihood endpoint (`ll(name) ~ expr`)
has `res.mod == 0`, so the residual M-step never runs (`src/saem.cpp`, the
`distribution != 4` guard), the residual warm start returns early, and nothing
zeroes `ares`/`bres` -- they keep the placeholder `10` and `1` they are
initialized with.  Every per-observation log-density was scored as a normal mean
with SD `10 + |ll|`.

This is not `fsaem`-specific and predates it: it is wrong for every `ll()` model
fit with plain `est="saem"`.

## Blast radius

Both defaults route through the broken code:

* `logLik=FALSE` is the `saemControl()` default, so `env$objective` is `NA` and
  the first access of `fit$objf`/`$AIC`/`$BIC`/`$logLik` runs
  `.nmObjEnsureObjective()` -> `setOfv(., "gauss3_1.6")` -> `calc.2LL`.
* `covMethod="sa"` is the default, but for `distribution == 4` the trailing
  log-sigma2 slot of the Louis FIM is hard-zeroed, so `Ha`/`HaSa` are singular,
  `.saemFimToCov()` returns `NULL`, and every `ll()` fit fell back to `linFim` --
  which is `calc.COV`.

## The two functions needed different fixes

**`calc.2LL`.** Excluding `ll()` rows from `g` is not enough: the row would still
be scored by a normal density, just with a different SD.  Those rows now take the
model's own per-observation log-likelihood (which is what `dopred()` returns for
such an endpoint) straight into `DYF`, and the `log(2*pi)` normalizer is applied
only to the normally-distributed rows, since an `ll()` expression already carries
its own constant.

**`calc.COV`.** It linearizes `dopred()` as if it returned a transformed mean --
`DF <- (f1 - f0)/dphi` -- and builds a Gaussian `Vi = DF Omega DF' + diag(g^2)`
from it.  For an `ll()` endpoint that Jacobian is of the log-likelihood, not of a
mean, so patching `g` cannot make the covariance right.

An intermediate form was tried first: because `DF` is then the per-observation
*score*, setting `g = 1` turns `DF' diag(g^2)^-1 DF` into the outer-product
(BHHH) information, with no other change.  Measured, that is biased -- it uses the
realized residual where the Gaussian route uses its expectation, and at the EBEs
the residuals are shrunk (SE ratios 1.14-1.88 against the Gaussian twin).

What shipped is the **observed information**: central differences of `dopred()`
with respect to phi, marginalized over the random effects with the Schur
complement

```
M = H - H[, i1] (Omega^-1 + H[i1, i1])^-1 H[i1, ]
```

which is exactly what the Gaussian branch already computes through `Vi` and the
Woodbury identity, so a normal model agrees by construction.  `M` is returned as
a matrix square root, leaving `.nlmixr2RobustCov()` and its rank-deficiency (`NA`
rows/columns) handling untouched.  The Omega variance block follows the same
identity, `tr(Vi^-1 dVi/da Vi^-1 dVi/db) = tr(E_a M11 E_b M11)`.

## Verification

The issue asks for the model's own Gaussian twin.  Both are measured.

**Objective, against a closed form.**  An exponential time-to-event model's
marginal likelihood is a one-dimensional quadrature:

| | -2LL |
|---|---|
| before | 1124.05 |
| after (`nnodesGq=13`, `nsdGq=5`) | 1392.08 |
| exact | 1392.157 |

**Standard errors, against the Gaussian twin** (theo_sd, an `ll()` written as the
exact normal log-density vs the equivalent `add()` model, residual SD fixed in
both so they estimate the identical free parameter set) -- ratio `ll/add`:

| | tka | tcl | tv | om.eta.ka | om.eta.cl | om.eta.v |
|---|---|---|---|---|---|---|
| before | 4.04 | 11.44 | 8.47 | 5.27 | 80.49 | 21.09 |
| after | 0.999 | 0.997 | 1.006 | 1.003 | 0.994 | 1.015 |

The two objective function values now agree outright rather than up to a
constant, matching what FOCEi already does (`test-focei-llik.R`) and improving on
npag, whose NEWS documents a residual `0.5*log(2*pi)` per observation.

A normal model is unchanged: the mask is all-`FALSE`, every expression reduces to
its current form, and the Gaussian branch of `calc.COV` is not touched.  All 18
`test-saem-*` files pass.

## Two things worth calling out

**`res.mod` was being dropped post-fit.** `R/focei.R` trims `saem.cfg` to a
whitelist that did not include it, so the per-observation mask could not be built
on the path that actually reports `$objf`.  It is now kept.

**`covMethod="sa"`/`"fim"` are refused for a general likelihood.** Unblocking
them was tried -- dropping the degenerate log-sigma2 slot so the remaining
mu/Omega block inverts, on the reasoning that those scores come from
`p(phi|mu,Omega)` alone.  Measured against the exact observed information that is
wrong: it returns SEs **0.41 and 0.14** of the true values where `linFim` lands
at **0.93 and 0.82**.  The complete-data Louis correction subtracts the
information lost to the latent eta as `Var[score]` from a handful of MCMC chains,
and a general likelihood has no residual error to anchor it -- the same
instability already recorded for proportional/combined residuals.  So the
combination is refused up front with a message saying why, and the SA covariance
phase is no longer run for such a model at all.

## Reviewed

Independently reviewed with Antigravity (Gemini).  Three findings fixed: the
second-difference step floor (`1e-10` squares to `1e-20`), `.saemLlObsMask()`
guessing where it cannot know, and the Omega variance block reading an unclamped
`M11`.  Two rejected with evidence: `ind.io` misnumbering a subject with no
observations is unreachable, because such a subject is dropped upstream (#687)
before `saem.cfg` is built; and the unprotected `solve(iOmega + H[i1, i1])`
matches what the Gaussian branch already does on a spectral-decomposition
failure, with the caller handling both identically.

## Not in this PR

* **#903** -- `calc.2LL`'s transform-both-sides Jacobian at `R/saem_fit_aux.R` subtracts
  `2*powerL` where FOCEi adds it (`src/inner.cpp`, `lik += fInd->tbsLik`), which
  would affect every `lnorm`/`boxCox`/`yeoJohnson`/`logitNorm` saem objective.
  Suspected, **not confirmed** -- the check via a `lnorm`/`ll()` twin was
  inconclusive because that quadrature has no log-sum-exp and overflows to `Inf`
  on such a model.  Different blast radius; needs its own issue and its own
  reference values.
* **#904** -- the residual entries of `calc.COV`'s variance block (`dVi/da`, `dVi/db`) are
  not masked to the owning endpoint, so they are already wrong for a
  multi-endpoint Gaussian model.  Untouched here; a pure-`ll()` model has no
  residual parameters, so the block is skipped rather than reported wrong for a
  future mixed model.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -35,6 +35,30 @@
 
 ### Estimation
 
+- Fixed `est="saem"` scoring a **general log-likelihood endpoint (`ll()`) as a
+  Gaussian observation** in both its objective function and its standard errors.
+  Such an endpoint estimates no residual error, so the residual step never runs
+  and the placeholder values it starts from survive: every log-density was scored
+  as a normal mean with standard deviation `10 + |ll|`.  The reported
+  `objf`/`logLik`/`AIC`/`BIC` and every standard error were meaningless, on the
+  default path -- `covMethod="sa"` cannot be computed for these models and
+  already fell back to the linearized Fisher information, which is where the
+  defect lives.  An `ll()` row now contributes its own log-density to the
+  objective, with the `log(2*pi)` normalizer applied only to normally-distributed
+  rows, and contributes the observed information of that log-density to the
+  covariance.  On an exponential time-to-event model with a closed-form marginal
+  likelihood the reported -2LL goes from 1124 to within 1e-3 of the exact 1392;
+  against the Gaussian twin of a one-compartment model (an `ll()` written as the
+  exact normal log-density versus the equivalent `add()` model) the standard
+  error ratios go from 4.0-80.5 to 0.99-1.02, and the two objective function
+  values now agree outright rather than up to a constant.
+
+- `saemControl(covMethod=)` `"sa"` and `"fim"` now say plainly that they do not
+  apply to a general log-likelihood endpoint and use the linearized Fisher
+  information, instead of reporting that the covariance "could not be computed".
+  The stochastic-approximation covariance phase is also skipped for such a model
+  rather than run and discarded.
+
 - `saem`'s uninformative-eta detection
   (`saemControl(handleUninformativeEtas=TRUE)`, the default) could freeze an eta
   that the data does inform.  The test asks whether perturbing an eta moves the
--- a/R/focei.R
+++ b/R/focei.R
@@ -3075,7 +3075,9 @@ attr(rxUiGet.foceiOptEnv, "rstudio") <- emptyenv()
       .saemCfg <- attr(.saem, "saem.cfg")
       # Delete unneeded variables
       .saemCfg2 <- list()
-      for (.v in c("i1", "nphi1", "nphi0", "N", "ntotal", "ix_endpnt", "y", "nmc", "niter", "opt", "inits", "Mcovariables")) {
+      # res.mod is kept because calc.2LL()/calc.COV() need it to tell an ll()
+      # observation from a normally-distributed one
+      for (.v in c("i1", "nphi1", "nphi0", "N", "ntotal", "ix_endpnt", "y", "nmc", "niter", "opt", "inits", "Mcovariables", "res.mod")) {
         .saemCfg2[[.v]] <- .saemCfg[[.v]]
       }
       attr(.saem, "saem.cfg") <- .saemCfg2
--- a/R/saem.R
+++ b/R/saem.R
@@ -238,7 +238,10 @@
                         odeRecalcFactor=rxode2::rxGetControl(ui, "odeRecalcFactor", 10^0.5),
                         maxOdeRecalc=rxode2::rxGetControl(ui, "maxOdeRecalc", 10^0.5),
                         indTolRelax=rxode2::rxGetControl(ui, "indTolRelax", TRUE),
-                        nSaCov=if (identical(rxode2::rxGetControl(ui, "covMethod", "linFim"), "sa"))
+                        # a general likelihood discards the SA covariance phase
+                        # (.saemCalcCov), so do not pay for it either
+                        nSaCov=if (identical(rxode2::rxGetControl(ui, "covMethod", "linFim"), "sa") &&
+                                     !.saemGeneralLik(ui))
                                  as.integer(rxode2::rxGetControl(ui, "nSaCov", 500L)) else 0L,
                         nres=ui$saemModNumEst,
                         perSa=rxode2::rxGetControl(ui, "perSa", 0.75),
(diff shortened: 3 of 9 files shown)


mattfidler:
Better, but could be even better, maybe apply from other method...

PR #910, opened 2026-08-11
fix(fsaem): map each structural theta to its own phi column (#875)

### LINKED ISSUE #875: fsaem: nonMuEtas make phiPopOk false -- measure, then gate or fix
**This one is contested -- measure before changing anything.**

`phiPopOk = (fsaemStructPos.n_elem == nphi1 + nphi0)` (`src/saem.cpp:3966`) **is**
false whenever `nonMuEtas` are present: `.structPos` (`R/fsaemInner.R:277`) counts
only structural thetas, while the SAEM phi vector also carries `nonMuEtas`
(`R/saemRxUiGet.R:161-162`, `R/saemRxUiGetModel.R:353`). So the fallback
`theta[fsaemStructPos(j)] = mprior_phi1(0, min(j, nphi1-1))` runs.

But `fsaemBuildInnerTheta` only overwrites `fsaemStructPos` entries, and nonMuEta
columns are appended **after** the structural roots. For a model with **no phi0**
the leading slots still line up 1:1 and the fallback happens to assign the right
values -- which is why an actual IOV `est="fsaem"` fit runs clean (`accRate 0.687`,
`nMapFail 0`, `nProp == nStep*nsweep*nmc*nsub`).

## Discriminating experiment

Add a **phi0** parameter (one with no eta) to an IOV model. The indices shift, the
fallback loads the wrong theta, and the correctness bug should appear.

Assert the **reconstructed inner THETA** on (a) a pure IOV model and (b) a
nonMuEta + phi0 model. Do *not* start from `fixef(fsaem)` vs `fixef(saem)` -- a
wrong inner theta costs acceptance and shifts the target without crashing, so
estimates are not a reliable pin.

## If confirmed

The gate is `length(ui$nonMuEtas) > 0L`, **not** IOV -- gating IOV alone leaves
every other `nonMuEtas` model broken. Note `attr(nlmixr2Est.fsaem, "iov") <- TRUE`
(`R/fsaem.R:96`) currently advertises IOV support.

## Separate defect found alongside

`.bounds <- if (.saemAnyGeneralLik(ui)) .fsaemPhi1Bounds(ui) else NULL`
(`R/fsaemInner.R:224`) -- so **a normal model with `lower=`/`upper=` on a mu-ref
theta gets no IMH bounds at all** and the proposal may wander outside the declared
bounds. Bounded parameters and `ll()` endpoints are orthogonal; the code ties them
together for no stated reason. Also, `.fsaemPhi1Bounds` (`:168-180`) maps
`.etaRows` through `ui$muRefDataFrame`, giving `NA` -> `nbd = 0` for IOV etas; if
that vector's length differs from the expanded `op_focei.neta`, `hasBounds`
(`src/inner.cpp:16238`) drops **every** bound silently.

There is no IOV fsaem test at all -- add one either way.

Refs #845

mattfidler:
Closes #875.

The issue asked to measure before changing anything, so the first commit adds
the instrument: `fsaemDiag()$lastTheta`, the inner `THETA` the last fast step
was re-parameterized with.  The kernel counters only show that the fast step
fired; a wrong phi -> theta map costs acceptance and moves the target without
erroring, so nothing existing could see it.

## Measured, before any fix

`nBurn=6`, `nEm=4`, `fastKernel="throughout"`, inner `THETA` read back from
`fsaemDiag()`:

| model | inner slot | got | fit's own value |
| --- | --- | --- | --- |
| IOV, every structural theta mu-referenced | IOV sd | `0` | live estimate |
| IOV + a no-eta theta (`tv`) | `tv` | `0` | `3.53` (accRate `0.499`) |
| `fix()`ed `tcl` (`literalFix=FALSE`, the default) | `tv` | `1.0` (tcl's value) | `3.51` |

So the issue's hypothesis is confirmed, and it is broader than it describes.
It is not only nonMuEtas + a phi0: **plain IOV is already wrong**, because the
IOV magnitude theta is itself a phi0, and so is **any `fix()`ed structural
theta**, which the existing "handles a fix()ed structural theta" test did not
catch.

The mechanism is the same in all three: `fsaemStructPos` and the phi columns
were walked in parallel, guarded by a length test (`phiPopOk`) that a nonMuEta
always fails, and the fallback it fell to walked `mprior_phi1`, which skips
phi0 columns entirely.

## The fix

Each structural theta is matched to its own phi column **by name** against
`ui$saemParamsToEstimateCov`, and the fallback is gone.  For a plain
mu-referenced model the map is the identity, so the permuted-ini finding this
code was built around still holds -- this is the phi vector itself (ini/ntheta
order), not the `muRefDataFrame` route that collapsed acceptance to 0.02.

After: every slot matches the fit's own value (`lastTheta` vs `fixef` max
`0.002`), and the IOV + phi0 acceptance rate goes **0.499 -> 0.922**.

The no-covariate R closure carried a second copy of the same wrong map and was
already unreachable (`fsaemNoCov` routes the SAEM loop straight to
`fsaemStepCpp_`); it is removed, and the "kernel installed" test in
`src/saem.cpp` is now `fsaemInstalled()`.

## Second bug found alongside

`est="fsaem"` on a model with **both a mu-referenced covariate and an IOV eta**
aborted the fit: the covariate inner absorbs each mu-ref intercept into a
per-subject data column and is sized from `muRefDataFrame`, so an eta with no
row there left it short -- `etaMat must have the same number of ETAs (cols) as
the model`.  It now degrades to standard SAEM, measured.

## The issue's bounds report

Both halves are **refuted by measurement**, so behavior is unchanged and only
the misleading comment is corrected:

- *"a normal model with `lower=`/`upper=` on a mu-ref theta gets no IMH bounds
  at all"* -- `fsaem` is an `unbounded` method, so `preProcessBoundedTransform`
  rewrites every finite structural-theta bound into an unconstrained internal
  parameter **before** the kernel is installed; the bound is enforced by the
  back-transform, not by the proposal.  Measured on `tka <- c(0, 0.45, 3)`:
  post-hook bounds are `-Inf/Inf` and `.fsaemPhi1Bounds` returns `nbd = 0,0,0`.
- *"`hasBounds` drops every bound silently on a length mismatch"* --
  `fsaemImhCheckDims` **errors** on `nbd.size() != 0 && != neta`.  And for an
  IOV model the length matches anyway (`neta = 4`, `length(nbd) = 4`).

## Tests

There was no IOV fsaem test at all.  New `test-fsaem-iov.R` (weekly batch 6,
beside `iov` / `iov-zero-eta`):

- plain IOV -- the inner must carry a **live** IOV sd, not the `0` a nonMuEta's
  phi column supplies;
- IOV + a no-eta theta -- the inner `tv` and the acceptance rate, which is the
  discriminator (`0.50` broken, `0.92` fixed);
- covariates + IOV -- must degrade, not abort.

`test-fsaem.R`'s fix()ed-theta test now also checks the inner `THETA`.  The
fsaem-vs-saem estimate comparison is deliberately a smoke band or absent, per
the issue's own guidance that estimates are not a reliable pin -- measured, the
fast kernel reaches `tcl = 1.02` on the IOV+phi0 model while plain SAEM is
still at `0.76` at the same budget.

All six fsaem test files pass (`test-fsaem`, `-iov`, `-inner`, `-multi`,
`-imh`, `-es`).

## Review

Three Antigravity (Gemini) passes.  One real finding: `test-fsaem-inner.R`
still asserted the removed no-covariate closure existed -- fixed.  Two further
claims about the covariate path were checked and **refuted by measurement**
rather than acted on:

- *`muRefCov=FALSE` leaves the covariate frozen* -- measured, `cl.wt` moves
  `0.5 -> 0.211` (fsaem) against saem's `0.272`, accRate `0.911`.  `Plambda` is
  never length 0; `nlambda` counts every mcov design cell.
- *a phi0 theta on the covariate path is frozen or overwritten* -- measured,
  `tv` moves `1.609 -> 3.519` against saem's `3.522`, accRate `0.918`.  `mcov`
  is numbered phi-column-major with the intercept first, which is exactly the
  order `saemParamsToEstimate` interleaves, so `match()` returns the right
  `Plambda` slot for an intercept as well as a covariate coefficient.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -139,6 +139,27 @@
   `combined2()` on such an endpoint -- which the control cannot override --
   degrades to standard SAEM instead (#874).
 
+- `est="fsaem"` now builds its Metropolis-Hastings proposal at the fit's own
+  population parameters when the model has a parameter the SAEM phi vector does
+  not line up with positionally.  The fast kernel filled the FOCEi inner's
+  `THETA` by walking the structural thetas and the phi columns in parallel, and
+  that pairing breaks three ways: a non-mu eta (every **IOV** eta is one) appends
+  a phi column with no theta, a `fix()`ed theta keeps its phi column but leaves
+  the structural list, and a theta with no eta shifts every later index.  The
+  inner was then handed some other parameter's value -- an IOV model got
+  `iov sd = 0`, an IOV model with a no-eta theta got `V = 1`, and a `fix()`ed
+  `tcl` gave `tv` tcl's fixed value -- which costs acceptance and moves the
+  target without erroring.  Each structural theta is now matched to its own phi
+  column by name; on an IOV model with a no-eta theta the acceptance rate goes
+  from 0.50 to 0.89 (#875).
+
+- `est="fsaem"` on a model with **both a mu-referenced covariate and an IOV**
+  (or any other non-mu) eta ran the standard SAEM kernel instead of aborting the
+  fit.  The covariate proposal's inner absorbs each mu-referenced intercept into
+  a per-subject data column and is sized from those, so an eta with no such
+  column left it short and the fit died with "etaMat must have the same number
+  of ETAs (cols) as the model" (#875).
+
 - `est="fsaem"` no longer replaces the fit's own control while installing its
   fast kernel.  The FOCEi inner behind the proposal was set up on the fit's `ui`
   rather than a copy, so the `foceiControl` it installs there *became* the fit's
--- a/NEWS.md
+++ b/NEWS.md
@@ -41,7 +41,9 @@
 
 - `est="saem"` fits now carry `fit$saemDiag`, the per-kernel Metropolis proposal
   and acceptance counts; `est="fsaem"` fits also carry `fit$fsaemDiag` with the
-  fast kernel's own step, proposal, acceptance and failure counts.
+  fast kernel's own step, proposal, acceptance and failure counts, plus
+  `lastTheta`, the inner `THETA` the last fast step was built with (the counts
+  show the kernel fired; only `lastTheta` shows what it was fired at).
 
 - Requires `rxode2` (>= 5.1.7).  The compatibility layer that also let this
   package build and run against 5.1.5 has been removed, so the event-sensitivity
--- a/R/fsaemInner.R
+++ b/R/fsaemInner.R
@@ -187,12 +187,13 @@
 
 #' Build the f-SAEM fast-simulation step for a SAEM fit and attach it to `cfg`.
 #'
-#' Sets up the FOCEi inner once and returns a closure the C++ SAEM loop calls
-#' each fast iteration with the current estimate.  The closure re-parameterizes
-#' the inner (theta = [structural fixed effects, residual] in THETA order; omega
-#' = current diagonal), optimizes the per-subject MAP + proposal covariance, and
-#' runs the independent Metropolis-Hastings kernel over the chains.
-#' @return `cfg` with `$fsaemStep` (closure) and `$fsaemInnerEnv` (keep-alive).
+#' Sets up the FOCEi inner once so the C++ SAEM loop can re-parameterize it each
+#' fast iteration (theta = [structural fixed effects, residual] in THETA order;
+#' omega = current diagonal), optimize the per-subject MAP + proposal
+#' covariance, and run the independent Metropolis-Hastings kernel over the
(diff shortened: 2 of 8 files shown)


mattfidler:
Follow-up filed as #911: the covariate proposal path still degrades to standard SAEM when a non-mu eta (IOV) is present. This PR only stopped it aborting; lifting the restriction means sizing the mprior-as-data inner by the model's full eta count and keying its mprior columns by phi1 column rather than by `muRefDataFrame` row.

Issue #916, opened 2026-08-11
saem: censoring is E-step only -- the residual M-step counts a BQL DV as observed


mattfidler:
#876 fixed the SAEM chain's censored observation loss, so the E-step now scores
an M2/M3/M4 row exactly the way `likInner0` does.  The **M-step still knows
nothing about censoring**.

`src/saem.cpp:2578` builds the residual sum of squares from `y_cur - f` over
every observation of the endpoint, and `src/saem.cpp:2898` divides it by that
endpoint's full observation count.  A BQL row contributes `(LOQ - f)^2` and a
count of 1, exactly as if the LOQ had been measured; neither place reads
`cens`/`limit`.

That biases the residual error LOW -- the censored rows sit at the LOQ by
construction, so their apparent residual is smaller than the real one.  Measured
on the Wang2007 M3 data in `tests/testthat/test-focei-cens.R`:

| | tvK | prop.sd |
| --- | --- | --- |
| focei | 0.757 | 0.280 |
| saem | 0.683 | 0.191 |

`tvK` agrees to under 10%; `prop.sd` is off by a third.  That is why the M3/M4
cases added in #876 pin `tvK` against focei but only bound `prop.sd` from above
-- a tighter tolerance there would fail for a real reason, not flakiness.

The standard treatment is data augmentation (Samson, Lavielle & Mentre 2006,
*Extension of the SAEM algorithm to left-censored data in nonlinear
mixed-effects model*): simulate a censored row's DV from the truncated normal
implied by the current `(f, g)` alongside phi, and let the M-step use the
simulated value.  That makes the residual estimate consistent rather than merely
less wrong.

## Action

1. Simulate the censored DV in the E-step, on the transformed scale: the lower
   tail for `cens = 1`, the upper tail for `cens = -1`, and both bounds under M4
   where `limit` supplies the other one.
2. Feed the simulated value to the residual SSR (`src/saem.cpp:2578`).  Leave
   `applyCensLoss` alone -- the chain's target is already right, and changing it
   would break the agreement with the FOCEi inner that lets `est="fsaem"` run on
   censored data without a gate.
3. Tighten the M3/M4 cases in `tests/testthat/test-focei-cens.R` to pin
   `prop.sd` against focei instead of only bounding it.
4. `est="fsaem"` needs no separate work -- it shares this M-step.

Refs #876

PR #917, opened 2026-08-11
fix(saem): score a censored observation the way FOCEi does (#876)

### LINKED ISSUE #876: fsaem: no censoring gate and no censoring test
`.fsaemSupported` (`R/fsaemInner.R:122-152`) has no censoring gate, so an
M2/M3/M4 dataset enters the fast kernel.

The chain scores censored rows via `rebuildDYF` -> `doCensNormal1`
(`src/saem.cpp`), while the IMH acceptance scores them via `likInner0`'s FOCEi
censoring. If those two disagree, the composed kernel does not target the chain's
distribution -- the same class of defect as the correlated-omega case that is
already gated.

Plain `saem` **is** covered: `tests/testthat/test-focei-cens.R:62` fits
M2-censored data with `est="saem"`. The gap is fsaem specifically, and that file
is the natural place to add the case.

## Action

1. Verify whether the two likelihoods agree on a censored dataset.
2. If they agree, add the fsaem case to `test-focei-cens.R` and no gate.
3. If they do not, gate censoring in `.fsaemSupported` and say why.

Refs #845

mattfidler:
Closes #876.

## What the verification found

#876 asked whether the SAEM chain and the FOCEi inner behind `est="fsaem"`'s IMH
acceptance agree on a censored row, and to gate censoring in `.fsaemSupported`
if they do not.  They did not -- but the disagreement was a bug in **plain
SAEM**, not something specific to the fast kernel.

`doCensNormal1` (`src/censEst.h`) speaks the FOCEi inner's language: it takes and
returns a **log-likelihood**, wants the **variance**, and reads the DV on the
**transformed** scale.  `src/nlm.cpp:432` calls it exactly that way.  All six of
the SAEM chain's DYF builders handed it the per-observation **loss**, the
residual **SD** (`_scratch_g`), and the **raw** `y`, then stored the
log-likelihood it returned back as a loss.

M2 hid this: with a limit well below the prediction the tail probability is
essentially 1, so its term is ~0 and neither a wrong sign nor a wrong scale
shows.  M3/M4 had no such cover -- the censored term arrived with its sign
flipped, and the chain drove censored predictions to the wrong side of the
limit.  On the Wang2007 M3 data:

| | tvK | prop.sd | omega |
| --- | --- | --- | --- |
| focei | 0.757 | 0.280 | 0.0014 |
| saem (before) | **-23.3** | **6627** | **8.48** |
| saem (after) | 0.683 | 0.191 | 0.00026 |
| fsaem (after) | 0.684 | 0.191 | 0.00023 |

## The fix

One `applyCensLoss` helper that translates in both directions, replacing the
repeated four-line pattern at all six sites (and the dead `doCens` it sits
where).  An uncensored row round-trips through `-doCensNormal1(..., -DYF, ...)`
unchanged, so every uncensored fit is bit-identical.

**No gate is added to `.fsaemSupported`.**  With the chain fixed, the two
likelihoods agree up to an eta-free constant, so the composed kernel targets the
chain's distribution and censored data legitimately stays on the fast kernel.
That decision is commented there so it does not get "fixed" back.

## Tests

- `test-focei-cens.R`: saem M3 **and** M4 (its own branch -- a limit turns the
  tail into a difference of two) pinned against focei's `tvK`; fsaem M3 pinned to
  saem's answer, with `fsaemDiag$nStep > 0` proving the fast kernel actually fired
  rather than degrading.
- `test-fsaem-inner.R`: the inner's M3 term checked against `pnorm` in R.
  Differencing the objective across cens/no-cens data **and** across two etas
  strips every eta-free constant (the `adjLik` offset, the transform Jacobian,
  the eta prior); what is left agrees to ~4e-5 relative, which is the reference's
  own accuracy -- the uncensored branch reproduces no better.

Run clean: `focei-cens` (50), `fsaem` (144), `fsaem-inner`, `fsaem-imh`,
`fsaem-multi`, `fsaem-es`, `fsaem-iov`, `saem-loglik`, `saem-mix`, `nlm-cens`
(87), `saem-aic`, `saem-mu`, `saem-drop`, `saem-resid-lambda`, `saem-no-pop`.

## Known gap, filed separately

Censoring is now correct in the E-step only.  The residual M-step
(`src/saem.cpp:2578`, `:2898`) still counts a BQL row's recorded DV as an
observation, so `prop.sd` comes back systematically low (0.19 against focei's
0.28 above) even though `tvK` agrees to under 10%.  That is why the new cases
pin `tvK` and only bound `prop.sd`.  Fixing it means data augmentation in the
E-step -- filed as #916.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -131,6 +131,21 @@
   `covMethod="linFim"` (the default) is unaffected throughout; its `calc.COV()`
   linearization was always per-endpoint.
 
+- `est="saem"` and `est="fsaem"` now score a **censored** observation the way the
+  rest of the package does.  The simulation step handed the shared censoring
+  likelihood the per-observation loss where it takes a log-likelihood, the
+  residual standard deviation where it takes a variance, and the untransformed
+  `DV`, then stored the log-likelihood it returned back as a loss.  M2 hid this,
+  since a limit well below the prediction leaves its tail probability at
+  essentially 1, but under M3/M4 the censored term arrived with its sign flipped
+  and the chain drove censored predictions to the wrong side of the limit: a
+  one-compartment M3 fit that FOCEi puts at `ke = 0.76` came back at `ke = -23`
+  with a proportional residual standard deviation in the thousands.  Censored
+  rows now score identically in the chain and in the FOCEi inner behind
+  `est="fsaem"`'s Metropolis-Hastings acceptance, so censored data stays on the
+  fast kernel rather than needing a gate.  An uncensored fit is bit-identical
+  (#876).
+
 - `est="fsaem"`'s Metropolis-Hastings proposal is now built against the residual
   variance the SAEM chain actually uses.  The FOCEi inner behind the proposal was
   built with the fit's `addProp`, whose default is `"combined2"`, while the SAEM
--- a/R/fsaemInner.R
+++ b/R/fsaemInner.R
@@ -182,6 +182,12 @@
   # mu-ref covariates are supported: non-time-varying ones are absorbed into the
   # per-subject mprior data, time-varying ones are kept as inner regressor betas
   # refreshed from the live Plambda each iteration.
+  #
+  # M2/M3/M4 censoring is supported and deliberately NOT gated (#876): the chain
+  # and the IMH acceptance both score a censored row with doCensNormal1 on the
+  # same f and variance, so the composed kernel still targets the chain's
+  # distribution.  test-fsaem-inner.R pins the inner's censored term against an
+  # independent evaluation and test-focei-cens.R pins fsaem to saem on M3 data.
   TRUE
 }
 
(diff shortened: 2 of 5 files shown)


mattfidler:
An independent review (Antigravity/Gemini) turned up one real gap next to this change, filed separately as #918: on an `ar()` endpoint `arDYFhyp` whitens a discarded copy of `g`, so `applyCensLoss` scores a censored row against the marginal normal while the uncensored rows of the same endpoint get the AR(1) conditional one. It is pre-existing (the same marginal `ft`/`g` went in before this PR), and plain-`saem` only -- `.fsaemSupported` already degrades an `ar()` model, so the chain-vs-FOCEi agreement this PR is about does not arise there.

The review's other findings did not survive triage: `cur_cens`/`cur_limit` are filled from `getIndCens`/`getIndLimit` off the event data (`src/saem.cpp:4624`), so they cannot vary with the proposed phi and re-assigning them on a rejected proposal is a no-op.

PR #921, opened 2026-08-14
Correlation-aware covariate selection for est="vae"

mattfidler:
Makes `est="vae"` covariate selection less likely to pick a false covariate when the candidates are colinear or the individual parameters are correlated.

## The problem, precisely

The obvious diagnosis is wrong. The search is *not* failing to consider the alternative covariate: the exact branch-and-bound enumerates every feasible subset, and the `L0Learn` path re-scores its proposals and then runs add/drop/swap to convergence. Swapping weight for lean body mass is a move both engines already score.

The criterion is what fails. The selection regresses the encoder's latent posterior means, so the sample size is the number of **subjects** and the response is an **estimate that moves every iteration**. When two columns correlate at 0.98 the difference in fit between them is smaller than the movement in the thing being fitted, so the exact minimizer is near-arbitrary and changes between iterations. Optimizing harder cannot fix that.

Separately, each latent dimension searches independently with the coupling between dimensions frozen, so no per-dimension search can see that a covariate assigned to one parameter belongs on a correlated one.

## What this adds

- **Colinearity clusters** built at the covariate level, so alternate *shapes* of one covariate never cluster with each other -- those are already arbitrated by mutual exclusion. The result is always a coarsening of the existing groups.
- **Hysteresis**: within a cluster the previous iteration's selection is not displaced by a mate that fails to beat it by `covSelectHysteresis`, in units of the `log(N)` penalty rather than as a fraction of the objective (the objective grows with N; the decision scale does not).
- **Near-tie reporting** in `$vae$colinear$alternates`, so a choice the data could not really make is visible instead of silent.
- **Cross-parameter refinement** over correlated dimensions, scoring the group jointly. Gated on the model declaring correlated etas: with a diagonal omega the objective is separable and each per-dimension search is already exact, so no cross-dimension move can improve anything -- the fit says so rather than doing the work and finding nothing.
- **`vaeCovariates()`** gained `colinearCut=` and a `cluster` column, so clusters are inspectable from data with no fit.
- Independent fix: an `L0Learn` failure that returned no proposals used to degrade a dimension to a search from the intercept-only model, silently.

`vaeControl(covSelectColinear = FALSE)` restores the previous behavior exactly, and is the only supported way to do so.

## Defaults are measured, not chosen

`tools/vaeColinearPreflight.R` records the measurements the defaults come from. No covariate design in the package binds at 0.8 or above (warfarin is highest at 0.775), and an ordinary one-compartment model shows 0.824 between its clearance and volume dimensions -- so 0.8 would group a well-identified fit. Both cuts default to 0.9.

## Blast radius

Measured as nil on the current suite: every covariate-bearing fixture is either single-group or far below the cut, and no existing vae test both declares a correlated omega and carries covariates. `test-vae-cov-groups.R`, `test-vae-miqp.R` and `test-vae-l0learn.R` pass unmodified, which is the zero-drift evidence.

## Verification

927 assertions across eleven essential covariate files, plus 61 in the new weekly `test-vae-colinear-fit.R` (batch 8). Every mechanism has both a "it ran" and an "it acted" counter so silence is distinguishable from never having been reached. The group-parallel refinement is bit-identical at a fixed thread count, with counters and selected covariates matching across thread counts.

## Review

Five rounds of independent review (Gemini via Antigravity). Rounds 3-5 found no correctness, memory-safety or concurrency issues. Fixed along the way:

- `vaeJointScore` stripped the intercept column with `cols(1, n-1)`, out of bounds for an intercept-only support -- it threw into the OpenMP catch-all and silently disabled a whole refinement group.
- The same function returned `+inf` when the block system had no free parameters, which is legitimate (fully clamped group, empty supports) and blocked dropping the last covariate.
- Near-tie rows were left stale for dimensions a joint move rewrote.
- Missing coverage for `covSelectPhiMaxDim`, the held-intercept path, all three `covSelectPhiCor` sources, and the hysteresis hold.

**One known gap**, documented next to the test: the specific claim "the correlation crossed below join and the pair stayed" is not directly asserted. The available fixture shows it as a single iteration, and this fit varies at 1e-14 across thread counts, so a test on that margin would be flaky rather than informative. `$vae$colinear$phiPairOn` is surfaced so a sharper fixture can assert it later.

Not run: the full package suite. Only the vae/covariate subset plus the two colinear files.


mattfidler:
didn't work

Issue #958, opened 2026-08-16
est="stan" tier 2: combined eta+theta sensitivity model (one ODE integration per evaluation instead of two)


mattfidler:
## Problem

The FOCEi C API's tier-2 path (nlmixr2stan's `est="stan"`) computes each log-density evaluation's gradients with **two separate ODE integrations per subject**:

1. `nlmixr2FoceiCondBatch` solves the **eta-sensitivity** model (rxInner) for the conditional value + `d/d(eta)`;
2. `nlmixr2FoceiCondThetaGrad` then solves the **theta-sensitivity** model (the impmap forward-sensitivity build) for `d/d(theta)` at the same point.

Same time grid, same events, same structural ODE — integrated twice. Measured on theo_sd (12 subjects, 2-state ODE, dense `dop853`, 4 threads): **0.58 ms (eta batch) + 0.78 ms (theta batch) = 1.36 ms** per full-population gradient at the C level.

## Why it matters

A NUTS chain is essentially a pure stream of these evaluations, so the duplicated stepping is the dominant per-evaluation overhead. In a matched benchmark against the same model written natively in Stan (`ode_rk45_tol`, same priors/inits/tolerances/sampler settings, same core budget), the per-chain wall ratio (~1.43) matches the per-evaluation cost ratio (~1.41) almost exactly — evaluation counts and geometry are equivalent, so **per-evaluation cost is the whole remaining gap**. Native Stan pays ~1.9 ms because it solves ONE coupled system carrying all sensitivities.

A combined model — one rxode2 build carrying **both** the `d/d(eta)` and `d/d(theta)` forward-sensitivity columns, solved once per subject per evaluation, with both gradient sets read from that single solve — should land the combined gradient around 0.8–1.0 ms, i.e. clearly below the native coupled solve. Projected effect with chain-level parallelism also in place: the linked ODE tier goes from ~0.70 ESS/s to roughly 2.5–3.5 ESS/s vs native's 1.66 on this model.

## Sketch

- A `combinedSens = TRUE`-style build (only `est="stan"` needs it; focei/impmap consume the two models separately and keep their current builds): the eta-sensitivity equations plus the estimated-theta forward sensitivities in one model, sharing the structural states.
- One batch entry (or an extended `nlmixr2FoceiCondBatch` signature) that fills `value`, `gradEta`, and `gradTheta` from the single solve.
[...]

PR #959, opened 2026-08-17
Combined eta+theta sensitivity model + fused single-solve tier-2 entry (#958)

mattfidler:
Implements #958. **Stacked on #957** (base branch `fix/cond-batch-nan-value`) — merge that first, then rebase/retarget this to main.

## What

`est="stan"`'s tier-2 gradient previously cost **three ODE integrations per subject per evaluation** (inner solve for value + d/d(eta); condThetaGrad's likInner0 pre-pass; the separate theta-sensitivity-model solve). With `foceiLikLoad(combSens=TRUE)` it costs **one**:

- **R**: the inner generator derives the eta+theta union in a single `.rxJacobian`/`.rxSens` pair and appends the theta columns (impmap chain rule, identical naming) **after** the FOCEi eta block — `likInner0`'s arithmetic `predOffset` reads stay byte-identical. State sensitivities only for etas + structural non-mu thetas; sigma thetas get columns only. Opt-in (`combSens` control), keyed into the model digest; the separate theta-sensitivity model is not built under it.
- **C**: theta offsets resolve by name against `odeSlotInner` (both setup paths); `impThetaSensCollect` gains a slot switch plus `reuseSolve` (skips its re-solve when the caller just solved the combined model at the same point); new fused entry `nlmixr2FoceiCondBatchThetaGrad` (table entry 10, additive/defensive install) fills value + gradEta + dTheta from one solve under **one** event-sensitivity shape (the combined model owns both jump-sensitivity sets). Dims flag `0x80` advertises the capability.

## Measurements

theo_sd, dense `dop853`, 4 threads: **0.62 ms** per full-population gradient vs 1.26 ms two-model (**2.03×**) — and below the native-Stan coupled solve (1.98 ms). Fused outputs are **bitwise identical** to the separate combined-path calls; vs the two-model path agreement is at solver tolerance (~1e-6), inherent since the theta states ride the same adaptive integration.

## Drive-by fix

`paste0()` recycles a zero-length index to `""` (R ≥ 4.0), so a sigma-only theta set fabricated a malformed `THETA__` sensitivity variable — guarded in the new path **and** in the pre-existing impmap builder, where the same latent bug existed.

## Tests

focei/nlm/imp suites: 8,136 passing, 0 failed. focei-ptrs golden updated to the 10-entry table; new tests cover the combined layout, two-path parity, the fused entry's bitwise contract, and the sigma-only build.

Downstream: nlmixr2/nlmixr2stan#12's follow-up will consume the fused entry (projected to take the linked ODE tier past native ESS/s together with the forked chains already landed there).

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -402,6 +402,10 @@ foceiLikCondGrad_ <- function(etaMat, cores) {
     .Call(`_nlmixr2est_foceiLikCondGrad_`, etaMat, cores)
 }
 
+foceiLikCondBatchThetaGrad_ <- function(etaMat, cores) {
+    .Call(`_nlmixr2est_foceiLikCondBatchThetaGrad_`, etaMat, cores)
+}
+
 foceiLikCondThetaGrad_ <- function(etaMat, cores) {
     .Call(`_nlmixr2est_foceiLikCondThetaGrad_`, etaMat, cores)
 }
--- a/R/focei.R
+++ b/R/focei.R
@@ -673,7 +673,7 @@ attr(rxUiGet.loadPrune, "rstudio") <- emptyenv()
 #' @author Matthew L. Fidler
 #' @export
 #' @keywords internal
-.sensEtaOrTheta <- function(s, theta=FALSE) {
+.sensEtaOrTheta <- function(s, theta=FALSE, extraThetaVars=NULL) {
   .etaVars <- NULL
   if (theta && exists("..maxTheta", s)) {
     .etaVars <- paste0("THETA_", seq(1, s$..maxTheta), "_")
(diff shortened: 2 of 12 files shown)


mattfidler:
Added one more commit: the per-evaluation Omega fetch went through rxode2's precomputed R closure (~1.2 ms/call — more than the fused ODE batch itself). It is now a C-side chol-factor rebuild: the tail→factor map is probe-discovered at setup, verified against the R path to 1e-10 (automatic permanent fallback on any mismatch), 0.015 ms/call even with the tail moving. Development surfaced an implicit invariant — updateTheta used to leave the env theta current, and innerOpt/impGetOmega/gradDirectOmega silently depended on it — now explicit via per-site env sync (the imp EM tests caught it; 8,534 tests green across focei/nlm/imp/saem/agq). Net effect downstream: the est="stan" ODE benchmark goes to **80.8 s / ESS 326 / 4.04 ESS/s vs native Stan's 142.7 s / 236 / 1.66 — 2.4x native**, with bit-identical draws at every optimization step. Escape hatches: NLMIXR2_NO_FAST_OMEGA / NLMIXR2_NO_OMEGA_MEMO.

Issue #1001, opened 2026-08-22
nlm-family: M2/M3/M4 censoring with ar() uses the marginal, not conditional, variance/mean


mattfidler:
## Context

Follow-up from #976 (fixed in the associated PR). #976 fixed nlm-family's
`rx_r_ ~ 0` bug, which made every M2/M3/M4-censored normal endpoint
mis-scored regardless of `ar()`. That fix restores a real variance for
`rx_r_` by referencing the `rx_rll_` variable rxode2 already computes for
the llik path (`.nlmFixCensRLine()`, `R/nlm.R`).

For a plain (non-`ar()`) normal endpoint this is exactly correct: `rx_rll_`
*is* the endpoint's real residual sd. For an `ar()` endpoint, though,
`rx_rll_` is deliberately the **marginal** (unconditional) sd --
`.rxArEstLlikLines()` (rxode2's `err-foceiBase.R`) sets it once from
`.rxGetVarianceForErrorType()` before AR-whitening, then builds the actual
per-observation log-likelihood with a *different*, conditional sd expression
(`rx_rll_ * sqrt(1 - phi^2)`, `phi` being the AR(1) decay to the previous
observation) that is never assigned to its own variable.

`src/nlm.cpp`'s censoring correction (`doCensNormal1()`) also reads
`rx_pred_f_` for the mean, which is nlm's raw/marginal structural
prediction, never AR-whitened either (documented at `src/nlm.cpp:60-67`).

So for an `ar()` + censored normal endpoint, nlm-family now computes the
censoring correction using the **marginal** mean and **marginal** variance
together (self-consistent with each other, so no blowup/boundary-pinning),
rather than the conditional mean/variance implied by the AR(1) process.
This is a real accuracy gap, not the "totally invalid" state #976 fixed --
confirmed empirically: fitting `theo_sd` with `add(add.sd) + ar(ar1.cor)`
and M3 censoring via `bobyqa` recovers `tka`/`tcl`/`tv`/`add.sd` reasonably
close to `focei`'s reference fit, but `ar1.cor` itself is noticeably off
(0.15 vs `focei`'s 0.007 in one run) -- exactly the parameter that the
marginal-vs-conditional substitution most directly affects.

## Why not fixed as part of #976

Fixing this exactly requires nlm-family's censoring path to receive the
AR(1)-conditional mean and variance the way `src/inner.cpp`'s `likInner0`
[...]

Issue #979, opened 2026-08-19
M2/M3/M4 censoring is silently ignored for t()/cauchy() endpoints (all estimation methods)


mattfidler:
## Context

Found by an independent Antigravity/Gemini review of #976's fix (PR pending
from branch `nlm-976`), while checking whether restoring a real `rx_r_` for
nlm-family's llik-forced normal path also helped `t`/`cauchy` endpoints (it
does restore `rx_r_` for them too -- `.nlmFixCensRLine` in `R/nlm.R` is not
distribution-specific). It doesn't help, because the C++ side never reads it
for those distributions -- and that part is pre-existing, unrelated to #976,
and not specific to nlm.

## Root cause

`src/nlm.cpp`'s M2/M3/M4 censoring correction guards its call to
`doCensNormal1()` with:

```cpp
if (dist == rxDistributionDnorm || dist == rxDistributionNorm) { ... }
```

at `nlmSolveFid` (lines ~466, ~493) and `nlmSolveGradId` (line ~571). A `t()`
or `cauchy()` endpoint compiles to `rxDistributionT`/`rxDistributionCauchy`,
so this guard is never entered for them -- censoring is silently skipped and
a censored observation is scored with the ordinary (uncensored) probability
density instead of the M2/M3/M4 CDF-based correction.

The same gap exists in `src/inner.cpp` (FOCEI/FOCE/Laplace/AGQ and everything
that shares `likInner0`): `rxDistributionT`/`rxDistributionCauchy` do not
appear anywhere in that file either, so there is no censoring correction for
a `t`/`cauchy` endpoint there either. This is not specific to nlm-family --
it looks like M2/M3/M4 censoring was only ever implemented for the normal
distribution, ecosystem-wide.

## Reproduction

```r
one.cmt <- function() {
  ini({ tka <- 0.45; tcl <- log(c(0, 2.7, 100)); tv <- 3.45; add.sd <- 0.7; nu <- fix(5) })
  model({
    ka <- exp(tka); cl <- exp(tcl); v <- exp(tv)
    linCmt() ~ add(add.sd) + t(nu)
  })
}
# M3-censor some observations, fit with est="nlm" (or est="focei"):
# the fit runs without error/warning, censInformation reports "M3 censoring",
# but the objective/parameters are identical to what add(add.sd)+t(nu) would
# give with those rows simply left at their (fabricated) LOQ value uncensored.
```

## Suggested fix directions (not investigated in depth)

[...]

PR #1002, opened 2026-08-22
fix(nlm): recover rx_r_ from rx_rll_ for censored normal endpoints (#976)

### LINKED ISSUE #976: nlm-family: M2/M3/M4 censoring silently broken for every normal endpoint (rx_r_ hardcoded to 0), not specific to ar()
## Context

Filed while verifying whether the marginal-vs-conditional AR(1) censoring bug
fixed in #918 (SAEM) also affects the other estimation methods. It does not
affect FOCEI/FOCE/Laplace/AGQ/FO/FOI (or imp/npag/npb/vae/advi, which share
the same C++ conditional-likelihood engine) -- see "Methods verified safe"
below. But checking nlm-family surfaced a **different, more fundamental**
defect: `est="nlm"` (and every nlm-family sub-method: `bobyqa`, `newuoa`,
`uobyqa`, `n1qn1`, `lbfgsb3c`, `optim`, `nlminb`) silently mis-scores **any**
M2/M3/M4-censored normal endpoint, whether or not `ar()` is present. Since
it's already broken without AR, there isn't a separate "AR(1) conditional
vs marginal" bug to report for nlm -- there's no valid scoring happening at
all once censoring is present.

## Root cause

`rxUiGet.nlmModel0` (`R/nlm.R:409-410`) unconditionally sets
`nlmixr2global$rxPredLlik <- TRUE` and converts every `norm`-distribution
endpoint to `dnorm` before building the model
(`.predDf[.predDf$distribution == "norm", "distribution"] <- "dnorm"`,
`R/nlm.R:413`). This routes **every** nlm normal endpoint through
`.handleSingleErrTypeNormOrTFoceiBase()`'s log-likelihood branch
(`rxode2/R/err-foceiBase.R:556-605`), which:

- reassigns `rx_pred_` to the scalar log-density `llikNorm(dv, mean, sd)`
  (or, for an `ar()` endpoint, `.rxArEstLlikLines()`,
  `rxode2/R/err-foceiBase.R:410-450`, whose `mean`/`sd` ARE the correct
  AR(1)-conditional values -- that part is fine), and
- **always sets `rx_r_ ~ 0`** (`err-foceiBase.R:601`), because a full
  log-likelihood value has no separate variance to report.

`src/nlm.cpp`'s censoring code was written for the *other* branch, where
`rx_pred_f_`/`rx_r_` hold a real structural prediction/variance
(`src/nlm.cpp:60-67` documents this layout assumption). At
`src/nlm.cpp:463-478` (and the gradient-contribution mirror at
`:568-596`):

```cpp
if (nlmOp.hasFR && (hasRxCens(rx) || hasRxLimit(rx))) {
  ...
  if (dist == rxDistributionDnorm || dist == rxDistributionNorm) {
    ...
    double f = lhs[po + 1]; // rx_pred_f_  <- marginal/structural pred, never AR-whitened
    double r = lhs[po + 2]; // rx_r_       <- ALWAYS 0 for nlm's llik-forced normal endpoints
    double ll = doCensNormal1((double)censi, dvi, limiti, -val, f, r, 0);
    val = -ll;
  }
}
```

`r=0` feeds a zero variance into `doCensNormal1()` for every censored
observation, corrupting the objective. This is unconditional on `ar()` --
it fires for the plainest `add()`/`prop()` model too.

## Reproduction (no ar() needed)

Using the exact model/data already in `tests/testthat/test-nlm-cens.R`
(`theo_sd`, `linCmt() ~ add(add.sd)`, LLOQ=2 M3 censoring):

```r
one.cmt <- function() {
  ini({ tka <- 0.45; tcl <- log(c(0, 2.7, 100)); tv <- 3.45; add.sd <- 0.7 })
  model({ ka <- exp(tka); cl <- exp(tcl); v <- exp(tv); linCmt() ~ add(add.sd) })
}
.dat <- nlmixr2data::theo_sd
.LLOQ <- 2.0
.datM3 <- .dat
.datM3$CENS <- ifelse(.datM3$DV < .LLOQ & .datM3$EVID == 0, 1L, 0L)
.datM3$DV[.datM3$CENS == 1] <- .LLOQ

fit_nlm <- nlmixr(one.cmt, .datM3, est = "nlm", control = nlmControl(print = 0))
```

| method (censored) | objf | tka | tcl (CL) | tv (V) | add.sd |
|---|---|---|---|---|---|
| **nlm** | 445.2 | 36.2 (pinned near box) | 4.61 (pinned at box upper bound) | -5.25 | 6.40 |
| **bobyqa** | 7219.0 | 0.62 | 0.89 | 3.54 | 1.43 |
| **nlminb** | 7422.9 | 0.46 | 0.98 | 3.46 | 0.70 |
| **focei** (reference, same data) | 189.2 | 0.45 | 1.03 | 3.45 | 0.76 |

`nlm` pins two parameters at their box bounds and produces a nonsensical
fit; `bobyqa`/`nlminb` land on a wrong objective scale entirely (`~7200`
vs FOCEI's `189`). Adding `ar()` to a mixed-effects version of this
reproduces the same kind of blowup (`add.sd` inflates ~70x, `ar1.cor` pins
near its `[0,1)` boundary) -- consistent with the same `r=0` mechanism,
not a separate AR-specific bug.

## Why existing tests didn't catch this

`tests/testthat/test-nlm-cens.R` only asserts `expect_s3_class(...)`, that
`censInformation` reports the right M2/M3/M4 text, and that the censored
`objf` differs from the uncensored one -- never that the resulting
parameter estimates are anywhere close to correct. A `r=0`-corrupted
objective still "runs", still reports the right censoring text, and still
produces a different (wildly different) `objf`, so every existing
assertion passes.

## Methods verified safe (no bug -- for reference / to close the loop on #918's ask)

FOCEI/FOCE/Laplace/AGQ/FO/FOI/EBE (`posthoc`) all build their inner model
via `rxUiGet.focei`/`rxUiGet.foce`/`rxUiGet.ebe`, which set
`nlmixr2global$rxArNorm <- TRUE` for `ar()` endpoints -- this keeps the
endpoint on the plain `norm` (mean/variance) path via
`.rxArEstNormLines()` (`rxode2/R/err-foceiBase.R:466-493`), which
reassigns `rx_pred_`/`rx_r_` **themselves** to the AR(1)-conditional
mean/variance. `src/inner.cpp`'s shared `likInner0` (used by all of the
above, plus `imp`/`impmap`, `npag`, `npb`, which route through the same
`ui$foceiOptEnv`) reads `f`/`r` directly from `rx_pred_`/`rx_r_`
(`src/inner.cpp:2623`, `:2668`) -- there is only one copy, so there's no
marginal-vs-conditional duality to get wrong. Verified empirically: FOCEI
on an `ar()` + M3-censored dataset recovers `tka`/`tcl`/`tv`/`add.sd`/
`ar1.cor` all close to their simulated truth (no boundary-pinning, no
blowup).

## Suggested fix directions (not investigated in depth)

- Have `.rxArEstLlikLines()`'s general llik path (and the plain non-AR
  `dnorm` llik path nlm always uses) build the M2/M3/M4 censoring
  correction **into** the emitted log-likelihood expression itself, the
  way a general `ll()` endpoint would need to for censoring to make sense
  at all in a pure-log-likelihood model, or
- Stop forcing every nlm normal endpoint through the llik/`dnorm` path --
  reserve `rxPredLlik=TRUE` for `t`/`cauchy`/actual `ll()` endpoints, and
  let a plain `norm` endpoint keep `rx_r_` as a real variance the way
  `src/nlm.cpp`'s censoring code already expects.

Refs #918.

mattfidler:
## Summary

Fixes #976: every NLM-family method (`nlm`, `bobyqa`, `newuoa`, `uobyqa`,
`n1qn1`, `lbfgsb3c`, `optim`, `nlminb`) silently mis-scored any M2/M3/M4-censored
normal endpoint, independent of `ar()`.

- `est="nlm"` forces every normal endpoint through the log-likelihood
  (`dnorm`) path so a single scalar objective can be emitted
  (`rxUiGet.nlmModel0`). That path always sets `rx_r_ ~ 0` (a full
  log-density has no separate variance to report), but `src/nlm.cpp`'s
  M2/M3/M4 censoring correction (`doCensNormal1()`) reads `rx_r_` as a
  real variance -- a hardcoded 0 corrupted every censored normal
  endpoint's contribution.
- Fixed in `R/nlm.R` only (no C++/rxode2 changes): a new helper
  `.nlmFixCensRLine()` walks each endpoint's quoted model lines and
  replaces `rx_r_ ~ 0` with `rx_r_ ~ rx_rll_^2` -- `rx_rll_` is the
  standard deviation rxode2 already computes and feeds into
  `llikNorm()`/`llikT()`/`llikCauchy()`, emitted immediately beforehand
  in the same branch.
- The fix references the `rx_rll_` **variable**, not its defining
  expression -- an antigravity review round caught that re-inlining the
  expression corrupts `propT()`/`powT()` (transformed prop/pow) models,
  since that expression contains the symbol `rx_pred_`, which is
  overwritten with the scalar log-likelihood between the `rx_rll_` and
  `rx_r_` lines.

**Verified:** `bobyqa`/`newuoa`/`uobyqa`/`nlminb` (derivative-free/FD-gradient
nlm-family members) recover parameters matching `focei` to ~1-2 decimals on
an M3-censored `theo_sd`/`linCmt()` reproduction, for both a plain `add()`
endpoint and a `propT()` endpoint.

**Known limitations, filed as follow-ups (not fixed here, same pattern as
this issue's own #979 carve-out):**
- #979 -- M2/M3/M4 censoring is silently ignored for `t()`/`cauchy()`
  endpoints everywhere (pre-existing, unrelated to this fix).
- #1001 -- combining `ar()` with censoring now gets a sane,
  self-consistent **marginal** approximation (a real fix from the totally
  invalid `r=0` state), rather than the exact AR(1)-**conditional**
  likelihood FOCEI/FOCE/Laplace/AGQ compute. Getting the exact conditional
  correction needs a larger, separate change to how nlm's censoring path
  reads the AR-whitened mean/variance.
- The analytic-gradient nlm-family members (`nlm` itself via `stats::nlm`,
  `n1qn1`, `lbfgsb3c`, `optim`) have a separate, pre-existing convergence
  stall on the bounded-`tcl` test model, present identically with or
  without censoring -- out of scope, so the new tests only assert
  parameter accuracy for the derivative-free/`nlminb` subset.

Reviewed with two rounds of an independent (Gemini-based) antigravity code
review; round 2 reported no remaining issues.

## Test plan

- [x] `tests/testthat/test-nlm-cens.R`: existing regression tests pass
- [x] New test: nlm-family M3-censored parameter estimates match `focei`
      for a plain `add()` endpoint (#976)
- [x] New test: nlm-family M3-censored parameter estimates match `focei`
      for a `propT()` endpoint (catches the antigravity-review-found
      AST-substitution bug)
- [x] Manual repro comparing pre-fix vs post-fix `bobyqa` output confirms
      the fix materially improves accuracy for both `add()` and `propT()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -113,7 +113,10 @@
   the correction for every censored observation. `rx_rll_` (the standard
   deviation actually used to build the log-density) is emitted immediately
   beforehand in the same branch, so it is now squared back into `rx_r_`
-  instead of being discarded.
+  instead of being discarded. For an `ar()` endpoint this restores a sane,
+  self-consistent (marginal) censoring correction rather than a corrupted
+  one; getting the exact AR(1)-conditional correction is a larger, separate
+  change tracked in #1001.
 
 - `est="saem"` with a `pow()` residual error model (`rmPow`/`rmAddPow`/
   `rmPowLam`/`rmAddPowLam`) never applied the estimated power exponent in the
--- a/R/nlm.R
+++ b/R/nlm.R
@@ -415,7 +415,12 @@ getValidNlmixrCtl.nlm <- function(control) {
 #' independent of `ar()` (#976).  `rx_rll_` (the standard deviation actually
 #' fed into `llikNorm()`/`llikT()`/`llikCauchy()`) is emitted immediately
 #' before `rx_r_` in the same branch, so square it back into `rx_r_` instead
-#' of leaving the hardcoded 0.
+#' of leaving the hardcoded 0.  Reference the `rx_rll_` VARIABLE, not its
+#' defining expression: for a transformed prop()/pow() error model (e.g.
+#' `propT()`), that expression contains the symbol `rx_pred_`, which is
+#' overwritten with the scalar log-likelihood between the `rx_rll_` and
+#' `rx_r_` lines -- re-inlining it would silently read the log-likelihood
+#' value back in as the mean.
 #'
 #' @param lines A single endpoint's list of quoted model lines, as returned
 #'   by `rxGetDistributionFoceiLines()` for one `predDf` row.
@@ -425,20 +430,16 @@ getValidNlmixrCtl.nlm <- function(control) {
 #' @noRd
 .nlmFixCensRLine <- function(lines) {
   if (!is.list(lines)) return(lines)
-  .rll <- NULL
-  for (.l in lines) {
-    if (is.call(.l) && identical(.l[[1]], quote(`~`)) &&
-        identical(.l[[2]], quote(rx_rll_))) {
-      .rll <- .l[[3]]
-      break
-    }
-  }
-  if (is.null(.rll)) return(lines)
+  .hasRll <- any(vapply(lines, function(.l) {
+    is.call(.l) && identical(.l[[1]], quote(`~`)) &&
+      identical(.l[[2]], quote(rx_rll_))
+  }, logical(1)))
+  if (!.hasRll) return(lines)
   lapply(lines, function(.l) {
     if (is.call(.l) && identical(.l[[1]], quote(`~`)) &&
         identical(.l[[2]], quote(rx_r_)) &&
         identical(.l[[3]], 0)) {
-      bquote(rx_r_ ~ .(.rll)^2)
+      quote(rx_r_ ~ rx_rll_^2)
     } else {
       .l
     }
--- a/tests/testthat/test-nlm-cens.R
+++ b/tests/testthat/test-nlm-cens.R
@@ -235,4 +235,45 @@ nmTest({
     }
   })
 
+  test_that("nlm-family M3-censored propT() parameter estimates match focei (#976)", {
+    # #976 follow-up (antigravity review): .nlmFixCensRLine originally
+    # rebuilt rx_r_ by re-inlining rx_rll_'s defining EXPRESSION (e.g.
+    # sqrt((b*rx_pred_)^2) for propT()/powT(), whose F is the TRANSFORMED
+    # prediction, i.e. the symbol rx_pred_ itself -- see
+    # .rxGetVarianceForErrorPropOrPowF()).  That expression is placed at the
+    # rx_r_ ~ 0 line's position, which comes AFTER rx_pred_ has already been
+    # overwritten with the scalar log-likelihood (rx_pred_ <- llikNorm(dv,
+    # rx_pred_, rx_rll_)).  Re-evaluating the AST there silently fed the
+    # log-likelihood value back in as the propT() mean, corrupting rx_r_ for
+    # any transformed prop()/pow() error model.  The fix instead references
+    # the already-computed rx_rll_ VARIABLE (rx_r_ ~ rx_rll_^2), which still
+    # holds its original value at that point.  A plain add()/prop() (F =
+    # rx_pred_f_, a column untouched by the llik overwrite) would not have
+    # caught this -- propT() is required to exercise the bug.
+    one.cmt.propT <- function() {
+      ini({
+        tka <- 0.45
+        tcl <- log(c(0, 2.7, 100))
+        tv <- 3.45
+        prop.sd <- c(0, 0.3)
+      })
+      model({
+        ka <- exp(tka)
+        cl <- exp(tcl)
+        v <- exp(tv)
+        linCmt() ~ propT(prop.sd)
+      })
+    }
+    f.focei <- .nlmixr(one.cmt.propT, .datM3, est = "focei", control = foceiControl(print = 0))
+    fit <- .nlmixr(one.cmt.propT, .datM3, est = "bobyqa", control = nlmControl(print = 0))
+    expect_equal(as.numeric(fit$theta[["tka"]]), as.numeric(f.focei$theta[["tka"]]),
+                 tolerance = 0.15)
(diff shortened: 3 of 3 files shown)


mattfidler:
Closing: while this PR was in flight, #976 was independently fixed and merged as PR #983 (same diagnosis, same \`rx_rll_\` -> \`rx_r_\` fix), then further refactored by PR #990 (#979's t()/cauchy() censoring work), which moved the logic from \`R/nlm.R\`'s \`.nlmFixCensRLine()\` into a shared \`.fixCensRNuLine()\` in \`R/focei.R\`. This branch (based on a now-stale point in history) duplicates that already-merged work and conflicts with current \`main\`.

The one genuinely new finding from this PR's antigravity review -- that re-inlining \`rx_rll_\`'s defining expression (rather than referencing the variable) silently corrupts \`propT()\`/\`powT()\` censored fits -- is still a live bug in current \`main\`'s \`.fixCensRNuLine()\`. Re-opening that fix, correctly rebased on current \`main\`, in a follow-up PR. The AR(1) conditional-vs-marginal gap this PR's review also surfaced is tracked in #1001 (still applicable to the current code).

PR #1011, opened 2026-08-24
perf(saem): drop duplicate THETA/ETA alias assignments in saemModelPred

billdenney:
The SAEM predOnly model assembled in rxUiGet.saemModelPred concatenated two alias-emitting blocks: the mu-reference replacement block (paste(names(.replaceLst), ...)) and the THETA/ETA alias block (.uiGetThetaEta()). For the common non-combined case these emit the *exact same* `lhs <- rhs` assignments (e.g. `tke <- THETA[1]`, `eta_DoT_ke <- ETA[1]`), which codegen then duplicated into both dydt and calc_lhs as trailing dead-store assignments.

Filter the THETA/ETA alias lines, dropping only those that are exact duplicates (same lhs AND same rhs, compared after parse+deparse normalization) of the replacement block, keeping the first occurrence. Lines whose rhs differs (mu-referenced combined
`lhs <- THETA[k] + ETA[j]` vs the split `lhs <- THETA[k]`) are not exact duplicates and are preserved unchanged, so the generated model is numerically identical; only redundant dead-store aliases are removed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,12 @@
 
 ## Internal
 
+- The SAEM `predOnly` model (used for residuals, tables and the covariance
+  step) no longer emits a THETA/ETA alias assignment that exactly duplicates
+  one the mu-reference replacement block already emitted.  These were trailing
+  dead stores that rxode2 repeated in both `dydt` and `calc_lhs`; the emitted
+  model, its solve column layout, and every value it produces are unchanged.
+
 - A covariate whose value is carried on the model in `rxode2::rxForcedPars()` is
   no longer required to be a column of the data.  Such a covariate is supplied
   by the model itself, so demanding it from the data rejected a well-specified
--- a/R/saemRxUiGetModel.R
+++ b/R/saemRxUiGetModel.R
@@ -588,11 +588,12 @@ attr(rxUiGet.interpLinesStr, "rstudio") <- ""
 #' @param names Character vector of variable names to look for
 #' @return Logical, one per `names`: TRUE when the text assigns (`=`, `<-` or
 #'   `~`) to that plain name.  A non-name target such as `d/dt(x)` or `f(x)`
-#'   never counts.
+#'   never counts, and `=(?!=)` keeps a comparison (`x == 1`) from reading as
+#'   an assignment to `x`.
 #' @author Matthew L. Fidler
 #' @noRd
 .lhsAssignedIn <- function(modelText, names) {
-  .lhs <- trimws(sub("[ \t]*(<-|=|~).*$", "", modelText))
+  .lhs <- trimws(sub("[ \t]*(<-|~|=(?!=)).*$", "", modelText, perl=TRUE))
   names %in% .lhs[grepl("^[A-Za-z._][A-Za-z0-9._]*$", .lhs)]
 }
 
--- a/R/saemRxUiGetModel.R
+++ b/R/saemRxUiGetModel.R
@@ -653,15 +653,36 @@ rxUiGet.saemModelPred <- function(x, ...) {
     .ret2
   ), collapse = "\n")
   .interp <- rxUiGet.interpLinesStr(x, ...)
+  ## The mu-reference replacement block (paste(names(.replaceLst), ...))
+  ## and the THETA/ETA alias block (.uiGetThetaEta()) can emit the *exact*
+  ## same `lhs <- rhs` assignment, which is then duplicated into both dydt
+  ## and calc_lhs by codegen.  Drop the THETA/ETA alias lines that are exact
+  ## duplicates (same lhs AND same rhs) of the replacement block, keeping the
+  ## first occurrence.  Lines whose rhs differs (e.g. a mu-referenced combined
+  ## `lhs <- THETA[k] + ETA[j]` in the replacement block vs the split
+  ## `lhs <- THETA[k]` alias) are NOT exact duplicates and are preserved, so
+  ## the emitted model is numerically identical.
+  .replaceLines <- paste(names(.replaceLst), "<-", .replaceLst)
+  .thetaEtaLines <- vapply(.uiGetThetaEta(x[[1]]), deparse1, character(1), USE.NAMES=FALSE)
+  .normAssign <- function(.l) {
+    vapply(.l, function(.s) {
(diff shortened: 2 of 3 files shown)


mattfidler:
Hm.  This would appear to break things like detecting which items are mu are mu referenced because they were time-varying.  

It isn't really clear from the code what exactly is happening either.  Nor are there any tests to make sure this works correctly...

mattfidler:
Nevermind, I see what it is doing, but the testing requirement still should be done.  It should be a proper function that can be tested.

mattfidler:
## Review

Reviewed, and pushed four follow-up commits. The de-duplication is the right
idea, but as written it was **not** behavior-preserving in one case, and
nothing tested it. Both are now addressed.

### The bug: a duplicate alias is not always a dead store

A model block may legally assign to an `ini()` parameter's own name:

```r
model({
  ka <- exp(tka + eta.ka)
  tv <- tv + 1          # legal; nlmixr2 accepts it
  cl <- exp(tcl)
  v  <- exp(tv)
  linCmt() ~ add(add.sd)
})
```

Both blocks then emit an identical `tv=THETA[3]`, so the filter treated the
trailing one as a duplicate and dropped it. But the body writes to `tv`
*between* the two, which makes the trailing alias a **live store** -- it is
what puts `THETA[3]` back into the reported `tv` column. In the generated
model:

```
 6: tv=THETA[3];          <- replacement block
...
17: tv=rx_expr__0;        <- the body overwrites it
...
24: tv=THETA[3];          <- the alias that was being dropped
```

Solving both forms with the same parameters, the reported `tv` column moves
from **3.45 (THETA[3]) to 4.45 (the body's value)**. Silent, and no test
would have caught it.

Fixed in 2e42517 by gating the drop on a new `.lhsAssignedIn()`: an alias goes
only when it duplicates a replacement line **and** the model body never
assigns its lhs. 65a9d5d then hardens that gate's regex -- it stopped at the
first `=`, so `tv == 1` read as an assignment to `tv` (a false positive, so it
only ever kept a dead alias, never dropped a live one; fixed anyway). `=(?!=)`
now distinguishes assignment from `==`/`<=`/`>=`/`!=`.

### Verification

Built the `predOnly` model on this branch and on `main` for 12 model shapes
(mu-referenced ODE, thetas carrying no eta, mu-referenced covariate, non-mu
eta, general log-likelihood, `linCmt()`, IOV, multi-endpoint, `boxCox()`, a
`fix()`ed theta, a string covariate, no population parameter). For every one:

- the branch model is **exactly** `main`'s minus duplicated lines
  (`identical(branchLines, mainLines[!duplicated(mainLines)])`),
- `$params`, `$lhs` and `$state` are identical, so the solve column layout
  does not shift -- which matters because SAEM/FOCEi C code reads some lhs
  columns by index,
- solving both model texts with the same parameters and events gives identical
  output, column for column.

A full seeded `est="saem"` fit of `theo_sd` is bit-identical between the two:
`objf`, `fixef()`, `omega`, the etas, and every residual column
(`PRED`/`IPRED`/`IRES`/`IWRES`/`CPRED`/`CWRES`) compare `identical()`.

SAEM regression sweep across 24 test files: **433 assertions pass, 0 errors**.
Two failures in `test-saem-cov-sa.R` are pre-existing and unrelated -- with
this PR's filter enabled and with it disabled the failure output is
byte-identical. Filed separately as #1022.

### Tests added

`tests/testthat/test-saem-model-pred-dedup.R`, 25 assertions, pinning the
mechanism rather than only the outcome:

- a non-mu-referenced parameter (`tv`, `pow`, `add.sd`) is aliased exactly once;
- a mu-referenced parameter keeps **both** `tka=THETA[1]+ETA[1]` and the
  trailing split `tka=THETA[1]`, in that order (not exact duplicates);
- **the dropped aliases really were dead stores** -- the test rebuilds the
  pre-de-duplication model by re-appending the full `.uiGetThetaEta()` block
  after the body, then asserts both models have the same `params`/`lhs`/`state`
  and `rxSolve()` to identical output;
- **the reassigning model above keeps all three `tv` assignments** and still
  solves `tv` to `THETA[3]`, while `tcl`/`add.sd` de-duplicate as before;
- `.normAssign()` and `.lhsAssignedIn()` unit tests, including an empty rhs
  (the placeholder a non-mu eta can leave in `.replaceLst`) and comparison
  operators.

The third and fourth are the load-bearing ones: they fail if the filter ever
starts dropping a line that is not dead.

### Smaller changes

- `.normAssign()` lifted out of `rxUiGet.saemModelPred()` so it is directly
  testable, and switched from `parse(text=)[[1]]` to `str2lang()`.
  `parse()[[1]]` silently keeps only the *first* expression of a
  multi-expression string; `str2lang()` rejects it, and the `try()` fallback
  then leaves the line untouched, so it can only match a byte-identical line.
- Dropped the redundant `length(...) > 0L` guard (`%in%` already handles empty
  vectors) and trimmed the comment to the repo's comment-style rules.
- `NEWS.md` entry under `## Internal`.

### One note on the framing

The branch is titled `perf(saem)`, but I would not claim a measurable speedup.
Each removed alias costs two redundant stores in the generated C -- one in the
dydt region, one in `calc_lhs` -- and the totals move from 3891 to 3881 lines
(multi-endpoint), 2886 to 2878 (IOV), 2843 to 2837 (`linCmt()`): about 0.3%.
They are plain local `double` writes that a C compiler at `-O2` eliminates
anyway. It is a worthwhile codegen-hygiene cleanup; it just is not a
performance win, and the NEWS entry is worded accordingly.

### Independent review

Ran Antigravity (Gemini) over the diff. It found the live-store bug above
(though its stated trigger, IOV, was wrong -- I verified IOV models solve
identically) and the `==` regex weakness. Three further findings I checked and
rejected: a parameter that is also a compartment cannot reach this code
(rxode2 rejects the model first -- *"in the ini block but not in the model
block"*); a `.replaceLst` entry with an empty value emits `name <- ` in both
the old and new code, so it is not introduced here; and `str2lang()`'s `try()`
fallback makes the duplicate test strictly *more* conservative, not less. The
final pass over the current diff came back clean.

mattfidler:
Its been 4 days, other runs have completed, I'm unsure what is going on here.  I will merge and start testing more extensively for now.

Issue #931, opened 2026-08-15
Implement priors for the focei family


mattfidler:
Implement prior distributions for the focei family, then declare the support level.

## Why

The family is 32 methods that share one kernel: `focei`, `foce`, `focep`, `fo`, `foi`, `posthoc`, the mu-referenced `mfoce*` and IRLS `ifoce*` variants, the `*f` fast aliases (`R/foceiFast.R:41-164`), and the quadrature variants `laplace`/`mlaplace`/`ilaplace` and `agq`/`magq`/`iagq`(+`f`). `fo`/`foi` are FO linearizations evaluated by the same objective; `posthoc` is focei with `maxOuterIterations=0`; `laplace`/`agq` differ only by `nAGQ`.

That means there is exactly one place to add the term:

- `foceiLik0()` (`src/inner.cpp:4412`) sums the per-subject `lik[0]` into a population log-likelihood.
- `foceiOfv0()` (`src/inner.cpp:4439`) turns it into the objective at **`src/inner.cpp:4462`**: `double ret = -2*foceiLik0(theta);`

Adding `-2 * log p(theta)` immediately after that line — before the `scaleObjective` handling — covers all 32 methods.

## What a fix looks like

1. Add the prior contribution at `src/inner.cpp:4462` using the kernel from #929.
2. **The analytic gradient will silently disagree with the objective.** `numericGrad()` (`src/inner.cpp:5314`) finite-differences the objective, so it picks the prior up for free. When the analytic path is active it delegates to `analyticOuterGrad()` (`src/inner.cpp:5275`, set up in `R/foceiGradAnalytic.R`), which *does* call `foceiOfv0()` — but only for its side effects, to refresh the inner solutions at the current theta:

```cpp
// Ensure the inner solutions (eta*) and omega are current at this theta.
foceiOfv0(theta);                       // src/inner.cpp:5279
```

The returned value is discarded; the gradient itself comes from `analyticOuterGradDirect()`. So adding the prior at `:4462` changes the objective **without** changing the analytic gradient, and the optimizer is handed an objective and a gradient that no longer describe the same function. It will not error — it will just converge badly or not at all, and only when `optExpression`/analytic gradients are active.

That path needs an explicit `d/dtheta log p(theta)` term, and this is the bulk of the work, not the objective line.
3. Confirm `nlmixr2EnvSetup()` (`src/inner.cpp:7200`) reports a sensible `objective`/`logLik`/`AIC`/`BIC` — see #936.
[...]

PR #1014, opened 2026-08-25
Implement priors for the imp family (imp, impmap, qrpem) (#932)

### LINKED ISSUE #932: Implement priors for the imp family (imp, impmap, qrpem)
Implement prior distributions for the imp family (`imp`, `impmap`, `qrpem`).

## Why

All three share `impmapControl()` and `.impmapFamilyFit()` (`R/impmap.R:629`); `imp` is `impmap` with `mapIter=0` and `qrpem` is the quasi-random variant. The fit routes through `.foceiFamilyReturn()` into `foceiFitCpp_`, which dispatches to `impOuter()` (`src/imp.cpp:833`) rather than `foceiOuter`.

There is a real trap here. The EM has no objective line of its own — it *removes* any stashed objective and finalizes through `foceiOuterFinal` → `nlmixr2EnvSetup`, so the reported OFV is a FOCEi evaluation at the converged estimates. If the prior is added only to that reported objective (`src/inner.cpp:4462`, via #931), **the number would include a prior the estimates were never fit under** — a silently wrong answer, which is worse than refusing.

## What a fix looks like

Two separate pieces, and the issue is not done with only the first:

1. The reported objective picks up the prior via #931 automatically.
2. The **M-step inside `impOuter()`** (`src/imp.cpp:833`) needs the matching MAP correction, so the estimates are actually fit under the prior.

Add a test that fits a model with a strong prior and asserts the estimates moved toward it — not merely that the OFV changed. An OFV-only test would pass while the estimates were still the maximum-likelihood ones.

Declare the level on `imp`, `impmap` and `qrpem` individually (see #930).

Depends on #929, #930, #931.

mattfidler:
## Summary

Closes #932. Adds `ini({})` prior support to `est="imp"`/`"impmap"`/`"qrpem"`, in two phases:

**Theta priors** (`nlmixr2Priors = "theta"` initially): the shared M-step for this family is an importance-sampling EM, not FOCEi's outer optimizer, so the objective already picking up the prior (via #931's plumbing) was not enough on its own -- the estimates it reported would otherwise still be the maximum-likelihood ones. Each M-step update now folds in the prior's own score/curvature before taking its step:
- the non-mu structural/residual-error Newton step (`impPriorStructThetaCorrect()`)
- the mu-referenced covariate regression (`updateMuGroups()`) and the plain mu-intercept mean-shift (`impMuInterceptStep()`), via an FD-Hessian one-step Newton correction -- exact for a Gaussian prior, a reasonable one-step approximation otherwise (Cauchy, `multiNormal()`)

**Omega priors** (gate flipped to `"general"`): the Omega EM moment-average update gets the EXACT joint posterior mode for a conjugate `invWishart()` term (NONMEM's own `"nwpri"` convention or the textbook `"general"` one), and a one-step Fisher-scoring (One-Step-Late) correction -- reusing the same `Abar` construction FOCEi's own omega-prior gradient already computes -- for a normal prior directly on an omega element (`"tnpri"`) or a `multiNormal()` block mixing omega with theta.

**Bonus (upstream feature, zero nlmixr2est source changes)**: also adds test coverage for a marginal prior on an omega COVARIANCE (off-diagonal) element, e.g. `prior(eta.cl, eta.v) ~ dnorm(0, 0.1)` -- a companion feature landing in rxode2 (nlmixr2/rxode2#1286) and lotri (nlmixr2/lotri#59). Confirmed FOCEi's `foceiPriorOmegaGradAdd()` and this PR's own `impPriorOmegaCorrect()` already operate on the full `gradOmega`/`Omega` matrices generically, so both pick up the new capability for free once those two packages are upgraded -- proven by new tests that pass unmodified.

A code-review pass (independent model, via `agy`) on the theta-prior M-step work caught and fixed a real bug: `updateMuGroups()`'s active-set bound-clamping retry loop was evaluating a joint prior's gradient at a PINNED column's stale pre-M-step value instead of its current clamped value, dropping a cross-curvature term for a `multiNormal()` prior spanning a free and bound-pinned coefficient.

## Test plan

- [x] New `tests/testthat/test-imp-prior.R`: gate declarations, struct-theta Newton step, mu-intercept mean-shift, mu-covariate regression, invWishart conjugate closed form, generic/tnpri Fisher-scoring, off-diagonal covariance (generic path) -- each targets the specific M-step code path it claims, not just the finalize objective recompute.
- [x] New tests appended to `test-focei-prior.R` for the off-diagonal covariance case (fit-level + standalone gradient-formula central-difference check).
- [x] Full regression pass: `test-focei-prior.R`, `test-imp-prior.R`, `test-impmap.R`, `test-imp-combsens.R`, `test-imp-auto.R`, `test-priors-assert.R`, `test-qrpem.R` -- all pass, no regressions.

## Dependencies

The off-diagonal-covariance tests require nlmixr2/rxode2#1286 and nlmixr2/lotri#59 to be merged and installed to pass; everything else in this PR is self-contained.


mattfidler:
An independent review (`agy`/Gemini) of this PR's diff (with full awareness of the two earlier review rounds already applied to the theta/omega M-step commits) found nothing further -- confirmed the off-diagonal-covariance test additions are genuine (assert the estimate/covariance moved, not just that the fit completed), confirmed `foceiPriorOmegaGradAdd()`/`impPriorOmegaCorrect()` operate on the full gradient matrices generically as claimed, re-scanned all array indexing for bounds safety, and confirmed no shared-state mutation inside an OpenMP region.

PR #1018, opened 2026-08-25
refactor(fitSim): drop the focei simulation model that was built and thrown away

billdenney:
`getBaseSimModelFit.focei()` — shared by `foce`, `focep`, `fo`, `foi` and
`posthoc` — was behaviorally identical to `getBaseSimModelFit.default()` at
about three times the cost:

```r
getBaseSimModelFit.focei <- function(x) {
  obj <- x[[1]]
    getBaseSimModelFit.default(x) # fall back to basic with new method   # <- result dropped
  if (all(obj$ui$predDf$distribution == "norm")) {
    .expr <- eval(parse(text = paste0("quote(rxode2({",
                                      rxode2::rxNorm(obj$foceiModel$predOnly),
                                      "}))")))
    ...                                                                   # ~35 lines building .e2
    .expr[[2]] <- .e2                                                     # <- .expr never used again
  }
  getBaseSimModelFit.default(x)
}
```

Two things were dead: the `getBaseSimModelFit.default(x)` on the second line
(its value discarded, and the same call made again at the end), and the whole
`if` block — `.expr` is built, including a `rxNorm()` of the focei `predOnly`
model, and then goes out of scope unused.

So lowering a focei fit to a simulation model lowered it three times. That ran
on every `rxSolve(fit, ...)` call before nlmixr2/rxode2#1289 was fixed, and
each of those lowerings also leaked (nlmixr2/rxode2#1289 and the rxode2
`fix/model-lowering-leaks` branch).

### What this does

Makes the focei-family methods aliases of the default, which is what they
already amounted to, and drops `.isAssignExpr()` and
`.replaceThetaEtaWithNamed()`, whose only caller was the dead block. The
methods are kept rather than deleted, so dispatch and the exported method list
are unchanged.

The discarded expression is preserved in history if the unfinished feature it
sketches — a focei-specific simulation model built from `predOnly` with
per-endpoint `error.<var>*(CMT==n)` terms — is ever wanted. If it is not, the
methods can go away entirely and dispatch will land on the default.

### Tests

`tests/testthat/test-fitSim.R`:

* the focei-family methods **are** the default method (`expect_identical`, so a
  body growing back fails);
* the default lowers the fit's own `$ui` (fit-shaped stand-in, no fitting);
* a focei fit's `$simulationModel` is unchanged by this (fixture fit).

15/15 pass in `test-fitSim|simulation|nmobj|rxsolve`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,15 @@
 
 ## Internal
 
+- `getBaseSimModelFit()` for the focei family (`focei`, `foce`, `focep`, `fo`,
+  `foi`, `posthoc`) no longer does three times the work for the same answer.
+  The method built a `predOnly`-based simulation model expression and then
+  discarded it, and called `getBaseSimModelFit.default()` twice -- once with
+  the result thrown away -- so lowering a focei fit to a simulation model
+  lowered it three times, one of those through a `rxNorm()` of the focei
+  `predOnly` model.  These methods are now aliases of the default, which is
+  what they already amounted to.
+
 - A covariate whose value is carried on the model in `rxode2::rxForcedPars()` is
   no longer required to be a column of the data.  Such a covariate is supplied
   by the model itself, so demanding it from the data rejected a well-specified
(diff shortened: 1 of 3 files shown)


mattfidler:
Thanks @billdenney

PR #1019, opened 2026-08-25
fix(focei): size the theta-reset buffers from the allocation, not a copy of it

billdenney:
`saveIntoEnvrionment()` / `restoreFromEnvrionment()` (src/inner.cpp) save five
focei working buffers whole and copy them back when the fit resets its thetas.
Each one computed its length from a *second copy* of the formula the allocation
uses, and every copy had drifted from the layout it described:

| buffer | allocated | saved |
| --- | --- | --- |
| `etaTrans` | `3*neta + (4 + (mixIdxN != 0))*(ntheta+omegan)` | `3*neta + 3*(ntheta+omegan)` |
| `fullTheta` | `4*(ntheta+omegan) + 2*aqn*neta` | `4*(ntheta+omegan)` |
| `etaUpper` | 11 eta blocks, `sum(nobs_i^2)`, `neta*6`, `+3*nall` (`gcHff`/`gcHfr`/`gcHrr`), mixture-expanded `nsub`/`nall` | 10 blocks, `nall^2`, `neta*5`, no `gcH*`, unexpanded counts |
| `gillRet` | `... + muRefN + mixIdxN + nsub` | those terms omitted |
| `gillDf` | `... + (mixIdxN+1)*(nsub+1) + mixIdxN*(nsub+1)` | those terms omitted |

Four saved **less** than had been allocated, so a theta reset silently restored
a truncated buffer and left the tail — `etaFD`/`mixTrans`, `aqx`/`aqw`,
`muRefEtaCovSkipReset`/`mixIdx`/`skipCov`, `mixProb`/`mixProbGrad`/`gillDf2` —
holding whatever the interrupted fit had put there.

The eta block claimed **more** than had been allocated, so saving read past the
end of the block and restoring wrote that overshoot back over the heap.

### How it was found

An AddressSanitizer build of rxode2 + nlmixr2est, with the *runtime-compiled
model DLLs* instrumented too (`R_MAKEVARS_USER` pointing at ASAN flags for the
test run — without that, ASAN sees nothing, because the wild access comes
through code compiled at solve time):

```
==2429496==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 88640 at 0x52a00c821118 thread T0
    #5 saveIntoEnvrionment(...)      src/inner.cpp:22358
    #6 thetaReset00                  src/inner.cpp:4079
    #7 thetaReset0                   src/inner.cpp:4223
    #8 thetaResetZero()               src/inner.cpp:4257
    #9 innerOpt()                     src/inner.cpp:4713
   #10 foceiLik0 <- foceiObjFromLik0 <- foceiOfv0 <- foceiOfvOptim <- lbfgsb3C_

0x52a00c821118 is located 0 bytes after 20248-byte region
allocated by thread T0 here:
    #2 foceiSetupEta_                src/inner.cpp:6573
```

### Impact

`tests/testthat/test-matexp.R` aborted the R process on `origin/main` —
`malloc(): corrupted double-linked list` (rc=134) on one build, `*** caught
segfault ***` with recursive gc (rc=139) on another, and it took the rest of
the test run down with it. With this fix that file runs to completion, exit 0,
no corruption markers.

The remaining assertion failures in `test-matexp.R` are a **separate, older**
problem — repeating the identical focei fit in one session gives different
answers, which reproduces on released nlmixr2est 7.0.3 + rxode2 5.1.7 with
default `foceiControl()` (filed separately).

### What this does

Records each buffer's allocated length beside its pointer (`nEtaTrans`,
`nFullTheta`, `etaBufferN`, `nGillRet`, `nGillDf`), uses it for both the save
and the restore, and makes a restore whose saved length does not match the
current allocation an **error** rather than a copy — a mismatch means the
problem changed under the save, which cannot be papered over safely.

### Tests

`tests/testthat/test-focei-theta-reset.R` fits the scenario that crashed and
asserts it completes with finite estimates. It deliberately does **not** assert
that two identical fits agree: they do not, for the separate reason above, so
asserting it here would fail for something this fix does not address. Under an
ASAN build this test is what catches the overflow directly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,24 +2,16 @@
 
 ## Bug fixes
 
-- The focei theta-reset path no longer reads and writes past the end of its
-  working buffers.  `saveIntoEnvrionment()`/`restoreFromEnvrionment()`
-  (src/inner.cpp) re-derived the length of each buffer they save from a copy of
-  the allocation's formula, and every copy had fallen behind the layout it
-  described: four saved *less* than had been allocated, so a reset silently
-  dropped the tail of those blocks (`etaFD`/`mixTrans`, `aqx`/`aqw`,
-  `muRefEtaCovSkipReset`/`mixIdx`/`skipCov`,
-  `mixProb`/`mixProbGrad`/`gillDf2`), and the eta block's copy claimed *more*
-  (ten eta blocks instead of eleven, `nall^2` instead of `sum(nobs_i^2)`,
-  `neta*5` instead of `neta*6`, no `gcHff`/`gcHfr`/`gcHrr`, and unexpanded
-  subject/observation counts), so saving read past the end of the block and
-  restoring wrote the overshoot back.  AddressSanitizer reports the read as
-  `heap-buffer-overflow ... 0 bytes after 20248-byte region`; the write
-  corrupted whatever followed the block and aborted the process outright in
-  `test-matexp.R` (`malloc(): corrupted double-linked list` /
-  `*** caught segfault ***`), which now runs to completion.  Each length is
-  now recorded where the buffer is allocated and used for both directions, and
-  a restore whose saved length does not match errors instead of copying.
+- Fixed a heap overflow in the FOCEi theta-reset path.  The buffers it saves
+  and copies back on a restart each had their length re-derived from a second
+  copy of the allocation's formula, and every copy had fallen behind the layout
+  it described -- most damagingly the eta block, which claimed `nall^2` where
+  `sum(nobs_i^2)` had been allocated, so a reset read past the end of the block
+  and wrote the overshoot back.  Depending on what followed it in the heap, a
+  fit that reset its thetas could return truncated state, corrupt an unrelated
+  allocation, or abort the R process outright (`test-matexp.R` did the last of
+  these).  Each length now comes from the allocation itself, and a restore whose
+  saved length does not match the current one errors instead of copying.
 
 ## Internal
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,26 @@
 # nlmixr2est 7.0.3
 
+## Bug fixes
+
+- The focei theta-reset path no longer reads and writes past the end of its
+  working buffers.  `saveIntoEnvrionment()`/`restoreFromEnvrionment()`
+  (src/inner.cpp) re-derived the length of each buffer they save from a copy of
+  the allocation's formula, and every copy had fallen behind the layout it
+  described: four saved *less* than had been allocated, so a reset silently
+  dropped the tail of those blocks (`etaFD`/`mixTrans`, `aqx`/`aqw`,
+  `muRefEtaCovSkipReset`/`mixIdx`/`skipCov`,
+  `mixProb`/`mixProbGrad`/`gillDf2`), and the eta block's copy claimed *more*
+  (ten eta blocks instead of eleven, `nall^2` instead of `sum(nobs_i^2)`,
+  `neta*5` instead of `neta*6`, no `gcHff`/`gcHfr`/`gcHrr`, and unexpanded
+  subject/observation counts), so saving read past the end of the block and
+  restoring wrote the overshoot back.  AddressSanitizer reports the read as
+  `heap-buffer-overflow ... 0 bytes after 20248-byte region`; the write
+  corrupted whatever followed the block and aborted the process outright in
+  `test-matexp.R` (`malloc(): corrupted double-linked list` /
+  `*** caught segfault ***`), which now runs to completion.  Each length is
+  now recorded where the buffer is allocated and used for both directions, and
+  a restore whose saved length does not match errors instead of copying.
+
 ## Internal
 
 - A covariate whose value is carried on the model in `rxode2::rxForcedPars()` is
--- a/src/inner.cpp
+++ b/src/inner.cpp
@@ -6304,8 +6304,10 @@ static inline void foceiSetupTrans_(CharacterVector pars){
   // doubles through R to rewrite what is already there.  nFullThetaSave is the
   // saved prefix; nFullTheta is derived from it so the two cannot drift apart.
   op_focei.nFullThetaSave = (size_t)4*(op_focei.ntheta+op_focei.omegan);
+  // (size_t) on each factor: _aqn is nAGQ^neta and neta is unsigned int, so the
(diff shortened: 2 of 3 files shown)


mattfidler:
It seems to stomp the agq.  Let me look a bit closer.

billdenney:
This requires nlmixr2/rxode2#1294

If tested after that, it may work with agq.

mattfidler:
## Review follow-up: AGQ, and two things the tests were not catching

Reviewed with attention to the AGQ path, then re-reviewed with Antigravity (Gemini) twice. Five commits pushed on top of `f77008076`.

### 1. The AGQ node grid should not be in the save at all

`fullTheta` is one block: `fullTheta | theta | initPar | scaleC | aqx | aqw`. The `aqx`/`aqw` tail is the AGQ Gauss-Hermite grid (`nAGQ^neta` nodes x `neta` x 2), and `setupAq1_()` (`src/inner.cpp:12309`) refills it from the fit environment's `qx`/`qw` on **every** `foceiFitCpp_` entry -- which happens *before* `restoreFromEnvironment()` at `:12370`. Saving it round-tripped up to ~1.5 MB (neta=6, nAGQ=5) through R per reset only to rewrite values that were already correct.

`nFullThetaSave` is now recorded at the allocation and `nFullTheta` is **derived from it** (`nFullThetaSave + 2*_aqn*neta`), so the two cannot drift apart -- which is the exact failure mode this PR exists to fix.

**The `fullTheta` row of the table is AGQ-only and was benign.** `2*aqn*neta` is zero unless `nAGQ > 0` (`foceiControl()` defaults `nAGQ=0`, `R/foceiControl.R:1014`), so on ordinary FOCEI/FOCE the old save length was already exactly the allocation; and where it was short, the dropped tail was rebuilt before it was read. `_aqn` is the only AGQ-dependent term in any of the five buffers -- `etaTrans`, `etaUpper`, `gillRet` and `gillDf` have none. **The `etaUpper` row is the one that mattered.**

### 2. The test never reached the code it covers

`resetThetaP` defaults to `0`, which leaves `resetThetaSize` infinite, and `thetaReset()` returns on `std::isinf(size)` before it can fire. Under default `foceiControl()` the ETA-drift reset **cannot happen**. The test fit converged straight through and never called `saveIntoEnvrionment()`/`restoreFromEnvrionment()` at all. Found by adding the assertion and watching it fail.

The same applies to the repeated-fit assertion added in `f77008076`: it passes, but on a fit that never resets, so it was comparing two runs that never exercised the round trip. It is now on a scenario that does reset.

The test now switches the reset on (`resetThetaP`/`resetThetaCheckPer`, the recipe `test-focei-theta-reset-bounds.R` already uses), counts the resets off the emitted messages and asserts the count, checks the restored state is well formed, compares two reset-triggering runs, and adds an `nAGQ>0` case. Counting has to wrap `nlmixr()` directly -- the `.nlmixr()` helper's `suppressMessages()` muffles the reset messages before an outer handler sees them.

Instrumenting the scenario the test now runs:

```
PROBE etaBufferN=3630 oldFormula=22444 overshoot=18814
```

18,814 doubles (~150 KB) read and written past the end of a 3,630-double block, on each reset the test triggers.

### 3. Evidence the fix does what it claims

Building `origin/main`'s `inner.cpp` in this tree and running the *pre-existing* `test-focei-theta-reset-bounds.R`:

```
focei-theta-reset-bounds: corrupted double-linked list
```

The process aborts. With this branch it does not. `test-matexp.R` also runs to completion here.

### 4. Determinism, measured

Worth correcting the description: after this fix, repeating an identical **non-resetting** fit reproduces bit for bit (`-112.6066161832` twice, identical `fixef`). A **reset-heavy** fit still drifts ~1e-6 relative between runs in one session. So the non-determinism is associated with the reset path rather than with fits generally. The new comparison uses a 1e-3 tolerance -- measured against that drift, still four orders of magnitude tighter than anything a truncated or overrunning restore would produce.

### 5. Smaller items

- `saveIntoEnvrionment`/`restoreFromEnvrionment` -> corrected spelling (NEWS referenced the old names).
- `2*(_aqn*op_focei.neta)` formed its product in `unsigned int` before widening; now `size_t` per factor. Unreachable in practice -- `.agq()` builds the grid with `expand.grid` first and would exhaust memory long before -- but free.
- `foceiCheckRestoreN()` passes counts as strings rather than `%llu`.
- NEWS entry trimmed to the bullet style `CLAUDE.md` asks for.
- Test kept out of `.slowBatches`: both files together run in ~14s.

### Antigravity findings, triaged

Two passes. Everything it raised about the AGQ reasoning, the length checks, buffer aliasing, stale lengths and the `.thetaReset` singleton came back clean and matched my own reading. Of its three concrete findings:

- **Integer overflow at `inner.cpp:6310`** -- real, already fixed in `4de6411`.
- **`Rcpp::stop` UB with `std::to_string`** -- false. `Rcpp/exceptions.h:204` routes to `tfm::format` (tinyformat, `std::ostream`-based); there is no `vsnprintf` in the bundled tinyformat. Compiled and ran the exact construct: correct message, no crash. Its suggested trigger `foceiControl(nF=1)` is also unreachable -- `nF` is in `.foceiControlInternal`, and `"nF" %in% names(foceiControl(nF=1))` is `FALSE`.
- **`.nlmixrFoceiRestartIfNeeded` forgets `control$nF2 <- 0`** -- false. There is no `nF2` R-side field; `op_focei.nF2 = as<int>(foceiO["nF"])` (`inner.cpp:7066`), and `R/focei.R:3483` already sets `.ret$control$nF <- 0`.
- Its test-robustness point was fair: the reset fired in 12/12 configurations measured, but a platform whose BLAS takes a different trajectory could converge straight through, so a non-trigger now `skip()`s with an explicit reason rather than failing -- and can never pass quietly.

### Verification

`focei-theta-reset` (28 assertions, stable across repeated runs), `focei-theta-reset-bounds`, `agq-cov`, `agq-fast-grad`, `matexp` all pass. habit-hooks reports `unused-variable` and `oversized-file` on `src/inner.cpp`; checked each line number against the diff hunks -- none fall inside this change.

One caveat on the environment: another local session reinstalled rxode2 (a working-tree build) partway through this run. `test-focei-theta-reset-bounds.R` passed twice on this branch before that swap and afterwards errors with "Maximum number of theta resets (10) exceeded". On `origin/main` under the same rxode2 it corrupts the heap instead. Flagging it rather than chasing it -- it is not this PR.

PR #1021, opened 2026-08-26
test(focei): lock in that the same fit twice gives the same answer

### LINKED ISSUE #1015: A fit's result depends on its position in the R session (any prior fit changes the next fit's outcome)
Found while diagnosing an unrelated test failure; **pre-existing**, reproduced identically on rxode2 main and on a pre-linCmt-stack build (bisect below), so this is not a recent regression.

## Symptom

The same FOCEi fit -- identical model object, identical data, identical control, 200 iterations -- gives a different answer depending on whether anything else was fitted earlier in the same R session:

| position in session | objective | event etas (eta.f / eta.lag) |
|---|---|---|
| first fit in a fresh session | -29.533874 | nonzero (max 0.0635 / 0.0958) |
| any subsequent fit | -29.721341 | approximately 0 (abs(eta) < 5e-7) |

The two results are at different (theta, omega) points, so they are two different converged optima, not two evaluations of one surface.

## What it is not

- **Not UI-object mutation.** Rebuilding the model object (a fresh `linToOde(rxode2(...))`) before every run does NOT restore the first-fit behavior.
- **Not random.** It is deterministic by position: the first 200-iteration fit in a fresh session gives the first answer, every later one gives the second.
- **Not caused by the recent linCmt work.** Bisected across the linCmt sensitivity stack (pre-#1281 `a46e4ed4d` vs main `6098c614f`): both builds produce both values, identical to six decimal places.

## The discriminating experiment

Running a cheap posthoc fit *before* the 200-iteration fit makes that fit produce the second answer. So **any prior fit consumes the first-call state** -- the dependence is session-global (model cache, or first-solve initialization), not per-object.

## Why it matters

A fit's converged result should not depend on what was fitted before it in the same session. It makes results non-reproducible between an interactive session and a fresh script, and it can silently change which local optimum a weakly identified model finds.

## Reproducer

A 1-cmt oral model with etas on `f()` and `alag()`, 6 subjects, fitted twice in one session with `maxOuterIterations = 200`; compare objectives and the event etas. (The model used here was the ODE reference arm from `tests/testthat/test-focei-lincmt-carry-jump-fit.R` on the `lincmt-speed` branch, built via `linToOde()` and solved with `useLinCmt = FALSE`.)

### LINKED ISSUE #1020: Fitting the same data twice in one session gives different focei results
## Summary

Running the identical `nlmixr2(model, data, est = "focei")` call twice in one R
session returns different objective functions and different parameter
estimates, and the later fits often stop at (or barely off) their initial
estimates. There is nothing stochastic in the call — same model object, same
data frame, same control — so this is state carried between fits.

Reproduces on **released** versions (nlmixr2est 7.0.3 + rxode2 5.1.7) with a
plain `foceiControl(print = 0)`, so it is not specific to a development branch,
to `covMethod = "analytic"`, or to `sigdig`.

Found while fixing the heap overflow in the theta-reset save/restore path
(PR `fix/focei-theta-reset-eta-buffer`); that fix removes the crash but not
this — it reproduces with and without it.

## Reprex

```r
library(nlmixr2est)
library(rxode2)

mod <- function() {
  ini({ tka <- 0.45; tcl <- 1.0; tv <- 3.45; eta.ka ~ 0.09; add.sd <- 0.7 })
  model({
    ka <- exp(tka + eta.ka); cl <- exp(tcl); v <- exp(tv)
    d/dt(depot)   <- -ka * depot
    d/dt(central) <-  ka * depot - cl / v * central
    cp <- central / v
    cp ~ add(add.sd)
  })
}

# simulated once, from parameters away from the initial estimates
dat <- rxode2::rxWithSeed(1234, {
  ev  <- rxode2::et(amt = 320, cmt = "depot", id = 1:6) |> rxode2::et(seq(0.5, 24, by = 1.5))
  sim <- rxode2::rxSolve(mod(), ev, params = c(tka = 0.6, tcl = 1.1, tv = 3.6))
  d <- as.data.frame(sim)[, c("id", "time", "cp")]
  d$cp <- d$cp + stats::rnorm(nrow(d), 0, 0.3)
  names(d) <- c("ID", "TIME", "DV"); d$AMT <- 0; d$EVID <- 0
  d <- rbind(data.frame(ID = 1:6, TIME = 0, DV = NA, AMT = 320, EVID = 1), d)
  d[order(d$ID, d$TIME, -d$EVID), ]
})

ctl <- foceiControl(print = 0)
for (i in 1:3) {
  f <- suppressMessages(suppressWarnings(nlmixr2(mod(), dat, est = "focei", control = ctl)))
  cat(sprintf("fit%d objf=%12.4f fixef=%s\n", i, f$objf,
              paste(sprintf("%.4f", fixef(f)), collapse = " ")))
}
```

## Output

Released nlmixr2est 7.0.3 + rxode2 5.1.7, default control:

```
fit1 objf=   1606.9170 fixef=0.6500 1.0000 3.4500 0.7000
fit2 objf=    801.6084 fixef=0.4566 1.0205 3.4538 0.6152
fit3 objf=    620.9300 fixef=0.4500 1.0000 3.4500 0.6999
```

Same versions, `foceiControl(print = 0, covMethod = "analytic", sigdig = 6)`:

```
fit1 objf=    813.1735 fixef=0.4545 1.1379 3.5375 0.7014
fit2 objf=   1729.1049 fixef=0.4500 1.2000 3.4500 0.7000
fit3 objf=   2132.1055 fixef=0.4500 1.2000 3.4500 0.7000
```

Note the estimates that sit exactly on the initial values (`0.4500 1.0000
3.4500 0.7000`) while the objective function still moves — the fit is not
converging, and what it reports depends on what ran before it in the session.

## Why it matters

* Any script that fits more than one model in a session may be reporting the
  wrong numbers for every fit after the first.
* It makes the `matExp()`-vs-ODE comparisons in
  `tests/testthat/test-matexp.R` fail more or less at random (that file has 25
  to 33 failures depending on the run, and its `.fO` ODE fits are the ones
  stuck at their initial estimates).
* Any benchmark or regression test that fits repeatedly in one session is
  measuring noise.

## Suggested direction

Something survives between `nlmixr2()` calls — `op_focei`, `inds_focei`, the
solved-object globals, or the parse/solve state on the rxode2 side. Fitting
each model in a fresh session should give the "fit1" answer every time, which
is a quick way to confirm that reading: run the loop above with `callr::r()`
per fit and compare.

## Session

```
nlmixr2est 7.0.3, rxode2 5.1.7 (both the CRAN-installed build and github main)
R 4.6.1, x86_64-pc-linux-gnu
```

billdenney:
Closes #1020

Regression test for issue 1020: fitting the same data twice in one session has
to give the same answer.

On a machine with at least twice as many threads as the problem has subjects,
two identical `nlmixr2(..., est = "focei")` calls returned different objective
functions and different estimates, and the optimizer frequently stopped at (or
beside) the initial estimates.

FOCEi solves one subject at a time through rxode2's `ind_solve()` and re-sorts
the subjects around every inner-optimization and every gradient loop. Two rxode2
bugs met there: `sortIds()` ordered subjects by their *measured* solve time, so
the order was a function of wall-clock timing, and several of the per-individual
drivers read `ind_solve()`'s **subject id** as a **position** in `rx->ordId` --
so once the order stopped being the identity, the wrong individual was
integrated. The reorder only fires above `cores >= nsub*nsim*throttle`, which is
why the problem appeared only on wide machines.

**No rxode2 dependency.** The id convention is fixed in nlmixr2/rxode2#1292 and
#1293, both merged, and this file passes `FAIL 0 | PASS 12` against rxode2 `main`
(`8e9dda2`). An earlier version of this description listed
`fix/ind-solve-subject-id` as a blocker; that is stale. Note also that the
unmerged `fix/deterministic-solve-order` branch must **not** be merged under this
test -- it makes every fit return a different, badly wrong objective
(nlmixr2/rxode2#1303).

Both the ODE drivers and the analytic `linCmt()` drivers were affected, so
`tests/testthat/test-focei-reproducible.R` covers both: the same structural model
is fitted as an ODE and as `linCmt()`.

Two things the test has to do deliberately, or it proves nothing:

* **raise the thread count** to 2x the subject count for its own duration,
  because the reorder does not trigger below that. It sets the count first and
  skips on what `setRxThreads()` actually granted -- `rxCores()` is an alias of
  `getRxThreads()`, i.e. the current *setting* (50% of logical CPUs by default)
  rather than the CPU count, so asking before setting made the test skip itself
  on a 22-thread machine.
* **give the subjects different numbers of samples** (subject `i` gets `8 + 2i`).
  Equal-sized subjects sort to the identity order, and an identity `ordId` is
  exactly where reading the argument as an id and reading it as a position agree.

It then asserts that

* two fits of the same data agree -- objective function and fixed effects,
* a third fit at 2 threads agrees with them, so the answer does not depend on
  how many threads are available,
* the fit moved off its initial estimates at all (the failure mode was an
  optimizer that gave up where it started), and
* it converges to the parameters the data was simulated from.

The problem is sized to the machine, and the file runs in weekly slow batch 5
(six fits), not the essential push/PR subset.

## Does not close #1015

The `Closes` line used to point at #1015; that was wrong, and #1015 is a
different bug that this test cannot see. #1015 is about a fit's dependence on its
*position in the session* -- the FIRST fit in a fresh session differs from every
later one. It still reproduces on rxode2 `main` today:

```
fit1 objf=    -29.006916  max|eta.f|=0.0834     max|eta.lag|=0.131
fit2 objf=    -29.641287  max|eta.f|=1.17e-07   max|eta.lag|=1.85e-08
fit3 objf=    -29.641287  max|eta.f|=1.17e-07   max|eta.lag|=1.85e-08
```

(the `linToOde()` arm of `test-focei-lincmt-carry-jump-fit.R`, `useLinCmt = FALSE`,
`maxOuterIterations = 200`; the event etas collapse to zero after the first fit.)

Note `fit2 == fit3`. This file compares two *consecutive* fits, and its first fit
is never the session's first -- the `helper-zzz-fits.R` fixtures run well before
it -- so it is structurally blind to #1015. Catching that one needs a fresh-session
harness, which is a separate piece of work.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,19 @@
 # nlmixr2est 7.0.3
 
+## Bug fixes
+
+- Fitting the same data twice in one session gives the same answer again
+  (#1020).  On a machine with at least twice as many threads as the problem
+  has subjects, the objective function and the estimates differed between two
+  identical `nlmixr2(..., est = "focei")` calls, and the optimizer frequently
+  stopped at (or beside) the initial estimates.  The cause is in rxode2 --
+  `sortIds()` ordered subjects by their *measured* solve time, so FOCEi, which
+  re-sorts around every inner-optimization and gradient loop, handed its
+  subjects to different threads on every pass and carried different per-thread
+  solver state into each likelihood and gradient.  It is fixed there by
+  ordering on the number of events a subject is solved over; this package
+  gains the regression test.
+
 ## Internal
 
 - A covariate whose value is carried on the model in `rxode2::rxForcedPars()` is
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,12 +6,12 @@
   (#1020).  On a machine with at least twice as many threads as the problem
   has subjects, the objective function and the estimates differed between two
   identical `nlmixr2(..., est = "focei")` calls, and the optimizer frequently
-  stopped at (or beside) the initial estimates.  The cause is in rxode2 --
-  `sortIds()` ordered subjects by their *measured* solve time, so FOCEi, which
-  re-sorts around every inner-optimization and gradient loop, handed its
-  subjects to different threads on every pass and carried different per-thread
-  solver state into each likelihood and gradient.  It is fixed there by
-  ordering on the number of events a subject is solved over; this package
+  stopped at (or beside) the initial estimates.  The cause is in rxode2, where
+  two bugs met on the path FOCEi takes: `sortIds()` ordered subjects by their
+  *measured* solve time, so the order was a function of wall-clock timing, and
+  several per-individual drivers read `ind_solve()`'s subject id as a position
+  in the reordered `rx->ordId` -- so once the order stopped being the identity
+  the wrong individual was integrated.  The fixes are in rxode2; this package
   gains the regression test.
 - Fixed a heap overflow in the FOCEi theta-reset path.  The buffers it saves
   and copies back on a restart each had their length re-derived from a second
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -102,7 +102,7 @@ if (identical(Sys.info()[["sysname"]], "Darwin")) {
   c(
     "focei-wang2007-boxcox-lnorm", "nlme-cov", "agq-fast-grad",
     "focei-ll-fast-grad-fit", "focei-fast-methods-fit", "odeswap-fit",
-    "focei-factr-fit"
+    "focei-factr-fit", "focei-reproducible"
   ),
   # batch 6
   c(
--- a/tests/testthat/test-focei-reproducible.R
+++ b/tests/testthat/test-focei-reproducible.R
@@ -1,18 +1,32 @@
 ## Fitting the same data twice in one session has to give the same answer
 ## (nlmixr2est issue 1020).
 ##
-## FOCEi solves one subject at a time through rxode2's `ind_solve()`, whose
-## argument is a subject id.  Several of rxode2's per-individual drivers read
-## that argument as a position in `rx->ordId` instead, and `rx->ordId` stops
-## being the identity once rxode2 reorders subjects most-expensive-first --
-## which it does when there are at least `throttle` (2) times more threads than
-## subjects.  Past that threshold the wrong individual was integrated, so the
-## likelihood, its gradient, and hence the whole fit depended on a solve order
-## that comes from wall-clock timing.  Fixed in rxode2; these tests need that
-## fix to pass.
+## FOCEi solves one subject at a time through rxode2's `ind_solve()` and
+## re-sorts the subjects around every inner-optimization and every gradient
+## loop.  Two rxode2 bugs met there:
 ##
-## The tests need threads >= 2 * subjects to exercise the path at all, so they
-## raise the thread count for their own duration.
+##   * `sortIds()` ordered the subjects most-expensive-first from
+##     `ind->solveTime`, the accumulated `clock()` time of their earlier
+##     solves, so the order was a function of wall-clock timing -- it differed
+##     between two runs of the same problem and between successive passes of
+##     one run.
+##   * `ind_solve()`'s argument is a subject id, but several of rxode2's
+##     per-individual drivers read it as a position in `rx->ordId`.  While
+##     `ordId` is the identity the two agree; once it is not, the wrong
+##     individual is integrated.
+##
+## The reorder only fires when there are at least `throttle` (2) times more
+## threads than subjects, so past that threshold the likelihood, its gradient
+## and hence the whole fit followed a solve order that came from thread
+## timing.  Both are fixed in rxode2; these tests are the gate.
+##
+## Two things the tests have to do deliberately, or they prove nothing:
+##
+##   * raise the thread count to 2x the subject count for their own duration,
+##     because the reorder does not trigger below it, and
+##   * give the subjects DIFFERENT numbers of samples.  Equal-sized subjects
+##     sort to the identity order, and an identity `ordId` hides the
+##     id-vs-position confusion completely.
 
 nmTest({
 
@@ -54,12 +68,22 @@ nmTest({
     })
   }
 
+  .ini <- c(tka = 0.45, tcl = 1.0, tv = 3.45, add.sd = 0.7)
+
   # simulated from parameters away from the initial estimates, so the optimizer
-  # has to travel -- a fit that stops where it started hides the problem
-  .reproData <- function(model, nid = 6, seed = 1234) {
+  # has to travel -- a fit that stops where it started hides the problem.
+  # Subject i gets 8 + 2i samples: the sizes have to differ for the solve order
+  # to be anything but the identity.
+  .reproData <- function(model, nid, seed = 1234) {
     rxode2::rxWithSeed(seed, {
-      .ev <- rxode2::et(amt = 320, cmt = "depot", id = seq_len(nid))
-      .ev <- rxode2::et(.ev, seq(0.5, 24, by = 1.5))
+      .ev <- do.call(
+        rbind,
+        lapply(seq_len(nid), function(.i) {
+          rbind(data.frame(id = .i, time = 0, amt = 320, evid = 1),
+                data.frame(id = .i,
+                           time = seq(0.5, 24, length.out = 8L + 2L * .i),
+                           amt = 0, evid = 0))
+        }))
       .sim <- rxode2::rxSolve(model, .ev,
                               params = c(tka = 0.6, tcl = 1.1, tv = 3.6))
       .dat <- as.data.frame(.sim)[, c("id", "time", "cp")]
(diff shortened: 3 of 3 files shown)


mattfidler:
## Antigravity review (gemini-3.1-pro-high), 2 rounds -- pushed 675e435

Round 1 returned four findings; three survived triage and are fixed here. Round 2
against the updated diff came back clean (its artifact was returned as a
`conversation://` link rather than a file, so that round is the printed summary
only, not a full `plan.md`).

**The headline: with the review fixes applied, this test runs -- and it fails.
#1020 is not closed.**

### Confirmed and fixed

**1. The skip guard skipped the machines the test was written for.**
`rxode2::rxCores()` is an alias of `getRxThreads()` (`rxode2` `R/utils.R:711`), i.e.
the current thread *setting*, which defaults to 50% of logical CPUs -- not the
number of CPUs. The guard asked it *before* raising the thread count, so on this
22-thread box it read 11, compared against the required 12, and skipped. The file
had never executed here. It now sets the thread count first and skips on what
`setRxThreads()` actually granted, which also covers the silent clamp
(`setRxThreads(1000L)` quietly yields `num_procs`, no error).

**2. Equal-sized subjects made the test blind to the bug it targets.**
Every subject had the same 16 samples. Once `sortIds()` orders on problem size
instead of measured solve time, equal-sized subjects sort to the *identity* order
-- and an identity `ordId` is exactly the case where "read the argument as an id"
and "read it as a position" agree. Subject `i` now gets `8 + 2i` samples, so
`ordId` is a genuine permutation.

**3. `NEWS.md` and the file header gave different root causes.** `NEWS.md` blamed
`sortIds()`'s timing-based order; the header blamed the id-vs-position confusion
in the per-individual drivers. Both are real and they interlock (the second is
only reachable once the first makes `ordId` non-identity). Both entries now say so,
and `NEWS.md` no longer enumerates a fix set -- see below for why.

**4. CI batching.** Six FOCEi fits were sitting in the essential push/PR subset.
Added to weekly slow batch 5 per `CLAUDE.md`.

Also: sized the problem to the machine (`.reproNid()`), factored the two
near-identical test bodies into one helper, and added `add = TRUE` to the
`on.exit()` handlers to match the convention in the rest of the suite.

### The gate is red

Measured against `nlmixr2est` 7.0.3 with rxode2 built from
`fix/deterministic-solve-order` (main + `sortIds()` ordered by `n_all_times`).
One dataset, six subjects of 11/13/15/17/19/21 records, three successive
`est="focei"` fits per thread count:

```
thr= 2  objf:  -121.2606  -121.2606  -121.2606
thr= 6  objf:  -121.2606  -121.2606  -121.2606
thr=11  objf:  -121.2606  -121.2606  -121.2606
thr=12  objf:   760.5205   780.2529  2024.1438
thr=16  objf:  1137.4662  1852.8414  1006.2266
```

The break is exactly at `sortIds()`'s `cores >= nsub * throttle` (6 * 2 = 12).
Below it the fit is reproducible and lands on the simulating values
(0.513/1.091/3.597 for 0.6/1.1/3.6). At and above it every fit differs and the
optimizer stops near the initial estimates -- the original #1020 symptom.

Cherry-picking nlmixr2/rxode2#1294 (`fix/per-subject-tolerance-state`) on top does
not change this:

```
thr=12  objf:   891.5430  1571.3926  1616.4593
thr=16  objf:   215.4191  1824.1926  1012.5466
```

Same threads, same parallelism, same data-generating code -- the *only* difference
between the passing and failing runs is whether `ordId` is a non-identity
permutation. So the remaining defect is on the permuted-`ordId` solve path, not in
threading as such. `nlmixr2est`'s own use looks right: `src/inner.cpp` consistently
takes `_gid = getOrdId(rx, _i) - 1` and passes that id downward.

### Two things to note before merging

* The dependency in the PR body needs updating. nlmixr2/rxode2#1292 and #1293 are
  merged, but the `sortIds()` ordering commit (`5ec4d88f`) is on
  `fix/deterministic-solve-order`, which is not in rxode2 `main` and has no open
  PR. rxode2 `main` still sorts on `ind->solveTime`.
* The `NEWS.md` bullet is written in the past tense of a shipped fix. It should
  land with the rxode2 fix that actually makes this file green, not before.

The test itself is now correct and I would keep it as is -- it is doing its job by
failing.

mattfidler:
## Correction: the gate is GREEN on rxode2 main, and this PR's dependency should be dropped

Filed the rxode2 side as nlmixr2/rxode2#1303.

**My previous comment was wrong on its main conclusion.** I measured the failing
runs against rxode2 built from `fix/deterministic-solve-order`, because this PR's
body names that branch as the dependency. I did not measure `main`. Having now
built `main` (`8e9dda2`) separately:

| build | ODE, 16 threads | `linCmt()`, 16 threads |
| --- | --- | --- |
| `main` 8e9dda2 | 1 distinct objf over 6 fits | 1 distinct objf over 6 fits |
| + `5ec4d88f` | **6** distinct (`6504.59 .. 20445.16`) | **6** distinct (`7797.63 .. 2.3e36`) |

So #1020 does **not** reproduce on rxode2 `main` in this configuration, and my
"#1020 is not closed" was an artifact of the baseline I chose. What the failures
actually show is that `5ec4d88f` -- the unmerged `fix/deterministic-solve-order`
commit -- introduces the corruption. Details, reproducer and the ruling-out of
nlmixr2/rxode2#1294 are in nlmixr2/rxode2#1303.

**The test file on this branch passes 12/12 against rxode2 `main`.** The PR is
green and mergeable as it stands.

The rest of the previous comment holds, and the review fixes are what make this
result meaningful rather than lucky:

* the thread guard now actually lets the test run (it had been skipping itself on
  every machine with fewer than 12 *configured* threads, which on a 22-thread box
  is the default),
* the subjects are ragged, so `ordId` is a real permutation rather than the
  identity -- which is exactly what separates `main` from `5ec4d88f` here, and
  what the original equal-sized data could not have detected either way,
* `NEWS.md` and the header no longer give different root causes, and
* the file is in weekly slow batch 5 rather than the essential push/PR subset.

One change to the PR body is still needed: drop the `fix/ind-solve-subject-id`
dependency. nlmixr2/rxode2#1292 and #1293 are merged, and the `sortIds()` ordering
commit is not something this test should wait for -- it is what breaks it.

mattfidler:
The note in this PR's description -- "the unmerged `fix/deterministic-solve-order`
branch must **not** be merged under this test (nlmixr2/rxode2#1303)" -- is stale.

`test-focei-reproducible.R` passes `FAIL 0 | PASS 12` against rxode2 `main`
(`020cf90bf`) with `5ec4d88f` cherry-picked, and also against the
`8e9dda2` + `5ec4d88f` build that nlmixr2/rxode2#1303 reported as failing.

Evidence is in nlmixr2/rxode2#1303. The short version: the reported baseline is
clean now too, and since that rxode2 tree is unchanged, the fix came from this
side rather than from rxode2 -- with rxode2 pinned at `8e9dda2` the reproducer's
objf moved from `-471.9262` to `-470.6069` between the report and today.

PR #1037, opened 2026-09-04
build: track header dependencies so an object cannot go stale

billdenney:
R's build rules make an object depend on its own source alone, so make reuses an object whenever only a header it includes has changed.  Nothing warns; the object simply keeps whatever layout that header had when it was last compiled.

That crosses package boundaries.  A LinkingTo package's headers are just another include path, so when a struct there gains or reorders a field, the objects here that were compiled against the previous layout are kept and linked alongside freshly compiled ones.  The result is a shared library whose halves disagree about member offsets -- a corrupt binary rather than a compile error, surfacing as a segfault far from the cause.

Compile with -MMD -MP so each object records its headers in a .d file beside it, and read those files back.  -MP emits a phony target per header so renaming or deleting one does not break the next build.

.DEFAULT_GOAL := all is required, not cosmetic: Makevars is read before shlib.mk defines "all", so without it the first target of the first included .d file becomes make's default goal and the shared library is never linked even though its objects rebuild.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -121,3 +121,4 @@ BinaryFiles$
 # The fixtures CRAN's own checks read are the CHECKED-IN csv references
 # (npag/pmetrics_reference.csv, npag/ipm/burke-golden.csv) and are kept.
 ^tests/testthat/fixtures/.*\.rds$
+^src/.*[.]d$
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,4 @@ inst/tools/*.Rout
 src/Makevars.win
 .habit-hooks/
 compile_commands.json
+/src/*.d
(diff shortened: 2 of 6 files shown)


mattfidler:
This will not pass CRAN. They are explicit about the state of files and the requirements of Makevars, so these are not something that can currently be run as is.

billdenney:
Is there another way to track header dependencies that can work? Or, did Claude get this wrong?

mattfidler:
As far as I know, there isn't and it isn't done across the R ecosystem.  We are ahead of the game just using function linkages instead of abi linkages.  I have been repeatedly asked to remove these sorts of compiler antics.

The rxode2 has more, and could be gated off separately, though.

The biggest issue is platform X doesn't support it, or this gives reviewer Y something that they don't want.  Fix now or else.

mattfidler:
This also makes it so we can't submit to cran without breaking all the reverse dependencies.

Issue #1039, opened 2026-09-04
FOCEi: gVid sizing intermittently reads invalid per-subject event counts (build-dependent)


mattfidler:
## Summary

On `main`, every FOCEi fit with more than one subject fails immediately with

```
focei: dataset too large -- the gVid allocation overflows
```

`theo_sd` (12 subjects, 132 records) reproduces it. The dataset is not too
large -- `getRxNobsSqAndMix()` reads the per-subject event counts before they
are populated, so it sums squares of uninitialized memory.

## Reproduce

```r
library(nlmixr2est)
one.compartment <- function() {
  ini({tka <- 0.45; tcl <- 1; tv <- 3.45
       eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1; add.sd <- 0.7})
  model({ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
         d/dt(depot) = -ka * depot
         d/dt(center) = ka * depot - cl / v * center
         cp = center / v
         cp ~ add(add.sd)})
}
nlmixr2(one.compartment, nlmixr2data::theo_sd, est = "focei",
        control = list(print = 0))
#> Error: Could not fit data
#>   focei: dataset too large -- the gVid allocation overflows
```

Both `linCmt()` and ODE models fail, at `.foceiFitInternal()`.

## Root cause

`getRxNobsSqAndMix()` (`src/inner.cpp`, introduced in 2b1589b5f) loops over
subjects computing

```c
size_t nobs = (size_t)(getIndNallTimes(ind) - getIndNdoses(ind) - getIndNevid2(ind));
tot = foceiSzAdd(tot, foceiSzMul(nobs, nobs, "gVid"), "gVid");
```

Instrumenting that loop with an `REprintf` of the three accessors, on `theo_sd`:

```
DIAG i=11 nsub=12 nallTimes=0          ndoses=0          nevid2=0
DIAG i=10 nsub=12 nallTimes=-978333256 ndoses=-978333616 nevid2=0
DIAG i=9  nsub=12 nallTimes=0          ndoses=0          nevid2=0
DIAG i=8  nsub=12 nallTimes=0          ndoses=0          nevid2=0
DIAG i=7  nsub=12 nallTimes=0          ndoses=0          nevid2=-1
DIAG i=6  nsub=12 nallTimes=0          ndoses=0          nevid2=0
[...]

PR #1041, opened 2026-09-04
focei: validate rxode2's per-subject event counts before sizing from them

mattfidler:
Refs #1039. Note this PR does **not** by itself fix that issue -- nlmixr2/rxode2#1357 does. This is what keeps the same class of breakage from reaching the allocator quietly.

## Background

#1039 reproduces as `focei: dataset too large`, an `R_Calloc` failure for an
absurd size, or a segfault, on every multi-subject FOCEi fit. The cause turned
out to be entirely on the rxode2 side: an installed `rxode2.so` linked from
object files compiled against two different `rx_solving_options_ind` layouts.
Under gdb, in one loaded library, `rx2api.c`'s translation unit reports
`sizeof == 976` while `rxData.cpp`/`par_solve.cpp` report `1664`. So the subject
array was laid out at a 1664 stride and `getSolvingOptionsInd()` returned
`base + 976*i`. Subject 0 sits at the start of the array either way, which is
why a single-subject dataset fits and every multi-subject one does not.

nlmixr2est was already clean here: it goes through function pointers only, never
dereferences `rx_solving_options_ind`, and never does subject-array arithmetic.
nlmixr2/rxode2#1357 removes the layout from the ABI accessor.

## The change

What was missing on this side is that a mis-read count reached the allocator at
all. Every per-subject block in the FOCEi setup -- `gVid`, `ga`/`gc`, `gB`,
`gcH*`, `llikObsFull` -- is sized and strided from rxode2's per-subject event
counts, and nothing re-derives them.

`foceiCheckIndCounts()` runs before anything is sized from them, and refuses
counts that cannot be right: a negative one, or a dose or `evid=2` count larger
than that subject's own record count. On a mismatched build the reproduction
now stops with

```
focei: rxode2 reports an impossible event layout for subject 6 (records:
101650104, doses: 435033104, evid=2: 435031952); reinstall rxode2 and
nlmixr2est from source
```

instead of a bogus allocation size or a segfault.

### Only rules that hold across rxode2 releases

Every rule is about **one subject's own three numbers**, deliberately.
nlmixr2est has to run against rxode2 releases it is not built alongside -- CRAN
pairs the current nlmixr2est with whatever rxode2 is already there -- and a
rule resting on an rxode2-wide total would turn a change in how rxode2 accounts
for records into a failed fit for everyone on that release, and a failed
submission with it. An earlier revision of this PR did exactly that (each
subject's records had to fit inside `rx->nall`, and the counts had to sum to
it); those two rules are gone.

The cost is stated rather than hidden: counts that are individually plausible
but collectively wrong -- a subject that comes back with none of its records --
are accepted here, and that is the shape that *under*-sizes `gVid`. What keeps
it from arising is nlmixr2/rxode2#1357, not this check. The sum invariant is
asserted in the test instead, on a known dataset, where a change in rxode2's
accounting surfaces as a test failure rather than as a broken fit.

The narrowed rules still catch the build that prompted #1039: three runs
against it, all stopping at subject 6.

The `// nocov` on the overflow branches stays. Those need a dataset with more
than 2^32 records on a 32-bit `size_t`; no test can reach them, and the case
that reaches users in practice is now caught earlier with a better message.

## Tests

`tests/testthat/test-focei-ind-counts-1039.R`. Both checks need more than one
subject with *unequal* record counts: subject 0 reads correctly under any
stride, and equal counts let a neighbouring subject stand in for the right one.

- `foceiIndEventCounts_()` exposes the counts, so the test asserts the layout
  against the dataset directly rather than inferring it from a fit's output.
- `foceiCheckIndCounts_()` drives the rules on counts supplied from R, which is
  the only way to reach the rejection paths -- they need a build whose rxode2
  and nlmixr2est disagree on the solve layout. It also pins what is
  deliberately *accepted*, so the trade-off above cannot be un-made by accident.
- A multi-subject `theo_sd` FOCEi fit, asserting `llikObs` comes back with its
  NA entries on the dose rows of the event table -- that block's per-subject
  offsets come straight from these counts.

The two assertions that read the counts off a real solve carry
`skip_on_cran()`, for the same reason the rules dropped the total: they are
rxode2's record accounting, and CRAN pairs this package with whatever rxode2 is
already installed. A release that accounts for records differently has to show
up as a test failure here, never as a check failure on CRAN. (`nmTest()`
already returns early off `NOT_CRAN`; the skips say per test which ones depend
on the installed rxode2.) The rules test carries no skip -- it never touches a
solve, which is exactly the property the rules were narrowed to have.

Ran clean across all 79 FOCEi-family test files. The single failure there is
the known `censInformation` process-global carryover, which passes when that
file is run on its own.

## Review

Three rounds of independent review (Antigravity, Gemini). Fixed from it:

- The sum check was gated on `getRxNsim(rx) == 1`, a value read through the same
  accessors the check exists to distrust -- garbage there skipped the check
  entirely. The gate was also unnecessary: `rx->nall` counts one replicate of
  the event data and subjects `[0, nsub)` are that replicate whatever `rx->nsim`
  asks for.
- Nothing covered the rejection paths; `foceiCheckIndCounts_()` was added for
  that.
- `foceiIndEventCounts_()` now guards its `getRxSolve_()` result the way
  `foceiIndLik_` does.

One finding rejected: a claimed NULL dereference when `rx->subjects` is unset.
`getSolvingOptionsInd()` bounds-checks and raises a clean R error there; checked
by calling the accessor after `rxClean()`.

One considered and dropped: cross-checking `getRxNsub(rx)` against nlmixr2est's
own count of the dataset's IDs. rxode2's `rxHomGroups` datasets carry one row
block per group and expand to more subjects than an ID-transition count sees, so
the cross-check would falsely reject them. The scenario it defends against needs
`rx_solve`'s own layout corrupted, at which point nothing about the solve is
readable anyway.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -110,8 +110,9 @@
   layouts those counts are read from the wrong bytes, and the setup sized
   megabytes of storage from garbage -- reported as "dataset too large", as an
   `R_Calloc` failure, or as a segfault, depending on what the mis-read bytes
-  held.  The counts have to add back up to the dataset's record count, and the
-  fit now stops with that as the reason instead (#1039).
+  held.  A negative count, or a dose or `evid=2` count larger than the
+  subject's own record count, now stops the fit with that as the reason
+  instead (#1039).
 
 - With two or more occasion parameters on one level, `fit$iov$<level>` had
   `NA` for every occasion (and the fit warned "NAs introduced by
--- a/NEWS.md
+++ b/NEWS.md
@@ -104,6 +104,14 @@
 
 ## Bug fixes
 
+- `focei` now checks rxode2's per-subject event counts before it sizes the
+  per-subject blocks it strides with them (`gVid`, `ga`/`gc`, `gB`, `gcH*`,
+  `llikObsFull`).  When rxode2 and nlmixr2est are built against different solve
+  layouts those counts are read from the wrong bytes, and the setup sized
+  megabytes of storage from garbage -- reported as "dataset too large", as an
+  `R_Calloc` failure, or as a segfault, depending on what the mis-read bytes
+  held.  The counts have to add back up to the dataset's record count, and the
+  fit now stops with that as the reason instead (#1039).
 
 - With two or more occasion parameters on one level, `fit$iov$<level>` had
   `NA` for every occasion (and the fit warned "NAs introduced by
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -75,6 +75,10 @@ impQrPoints_ <- function(isample, neta, shift) {
     .Call(`_nlmixr2est_impQrPoints_`, isample, neta, shift)
 }
 
+foceiCheckIndCounts_ <- function(counts, nall) {
+    invisible(.Call(`_nlmixr2est_foceiCheckIndCounts_`, counts, nall))
+}
+
 foceiIndEventCounts_ <- function() {
     .Call(`_nlmixr2est_foceiIndEventCounts_`)
 }
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -75,8 +75,8 @@ impQrPoints_ <- function(isample, neta, shift) {
     .Call(`_nlmixr2est_impQrPoints_`, isample, neta, shift)
 }
 
-foceiCheckIndCounts_ <- function(counts, nall) {
-    invisible(.Call(`_nlmixr2est_foceiCheckIndCounts_`, counts, nall))
+foceiCheckIndCounts_ <- function(counts) {
+    invisible(.Call(`_nlmixr2est_foceiCheckIndCounts_`, counts))
 }
 
 foceiIndEventCounts_ <- function() {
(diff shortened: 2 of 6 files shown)


mattfidler:
This also gives the same issue as #1037 it will break all reverse dependences.

mattfidler:
Narrowed the guard after the point that it must survive an rxode2 it is not
built alongside (CRAN pairs the current nlmixr2est with whatever rxode2 is
already there).

Two of the four rules rested on `rx->nall` -- each subject's records had to fit
inside it, and the per-subject counts had to sum to it exactly. Those are
rxode2's own record-accounting conventions, not impossibilities, and nlmixr2est
does not control them across releases. A release that counted differently would
have failed **every** fit on it, and every submission with it. Both are gone in
f9a6742.

What is left is per-subject only: a negative count, or a dose or `evid=2` count
larger than that subject's own record count. No working rxode2 of any version
reports those.

The cost is real and is now stated in the code comment and pinned in the test: a
subject that comes back with none of its records is individually plausible and
is accepted, and that is exactly the shape that under-sizes `gVid`. What keeps
it from arising is nlmixr2/rxode2#1357, not this check.

The narrowed rules still catch the build that prompted #1039 -- three runs
against it, all stopping at subject 6 with doses + `evid=2` records far past
that subject's record count. The dropped sum invariant is asserted in the test
instead, on a known dataset, so a change in rxode2's accounting shows up as a
test failure rather than as a broken fit.

Issue #1044, opened 2026-09-05
innerOpt="trust": inner solves fail at some parameter sets, and nothing in the fit says so


mattfidler:
Split out of #1040, where the `mceta` half is fixed by #1042. This is the other half of the re-measurement table there, and it is not an `mceta` problem.

## What the numbers show

Same model (300 subjects, fixed-omega inverse-CDF etas), `maxOuterIterations = 0`, `maxInnerIterations = 5000`, `covMethod = ""`, `innerOpt = "trust"`:

| parameters | `mceta=-2` | `mceta=0` | `mceta=10` | `mceta=100` |
|---|---|---|---|---|
| set A | -8468.6917 | -8468.6917 | -8584.6519 | -8676.7767 |
| set B | **3007.2819** | **3007.2819** | **2750.1103** | -8619.5409 |

Set A is the intended shape: monotone, each increase in `mceta` improves it.

Set B is not. `-2`/`0`/`10` sit about 11,600 objective units above `mceta = 100`, and on the *positive* side of zero. That is not "suboptimal starting points" -- those inner solves are failing at that parameter set, and only enough random restarts get away from it.

## Why this is not about mceta

At `mceta = -2` and `mceta = 0` there is exactly one starting point: no draws, no eta=0 floor pass, no candidate ranking. None of the machinery #1042 changes is reachable, and the two agree bit-for-bit (3007.2819 both, as they should at `maxOuterIterations = 0` where the etas start at zero either way). So set B is a `trust` inner-solve failure that `mceta = 100` masks by restarting enough times.

## The part that makes it hard to see

There is no per-fit diagnostic that would say the solves failed.

`op_focei.nTrustInner` (`src/inner.cpp:561`, incremented at `:4239`) counts `trust_solve_c()` **calls**, not solves that came back non-converged or with `tres.error < 0`. It is not on the fit environment either -- the only access is the `.nTrustInner()` Rcpp export, used by `test-focei-trust-inner.R`. So a fit whose trust solves are all failing is indistinguishable, from the fit object, from one where they all converged.

The arm already distinguishes the cases it would need to count (`src/inner.cpp:4241-4280`):

- `tres.error < 0` -- hard error; `f` is forced to `+Inf` so the shared `restoreBest()` guard fires.
- `tres.error >= 0` but `tres.converged` false, or `conv` downgraded because the Newton step `pushDist > pushTol`.
[...]

Issue #1040, opened 2026-09-04
mceta=10 gives the same objective as mceta=-2, and both are far worse than mceta=0 (fixed-omega etas)


mattfidler:
On a model whose random effects have a **fixed** omega, the FOCEi objective at **fixed parameters** moves by thousands of units depending on `mceta`, and `mceta = 10` returns a value bit-identical to `mceta = -2` rather than improving on `mceta = 0`.

`foceiControl()` documents `mceta = n` (`n > 0`) as

> the last eta, eta=0, and n-1 etas sampled from omega are each evaluated and the best (by inner objective) is used

Since `eta = 0` is among the candidates, `mceta = n` should never be worse than `mceta = 0`.  It is, by a lot.

## Measured

Two-compartment model, 300 subjects, `maxOuterIterations = 0` (pure evaluation), `maxInnerIterations = 5000`, `covMethod = ""`.  Same data, same parameters, only `mceta` changes:

| parameters | `mceta=-2` | `mceta=0` | `mceta=10` | `mceta=100` |
|---|---|---|---|---|
| set A | -1756.7425 | -8413.3392 | **-1756.7425** | -8579.1984 |
| set B | -2708.5246 | -8036.7935 | **-2708.5246** | -8492.6439 |

Two things stand out:

1. `mceta = 10` reproduces `mceta = -2` **exactly**, to the last digit, at both parameter sets.  It looks like the extra candidates are not being used -- as if a stale "last eta" wins regardless.
2. Both are ~5300-6700 objective units worse than simply starting every inner problem at `eta = 0`.  `mceta = 100` is better still, so more starts *do* help once there are enough of them; 10 does nothing at all.

## Model shape that triggers it

The model has a **fixed** omega on some etas -- unit variance, `FIX` -- alongside an ordinary estimated block:

```
rxz.eta.cl ~ fix(1)
rxz.eta.v1 ~ fix(1)
eta.q + eta.v2 ~ c(0.0305,
                   0.0107, 0.0285)
```

and the fixed etas enter through a strongly non-linear transform (an inverse CDF), so the inner problem has real curvature.  I have not checked whether the fixed omega, the non-linear transform, or both are needed to reproduce it.

## Why it matters

[...]

PR #1046, opened 2026-09-05
focei: report whether the inner solves converged, and stop selecting failed candidates (#1044)

mattfidler:
Closes #1044.

#1044 asked for one thing first: a per-fit diagnostic that says whether the
`innerOpt="trust"` inner solves converged, because `op_focei.nTrustInner`
counted `trust_solve_c()` **calls** and a fit whose inner solves were all
failing looked identical, from the fit object, to one where they all
converged. That diagnostic is here, and building it turned up four real
defects in the candidate handling and the retry cascade.

## What the diagnostic says

`fit$env$nTrustInner` now breaks the attempts down by outcome, and
`fit$env$nInnerRerank` gains two entries:

```
nTrustInner: calls, error, notConverged, solverFail, newtonGate,
             warmRetry, radiusRetry, nudge, failed
nInnerRerank: ranked, flipped, noGood, dropped
```

`notConverged == error + solverFail + newtonGate` by construction (pinned by
a test). On the #1040 model at 300 subjects, `maxOuterIterations = 0`, walking
the parameter set away from the truth (`d` = how far each theta is displaced):

| case | failed / 300 | notConverged | solverFail | newtonGate |
|---|---|---|---|---|
| d=1 | 1 | 24 | 0 | 24 |
| d=2 | 12 | 298 | 0 | 298 |
| d=3 | 55 | 482 | 0 | 482 |
| d=4 | 195 | 1450 | 0 | 1450 |

That is the answer to the issue's question, and it is not the one the issue
guessed at: `trust_solve_c()` never reports failure on this model
(`solverFail = 0` in every run measured). Every retry is driven by our own
Newton-decrement gate, and at a poor parameter set two thirds of the subjects
can end with the whole cascade spent.

## Bad candidates are no longer selectable

The restart candidates the marginal re-rank chooses from carried no record of
whether the attempt that produced them succeeded, so a failed attempt's eta
could win -- its Laplace `log|H|` is measured at a point the inner objective
never descended to. `mceta >= 1` is where a mixed candidate set is common,
since the extra starting points are what produce one. `keepCand()` now takes
that verdict; the selection uses a failed attempt only when nothing else
survived, and `nInnerRerank[["dropped"]]` counts when it fires.

**"Failed" is the optimizer's own verdict, not the Newton gate.** A first cut
used the gate and it was measurably wrong -- the gate is deliberately eager
because its job is to trigger retries. Measured against n1qn1 on the same fit:

| mceta | n1qn1 | trust, gate = failure | trust, optimizer's own verdict |
|---|---|---|---|
| 0 | 34004.40 | 41821.79 | 35097.24 |
| 10 | 33548.35 | 38192.62 | 34248.75 |
| 100 | 33525.64 | 37498.38 | 33799.14 |

## Fewer solves fail, and fewer solves run

When the gate withdraws convergence and the Newton step still **fits** inside
the radius, `trust_solve_c()` stopped on its own step-size criterion, measured
against the previous iterate rather than the model's remaining decrease. The
cascade went straight to the eta nudges, which throw that point away and
restart from a fill. It now re-solves once in place first:

| case | failed | calls | objective |
|---|---|---|---|
| d=1 mceta=0 | 1 -> 1 | 318 -> 323 | 3032.9855 -> 3032.9864 |
| d=2 mceta=0 | 13 -> 12 | 527 -> 586 | 35097.2404 -> 35097.2144 |
| d=3 mceta=0 | 96 -> 55 | 814 -> 727 | 434969.7132 -> 434742.5472 |
| d=3 mceta=10 | 185 -> 92 | 1580 -> 1353 | 229954.0707 -> 229954.0559 |
| d=4 mceta=10 | 351 -> 278 | 2566 -> 2398 | 2010047.2519 -> 2010041.0607 |

Converting a gate withdrawal early is cheaper than the four nudges it
replaces, so the total number of solves falls with the failures. A parameter
set where nothing trips the gate is unchanged (300/300/550/596 calls before
and after).

## Other defects fixed along the way

- **A failed pass discarded an earlier pass's candidate.** `haveBest` is
  reset at the top of every `mceta` starting-point pass, but both the trust
  arm and the n1qn1 restart cascade read it as "this subject produced
  nothing" and returned a failed inner solve. Reachable with `mceta >= 1`
  whenever a later pass fails outright. Both now take the same three-way exit
  the first call in a pass already did.
- **A NaN trust result was treated as a success.** Only `trust_solve_c()`'s
  error code forced `f` to `+Inf`; a call that returned normally with a
  NaN/Inf objective kept the solver's converged flag, so an unusable attempt
  ended the retry cascade.
- **Exception safety.** The result arrays were freed only on the lambda's
  normal exit while the Armadillo work between can throw `bad_alloc` (which
  the branch catch swallows); an RAII guard armed before the solve takes it.
  `keepCand()` does every allocation before any push, so the three candidate
  vectors cannot end up ragged and be indexed against each other.

## Review

Seven rounds of `agy` (Gemini) review; the last is clean. Its findings drove
the pass-local, NaN, exception-safety and per-stage-coverage commits above.
One finding was rejected on inspection (it read `haveBest` as global; it is
reset per pass at `src/inner.cpp:4130`). One is acknowledged and **not**
fixed here: a `std::bad_alloc` out of the n1qn1/lbfgsb3C arms or the re-rank
snapshot escapes the OpenMP region, since only the trust arm carries a catch.
That is pre-existing and a correct fix changes the whole function's error
contract, so it does not belong in this PR.

## Tests

`tests/testthat/test-focei-trust-inner.R` gains coverage for the counter
breakdown, the candidate rule (`dropped`/`noGood`), each retry stage
(`newtonGate`, `warmRetry`, `radiusRetry`, `nudge`, `failed`) on a fit that
reaches it, the hard-error outcome, and the `mceta >= 1` mixed candidate set.
Every count is asserted `> 0` rather than pinned to a value, since the exact
numbers move with rxode2's solver.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -104,6 +104,32 @@
 
 ## Bug fixes
 
+- A `focei`-family fit now reports whether its inner solves actually
+  converged.  `fit$env$nTrustInner` breaks the `innerOpt="trust"` per-subject
+  Newton solves down by outcome (`calls`, `error`, `notConverged`,
+  `solverFail`, `newtonGate`, `warmRetry`, `radiusRetry`, `nudge`, `failed`)
+  and `fit$env$nInnerRerank` gains `noGood` and `dropped`.  Only the call count
+  existed before, and only through an internal accessor, so a fit whose inner
+  solves were all failing was indistinguishable from one where they all
+  converged.
+
+- The inner restart candidates are no longer chosen from without regard to
+  whether the attempt that produced them succeeded.  A failed attempt's eta
+  could win the marginal re-rank over a converged one -- its Laplace
+  `log|H|` term is measured at a point the inner objective never descended to
+  -- which `mceta >= 1` makes more likely, since the extra starting points are
+  what produce a mixed candidate set.  A failed attempt is now used only when
+  no succeeded one is available.
+
+- `innerOpt="trust"` re-solves in place before falling back to its eta-nudge
+  restarts.  When the inner Newton step still fits inside the current trust
+  radius, the solve stopped on its own step-size criterion rather than on the
+  model's remaining decrease, and the cascade threw that point away to restart
+  from a nudge fill.  On a 300-subject fit swept away from the true parameters,
+  subjects whose whole cascade is exhausted drop by up to half and the total
+  number of inner solves drops with them; a fit where nothing trips the check
+  is unchanged.
+
 - A fit no longer reports the PREVIOUS fit's censoring.  `$censInformation`
   is built from a process-global flag recording which censoring methods
   (M2/M3/M4) a fit used, and that flag was cleared only after being read at
--- a/NEWS.md
+++ b/NEWS.md
@@ -121,6 +121,12 @@
   what produce a mixed candidate set.  A failed attempt is now used only when
   no succeeded one is available.
 
+- `innerOpt="trust"` with `mceta >= 1` no longer fails a subject whose eta=0
+  floor pass produced nothing.  The "did this pass find anything" check is
+  per-pass, but the trust arm treated it as "did this subject find anything"
+  and returned a failed inner solve, discarding a converged candidate an
+  earlier starting point had already produced.
+
 - `innerOpt="trust"` re-solves in place before falling back to its eta-nudge
   restarts.  When the inner Newton step still fits inside the current trust
   radius, the solve stopped on its own step-size criterion rather than on the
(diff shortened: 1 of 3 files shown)


mattfidler:
Full essential test subset (199 files, the same set push/PR CI runs): **5382 passed, 0 failed, 0 errors**, 1 skip (`{qs}` unavailable) and 4 pre-existing warnings unrelated to this change.

PR #1049, opened 2026-09-07
Add analytical outer Hessians for fast FOCE, FOCE+, FOCEI and AGQ fits

hiddevandebeek:
This change makes analytical curvature available to existing inner and outer optimizers through the C++ sensitivity pool. Outer curvature supports Gaussian FOCE, FOCE+, FOCEI and AGQ, including optimized EBE dependence and M2/M3/M4 censoring with `censOption="gauss"`. Optional full conditional inner curvature supports fast Gaussian FOCEI while preserving its marginal objective.

- Reuses the existing C++ observed-information contractions. Third-order terms come from ETA differences of second-order sensitivities. AGQ includes moving-node and quadrature-weight derivatives.
- Adds one native callback and the matching `control$hessian(par, relStep=1e-3)` interface. It uses optimizer coordinates and objective scaling, retains negative curvature, and restores sensitivity-probe state.
- Existing `outerOpt="nlminb"` consumes outer curvature automatically in supported `fast=TRUE` fits. If it is unavailable, nlminb restarts with gradients only. No new outer optimizer is introduced.
- `innerHessian="conditional"` assembles value, gradient and full conditional curvature jointly from one ETA-only second-order sensitivity solve, including censored likelihood terms. Inner trust uses the signed matrix at each trial; n1qn1 uses the existing positive-definite seed preparation with `warm="calc"`. The default remains `innerHessian="focei"`. The FOCEI determinant and EBE warm predictor are unchanged; no extra warm-EBE correction or cached curvature is included.
- Two prerequisite corrections are isolated in separate commits: objective scaling of fast gradients and FOCE's sensitivity-buffer row stride.

FOCE+ now refines its truncated live-variance ETA score root before evaluating the marginal objective, aligning the objective with its analytical derivatives. This can change FOCE+ estimates. Censored AGQ evaluates the censored likelihood at its quadrature nodes; fast censored Laplace gradients also use consistent node placement.

Censored `censOption="laplace"` outer curvature, priors, clipped AGQ and estimated transformations are outside the supported outer-Hessian scope. Failed inner likelihoods and an active variance floor also return unavailable outer curvature. The outer censoring scope matches the existing analytical-SE kernel; censored Laplace uses gradient-only optimizer fallback. The optional conditional inner Hessian requires analytical ETA sensitivities and is not enabled for generalized endpoints or unsupported sensitivity carries.

Validation: Windows R 4.5.3 build passed. The initial implementation passed 284 focused assertions covering independent objective checks, scaling, correlated Omega, AGQ with three/five nodes per ETA, native ABI/state handling, automatic nlminb/fallback, and fast-gradient regressions. The extension passed the expanded family (48), automatic nlminb (27), FOCE+ regression (13), and independent analytical-SE reference (13) assertions. A focused `devtools::load_all(helpers=FALSE)` run also checked M2/M3/M4 curvature and censored Laplace gradient/fallback. Multi-fit tests are in the weekly batches. Full R CMD check was not run.

Twenty complete fitting runs compared built-in analytical-Hessian nlminb with gradient-only nlminb across FOCE/FOCEI/AGQ, two starts, two small models and a 30-subject two-compartment ODE model. All ten analytical fits converged with actual Hessian calls and no fallback. All pairs reached matching objectives; one gradient-only fit reported non-convergence despite reaching the target. Maximum objective difference among successful pairs was 6.8e-7. Full-fit time included covariance and tables after compilation was warmed: the larger ODE fits fell from 36.6 to 31.8 seconds and 52.2 to 41.9 seconds. Small-model gains were modest, and AGQ timing was mixed. This does not establish a general robustness or speed guarantee.

The joint-versus-separate full-curvature comparison used two small ODE models, two starts, trust/n1qn1, and three timed repetitions after full-fit warm-up. All 24 fits converged for each version; maximum paired objective difference was 9.9e-8. Joint evaluation reduced median CPU time in all eight cases, with median case reductions of 29% for trust and 31% for n1qn1. These measurements include second-order solve costs and exclude compilation, covariance and tables. They do not establish a general advantage over ordinary FOCEI curvature, which remains the default. Independent EBE references cover uncensored/M2/M3/M4 data with both inner solvers.

Benchmark scripts, figures and warm-correction experiments are outside this PR.

Focused validation with `devtools::load_all(helpers=FALSE)` passed the native-API (319), outer-censoring (152), and outer-Hessian API (14) assertions. The final joint implementation passed 43 inner-curvature assertions, 48 outer-family assertions, and 13 FOCE+ regression assertions. The native table test includes the appended outer-Hessian callback. Model solves, contractions, and AGQ node calculations are separated into named steps.


hiddevandebeek:
Independent agent review of #1049 at e65009db7

Requested focus: compatibility across inner solvers.

[P2] Gate all delay models, including default-history models
[R/focei.R:1482](https://github.com/nlmixr2/nlmixr2est/blob/e65009db7/R/focei.R#L1482)
The supported-model guard tests only length(.s$..pastLines), which is empty for delay() models that use the default history. These models therefore advertise handle$innerHessian=TRUE despite the explicit DDE exclusion in foceiLikInner's contract. Reproduced against the installed branch with y(0)=2; d/dt(y)=-k*exp(etaK)*delay(y,1), no past() declaration: loading succeeds, capability is TRUE, and inner evaluations return nBad=0. This bypasses the intended unsupported-model boundary; testing only explicit past() does not enforce it. Check rxode2::rxModelVars(.s)$flags[['hasDelay']] directly and add a default-history rejection regression. At atol/rtol=1e-11 its derivatives matched finite differences, so this is a capability-contract finding, not evidence of incorrect DDE curvature in that example.

Integration prerequisite (not a current batch-path correctness failure)
[src/inner.cpp:3747](https://github.com/nlmixr2/nlmixr2est/blob/e65009db7/src/inner.cpp#L3747), [src/inner.h:42](https://github.com/nlmixr2/nlmixr2est/blob/e65009db7/src/inner.h#L42)
foceiInnerDeriv() is declared as a reusable per-subject callback but depends on batch-owned solver context. Its only current caller correctly installs OdeSwapEsBatch(odeSlotHess2), resets tolerance/error state, and enters the rxode2 thread slot. A future trust callback inside the ordinary parallel inner loop cannot call it without supplying equivalent context: OdeSwapScope changes the per-subject equation count but does not install the global event-sensitivity shape. Document these prerequisites or add a facade that installs/restores Hess2 context for serial calls and rejects mismatched parallel calls before mutating state. Compare exact slot identity, not shared model role. The ordinary first-order gradient calls must retain their own shape, and global shape changes must stay outside OpenMP regions. Preserve explicit failure/fallback behavior for unsupported models and solve failures.

Validation and scope
Read the complete branch diff plus ordinary likInner0/calcEtaHessian, setup paths, batch/thread helpers, model generation, and odeSwap context contracts. Gaussian objective/gradient/Hessian formulas are consistent, including variance curvature and an indefinite Hessian away from the mode. The new table entry is additive, registration arity matches, innerHessian defaults FALSE, and the new provider is not substituted into calcEtaHessian/determinants or existing optimizer callbacks. The batch uses per-subject buffers, caps cores to the pool, and installs its model context outside OpenMP. Separate local validation tested n1qn1/BFGS and warm calc/save fit parity. No additional demonstrated production correctness defect found. The prospective trust integration requires explicit adapter work; #1049 alone does not make existing optimizers consume the new derivatives.

Regression coverage recommendation
[tests/testthat/test-focei-inner-analytic.R:140](https://github.com/nlmixr2/nlmixr2est/blob/e65009db7/tests/testthat/test-focei-inner-analytic.R#L140)
The committed fit-parity case exercises only the default optimizer with an algebraic model. Make the successful independent ODE parity check a maintained regression: n1qn1 and BFGS, each with warm='calc' and warm='save', innerHessian FALSE/TRUE. Compare both EBEs and objective, and retain an interleaving test between ordinary conditional gradients and the new Hessian provider. This is coverage hardening, not a demonstrated solver regression.

hiddevandebeek:
Review addressed in c9be86f71:

- Rejected all delay models using the model's `hasDelay` flag, with a regression for default history.
- Guarded the per-subject evaluator with exact Hessian-model context validation. Missing context returns `-5` before any solver-state mutation; caller setup and failure handling are documented. Global event-sensitivity state is never switched inside a solver callback.
- Added maintained ODE fit-parity coverage for n1qn1/BFGS and warm `calc`/`save` in weekly batch 5. Essential derivative and interleaving tests remain in the core batch.

The independent reviewer re-read the fixes and found no remaining blocking issue. Clean rebuild and 73 focused assertions passed; CodeFactor now reports no issues.

Additional compatibility check: the reviewer built these changes together with the current trust implementation in #987 in an isolated worktree. Combining the branches required mechanical resolution of adjacent additions in three files, with no overlapping numerical changes. All four `hessianMethod` settings (`fd`, `bfgs`, `sr1`, `bofill`) preserved ETA estimates and objective values with `innerHessian` disabled/enabled, at tolerance 1e-9. Each fit reported actual trust-optimizer execution. The non-fd settings retain their documented Gaussian behavior.

This verifies coexistence with current solvers; adapters that make optimizers consume the new Hessian remain a separate change. GitHub CI is still running.

hiddevandebeek:
@mattfidler working on this a bit, expect to finish it tomorrow. also looking into ebe warm start now

mattfidler:
I think the warm ebe start is ok.  I may be more valuable to allow outer Hessian calculation in C/c++ to see if a trust outer opt will work for foceif

hiddevandebeek:
okay good

mattfidler:
Well, I guess the warm start for `foceif` may not be good... Not clear, needs some investigation.

hiddevandebeek:
this turned into a goosechase. doing it again properly

hiddevandebeek:
Fresh independent review of commit `83c8f02e1eb010e325302b4c331a4bcd4992c147` found no actionable findings within its documented Gaussian FOCE/FOCEI/unclipped-AGQ scope. The review checked mathematical assembly (including AGQ node motion and weighted curvature), native lifecycle/ABI, probe restoration, parameter/objective scaling, and nlminb fallback. This was a read-only review; it did not run additional fits.

FOCE+ and censoring coverage are being extended following the latest scope request. The review identified the necessary censoring work: reuse the existing censored score partials, use the actual censored likelihood and score/Hessian at AGQ nodes, and preserve the determinant convention selected by `censOption`. The current analytical SE path uses Gaussian determinant curvature for its analytic censoring support; its censored-Laplace curvature path falls back to finite differences. The extension must distinguish those paths explicitly.

hiddevandebeek:
@mattfidler let me know what you think

hiddevandebeek:
Robustness scout at current head `20a26430b`

I ran an isolated, read-only comparison after the implementation was complete. No scout code was committed or pushed.

Recovery normally required reported outer convergence, an independently calculated maximum EBE score below `0.001`, and an objective within `0.01` of the best valid fit for that likelihood. AGQ used convergence plus matched objective because its reported ETAs are posterior summaries rather than conditional modes. The matrices used ordinary convergence settings and 1000-iteration limits.

### Analytical outer Hessian

| Inner solver | Paired configurations | Analytical Hessian | Gradient-only nlminb | Median time ratio on matched recoveries |
| --- | ---: | ---: | ---: | ---: |
| n1qn1 | 58 | 54 | 52 | 0.92 |
| trust | 58 | 48 | 39 | 0.95 |
| BFGS | 18 | 16 | 15 | 0.93 |
| **Total** | **134** | **118** | **106** | -- |

The main recovery gain was with inner trust: 11 configurations recovered only with the analytical outer Hessian, while two recovered only gradient-only. Among paired recoveries, objectives differed by at most `2.1e-6`.

The comparisons covered FOCEI, FOCE, FOCE+, AGQ3, M2/M3/M4 censoring, censored FOCE+/AGQ, PK and Michaelis-Menten ODEs, three inner solvers, difficult starts, and a two-ETA model whose fitted Omega condition measure was about 4,726. The correlated-ETA model recovered all 24 configurations without a Hessian fallback.

### Full conditional inner Hessian

| Inner solver | Outer strategy | Paired configurations | FOCEI curvature | Conditional curvature | Median conditional/FOCEI time |
| --- | --- | ---: | ---: | ---: | ---: |
| n1qn1 | analytical | 28 | 25 | 25 | 1.10 |
| n1qn1 | gradient-only | 20 | 19 | 18 | 1.10 |
| trust | analytical | 28 | 18 | 25 | 1.20 |
| trust | gradient-only | 20 | 11 | 18 | 1.08 |

Conditional curvature gave trust seven additional recoveries under each outer strategy, with no paired case recovered only by standard curvature. In the M2/M4 grid, trust improved from 4/8 to 8/8 with analytical outer curvature; problematic M4 EBE scores fell from 0.0031-0.0052 to about `2e-6`.

It did not help n1qn1 recovery and cost about 10%. That is consistent with n1qn1 consuming the Hessian only as its initial seed before its own quasi-Newton updates.

### Direct checks

- 118 focused assertions passed: conditional inner curvature, outer family assembly, and nlminb integration.
- 60 outer-Hessian probes across five third-order steps (`1e-2` through `1e-4`) were finite and symmetric. Maximum relative error against central differences of the analytical gradient was 0.12%; changing the step had negligible effect.
- A nine-parameter model with estimated `f(depot)` and `lag(depot)`, including random effects in the dosing event, had 0.166% Hessian error and immediate objective-state drift of `1.3e-9`. The jump-sensitivity path is working; delay derivatives were not missing.
- One versus two persistent-pool solve threads produced exactly identical objectives, gradients, outer Hessians, and EBEs for ordinary PK and the lag/jump model, with both inner-curvature choices.

### Ideas that did not earn their complexity

- Retrying the outer Hessian at a nearby third-order step had no observed target: all probes were available and step-stable.
- Restarting gradient-only nlminb from the best point after an injected late Hessian failure changed no recovery result and had mixed evaluation cost.
- Positive-definite or conditioning gates targeted the wrong cases. Successful fits used indefinite Hessians and condition measures above 300, while some false-convergence fits saw only positive, better-conditioned curvature.
- Retrying the entire trust fit with conditional curvature matched always-conditional recovery but was slower in two of three matrices and cannot detect a converged local optimum.

My recommendation is to keep this PR's logic as it is: retain automatic analytical outer curvature for supported nlminb fits, keep `innerHessian="conditional"` available, and use it particularly with `innerOpt="trust"` when EBE robustness matters. Keep n1qn1 on the standard FOCEI curvature default. These are deliberately difficult small tests, so changing the trust default would deserve a separate larger-model evaluation.

hiddevandebeek:
Addressed two of the three review comments in 566c0696. The third is deliberately left
out of this PR -- reasoning below.

### Fixed here

**Conditional inner path vs. likelihood contributions.** Correct, and it was a wrong-answer
bug: the conditional assembly replaces `likInner0` and builds only the Gaussian density, so
a contributor's added LL and its eta derivatives were dropped. With `innerOpt="trust"` the
optimizer minimizes a different inner objective than the fit reports. Now refused in
`foceiSetup_`. Worth noting the two solvers differ in exposure -- `trust` consumes the value
and gradient at every trial, whereas `n1qn1` only seeds `warmZm` from it -- but the refusal
is unconditional and tested over both.

**`innerHessian="conditional"` + `warm="save"`.** Correct. Enforced now, and the check moved
below the `warm` normalization so it compares against the resolved `1L`/`0L`. Two
combinations that must stay legal (`trust` + `warm="save"`, `n1qn1` + `warm="calc"`) are
pinned by tests. `innerOpt="auto"` needs no separate guard: conditional curvature already
rejects `needOptimHess`, and auto resolves as `needOptimHess ? n1qn1 : trust`, so it
provably lands on trust here.

### Not fixed here: the outer-Hessian contributor gate

The observation is right -- `foceiHessianObjectiveSupport()` has no contributor check, so
`nlminb` can consume curvature for an objective the contraction does not represent. But this
is **not a regression in this PR**, and gating it here would be the wrong shape of fix.

The same hole already exists on `main` in the analytic outer **gradient**:
`nlmixrHasLikContrib()` is defined and exported but called zero times anywhere in `src/` or
`R/`, and the gradient kernel never touches the contributor registry. Filed as #1051, with a
measurement on a clean `main` build -- the analytic arm converges 0.88 objf units worse than
the FD arm and moves in the wrong direction. A wrong gradient relocates the optimum; a wrong
Hessian mostly costs iterations, since `nlminb` still tests optimality on the gradient. So
the gradient is the more damaging half, and it is the one that predates this PR.

Landing the Hessian gate alone would also invite a subtle regression later. The two need
**different** things:

* the **gradient** needs the contribution's first derivatives `dC/df`, `dC/dr`, to contract
  against the existing `df/dtheta` and `dR/dtheta` sensitivities;
* the **Hessian** needs those *and* second derivatives -- `log|H|` depends on eta, so the
  curvature carries `d(eta*)/d(theta)`, and the true eta\* is contributor-aware.

The Hessian condition is therefore strictly stronger. If #1051 is fixed first-order and a
single shared "contributions supported" predicate is relaxed, the outer Hessian silently
goes wrong again. Whatever lands should express the two as separate predicates, with the
Hessian requiring both.

Worth recording for whoever picks up #1051: the current ABI cannot carry either fix.
`inst/include/nlmixr2estLikContrib.h` lets a contributor write back only `llik` and
`dLL_deta`; `dLL_df`/`dLL_dr` are passed **in** read-only and never returned. Extending the
existing `obs` hook is also not safe -- it is not side-effect free (it may step a downstream
optimizer), so it must not be called a second time per observation just to harvest
derivatives. A separate additive derivative registry, mirroring `nlmixrRegisterEmLik`, avoids
that. A pure observer (`llik` never written) is already exact and should keep the fast path
either way.

hiddevandebeek:
@mattfidler leaving it to you now

mattfidler:
## Merged `main`, ran an independent review, and added two features

`main` is merged in (the branch was `CONFLICTING`; conflicts were in `NEWS.md`,
`R/focei.R`'s model-cache digest, and `test-focei-foce-plus.R`). All of the
branch's own tests pass on the merged tree, and a 61-file sweep of the
`focei`/`foce`/`agq`/`laplace`/`cov`/`broom` area is clean.

### Independent review (Antigravity, `gemini-3.1-pro-high`)

Seven rounds against the branch diff, each verified against the code and, where
the claim was numerical, by measurement. Round 3 found three real defects in the
new outer optimizer; every other finding was rejected. Rounds 5-7 returned only
repeats of already-rejected claims, so the review has converged.

**Fixed** (all in the new `outerOpt="trust"` code, `60380674d` / `0efe43dc9`):

* `outerTrustHessian="fd"` read the outer gradient at points the objective had
  not settled. The gradient callback warm-starts the inner problem from the last
  evaluation's ETAs, so a probe read cold returns the gradient at a stale
  conditional mode -- measured on `theo_sd`, ~9e-4 on components of order 200,
  which a 1e-3 difference step turns into an O(1) error in the Hessian entries.
* `outerTrustRelStep` never reached the analytical outer Hessian: the supplier
  called `control$hessian(x)` and `foceiOuterH()` kept its own `1e-3` default.
* The damped-BFGS update accepted any secant pair with a positive denominator;
  it now uses `trustHessianUpdate()`'s near-zero-denominator skip
  (`src/trustHessianUpdate.h`), so a reject-then-shrink step cannot inject an
  enormous rank-2 correction.

**Rejected, and why** -- each checked rather than argued:

* *"`foceiRSubjectFoceFR_` drops the third-order `eta_a' S_etaeta eta_b` term
  from `eta2`, and the 2e-3 tolerance in `test-focei-outer-hessian-family.R`
  hides it."* The `Tnf` loop is present (`src/foceiGrad.cpp:1100-1102`); the
  quoted "buggy" snippet does not exist in the file. Measured against tight
  central differences, the FOCE Hessian converges at the finite-difference rate
  (relative Frobenius 1.9e-5 at h=1e-3 -> 2.5e-7 at h=5e-5), which a missing
  O(1) term could not do. Deleting the loop as an experiment moved the FOCE
  Hessian by 87% relative -- so the existing tolerance catches this class of
  bug by a factor of ~400, not marginally.
* *"`foceiHessianExpand` reads `e.Rsig.col(g.sigCol[k]-1)` out of bounds when a
  fixed sigma precedes an estimated one."* `sigCol` is `seq_len(nsg)` built from
  an already-fix-filtered `sgName` (`R/foceiGradAnalytic.R:523`, `:792`), so it
  is the identity. Measured on an `add()+prop()` model with `add.sd` fixed: the
  analytic Hessian agrees with central differences to 6.4e-6, the same as with
  both estimated, and is exactly the all-estimated matrix with that row/column
  removed.
* *"`outerSolveFill` takes `OdeSwapCmtScope` inside `fillFromSolve`, so the
  solve runs outside the compartment guard and races across threads."*
  `OdeSwapCmtScope::shift()` mutates the individual's own event rows, and the
  OpenMP loop is over subjects, so each thread has its own `ind`. The hunk is
  also unchanged by this PR -- it only gained a `subject` parameter.
* *"`calcModelEtaHessian`'s `resetOpBadSolve(op)` races on shared state."* Same
  pattern as `likInner0`, which documents it in place: `op->badSolve` is racy
  under `cores>1` and the decision is read from the per-subject
  `indHasBadSolve()`.
* *"The Hessian probe leaves the engine at a perturbed point."* Control
  experiment on `theo_sd`, same `fn`/`gr` pair with and without the probe in
  between: without it the objective drifts 1.2e-5 and the gradient 8.6e-4 (the
  ordinary warm-start settling); with it, 2.3e-11 and exactly 0.
* *"`innerHessian="conditional"` drops the transform-both-sides Jacobian, so the
  inner problem optimizes a different objective."* `fInd->llik` excludes
  `tbsLik` in both paths by the same convention -- `LikInner2` adds
  `fInd->tbsLik` separately. Checked against a brute-force conditional mode on
  `theo_sd`: the conditional curvature lands *closer* to it (3.1e-5) than the
  FOCEI curvature does (7.6e-5), which is the point of the option.
* *"`R/foceiControl.R` calls `_nlmixr2est_foceiOuterH` with 3 arguments."* It is
  2 everywhere -- `R/RcppExports.R`, `src/RcppExports.cpp`, `src/init.c:359` --
  and `R/foceiControl.R` does not reference it.
* *"`LikInner2` builds the Laplace/AGQ determinant with the Gauss-Newton
  Hessian even under `innerHessian="conditional"`."* That is the documented
  design ("The marginal objective's FOCEI curvature is unchanged"), and
  `test-focei-inner-conditional.R` asserts it.

### Two additions

**`foceiControl(outerOpt="trust")`** -- a trust-region Newton outer optimizer
(`RcppTrust`) driven by this branch's analytical outer Hessian, which until now
only `nlminb` consumed. `outerTrustHessian=` selects the curvature (`"auto"`,
`"analytic"` under `fast=TRUE`, a damped-BFGS update, or a finite difference of
the outer gradient), with `outerTrustRinit`/`Rmax`, `outerTrustFterm`/`Mterm`,
`outerTrustRelStep` and `outerTrustRestarts` for the region, its tolerances and
the Hessian step.

`trust_solve_c()` reports `converged=TRUE` off its own step/model tolerance,
which a collapsing trust region satisfies at a point that is not stationary at
all -- measured, a fit exiting at 133.34 whose full Newton step still predicted
a decrease of 530. The reported point is therefore checked with its Newton
decrement and the region re-entered (`outerTrustRestarts`, budget shared with
`maxOuterIterations`); a run that still stops short says so rather than
claiming convergence. On four `theo_sd` variants (add+prop, correlated omega,
a covariate, three ETAs) `"trust"` matches or slightly beats `outerOpt="nlminb"`
on the objective; with a warm model cache the two cost about the same.

**`est="flaplace"`, `"mflaplace"`, `"iflaplace"`, `"fagq"`, `"mfagq"`,
`"ifagq"`** -- the Laplace and adaptive-quadrature methods, plus their
mu-referenced `"lin"`/`"irls"` variants, run with the full conditional inner
curvature (`fast=TRUE, innerHessian="conditional"`). They report as
`Full Laplace` / `Full AGQ` in `$method` and in the `objDf` row, and require
Gaussian endpoints. Note what the name does and does not claim: the *inner*
problem uses the full conditional Hessian, while the marginal objective keeps
this branch's FOCEI determinant, per the design above.

New tests: `test-focei-outer-trust.R` (57 assertions) and
`test-focei-full-inner.R` (37). Both stay in the essential push/PR subset --
they run in about 3s each.

mattfidler:
The AI merged this, but I do have comments -- The outer gradient needs to be in the ode pool so that it can be used efficiently without swapping out memory.  Once that is done, we can work on the trust outer opt too.

@hiddevandebeek would you like to try to fix this?

hiddevandebeek:
@mattfidler Ah, will look at it

PR #1052, opened 2026-09-09
fix(mix): a mixture fit's table reports the mixture it actually fit

mattfidler:
> **Depends on nlmixr2/rxode2#1358, now MERGED.** The root cause is in rxode2;
> this PR fixes the nlmixr2est-side faults stacked on top of it and stops
> working around it. The end-to-end test here fails against an rxode2 built
> before that PR, so a local checkout needs rxode2 from source to see it pass
> (the installed 5.1.7 predates it).

## The bug

Reported in #1041: a SAEM mixture fit completes, but the returned table has

- the model's `mixest` output 0 for every row,
- its `mixnum` output 0 for every row,
- the result of `mix()` itself 0 for every row,

and `PRED`/`IPRED` computed from those zeros -- so a `Kel` of 0, a compartment
that never eliminates, and a flat profile. Silently wrong predictions, not an
error, and everything derived from them (`IWRES`, `CWRES`, `NPDE`, every
goodness-of-fit plot) inherits it.

Four separate faults stacked up on the same path.

## Root cause (fixed in rxode2)

Every method's prediction model is built through symengine, which expands the
`mix()` call into per-component selectors. rxode2 then no longer read the model
as a mixture at all, discarded the per-individual `mixest`, and left every
selector false. nlmixr2/rxode2#1358.

## Fixed here

**The eta/theta pairing shifts -- and this one is not mixture-specific.**
The eta-to-theta map is an index into the SAEM *estimation* parameter vector; it
was used to subscript the model text built in `ini()` order. The two differ by
every theta `saem` does not estimate as a parameter of its own. In the reported
model (`kel1`, `kel2`, `p1`, `tv`) the generated model came out as

```
p1 = THETA[3] + ETA[3];
tv = THETA[4];
```

so the volume's eta landed on the mixture probability and the volume got none.

**The mixture probabilities are not the only thing dropped from that vector --
the mu-referenced covariate parameters are too.** So an ordinary covariate model
with no mixture anywhere, declaring `tcl.wt` before `tv`, had exactly the same
shift. On `theo_sd`:

| | `v` distinct values in the table |
|---|---|
| before | **1** (constant -- the eta never applied) |
| after | 12, one per subject |

That is a long-standing bug in every SAEM fit whose `ini()` declares a covariate
parameter before a mu-referenced theta: the table's individual parameters and
`IPRED` were missing that eta entirely. Found by the antigravity review round on
this branch. Now paired by name, and an unresolvable name is a `stop()` rather
than a silent fall back to the position that was wrong to begin with.

**The iCov `ID` is always a factor.** `.saemMixFix()`/`.mixFix()` build it as an
integer, but output creation re-levels every `ID` in the fit environment
afterwards, so rxode2 rejected it. The rejection message did not match the
retry-without-iCov guard either, so the *whole table step* was dropped and the
fit came back as a bare `nlmixr2FitCore`. Coerced at the point of use, and the
guard widened so a rejected iCov can never take the table step down again.

**The zeros are no longer silent.** An rxode2 without nlmixr2/rxode2#1358 does
not *reject* the iCov, it just never reads a mixture out of the expanded model
-- nothing errors, and the table's mixture columns are quietly 0. The table step
now checks whether the model it is about to solve reports a mixture at all and
puts it in the fit's `$runInfo` when it does not.

**`.mixFixTable()` is removed.** It post-corrected the mixture output columns,
but only for columns literally named `me`, `mn` and `mu` -- the reported model
named them `fitted_mix_selector`/`number_of_mix_components`, so nothing fired.
And with the solve fixed it was actively corrupting correct values: it wrote the
per-subject component into the column holding `mixnum`, which is the component
*count*, not the assignment. Measured on a model that does use those names,
before removal: `me` 1/2 (right), **`mn` 1/2 (wrong, should be a constant 2)**,
`mu` a fabricated `0.5`.

## Result

Same 40-subject two-component fit, before and after:

| column | before | after |
|---|---|---|
| `fitted_mix_selector` (`mixest`) | `0` everywhere | 1 / 2, matching `fit$mixNum` |
| `number_of_mix_components` (`mixnum`) | `0` everywhere | `2` |
| `kelRaw` / `Kel` (the `mix()` result) | `0` everywhere | 18 distinct, equal to the selected component |
| `Vol` | constant (`tv`, no eta) | 40 distinct |
| `IPRED` | 3 distinct, identical to `PRED` | 640 distinct |

`IPRED` now matches a closed-form solution of the model at the fitted
parameters and etas to 1.1e-06 relative.

Mixture **FOCEi** is fixed by the same rxode2 change: its `PRED` was also
computed with `Kel = 0` (3 distinct values) and is now 32 -- 16 times x 2
components, which is what a population prediction under a two-component
mixture should be.

## Note on `mixunif`

A model that reads `mixunif` in an expanded prediction model now gets the
supplied component back rather than the fabricated `1/nMix`. The expansion
drops the mixture probabilities along with the `mix()` call, so there is nothing
to sample from there; use `mixest`.

## Testing

- `test-mix.R` (essential subset, no fit): the generated model pairs each eta
  with the theta it is mu-referenced to, and neither a mixture probability nor a
  covariate parameter carries one. Fails 5 assertions without the fix.
- `test-saem-mix.R` (weekly batch): an end-to-end fit asserting `mixnum` is the
  constant count, `mixest` is the component `fit$mixNum` assigned, `mix()`
  returns the selected component, and `IPRED` still separates from `PRED`. The
  model deliberately declares `p1` before a mu-referenced theta so it covers the
  pairing shift too. Fails 1 assertion without the pairing fix and 4 against an
  rxode2 without nlmixr2/rxode2#1358.

`test-mix.R` (56), `test-saem-mix.R` (54), `test-saem-mix-regress.R` (4) and
`test-npag-mixture.R` (20) pass, plus `test-cwres.R` (32),
`test-resid-ode-fallback.R` (7) and `test-npde.R` (21) since `resid.R` is on
every method's table path -- 0 failures, 0 skips. Non-mixture `saem`/`focei`
fits still build normal tables. The rest is left to CI.

## Merged `origin/main`

Brought up to date with `origin/main`, which since this PR opened has picked up
#1061 (the fix for #1058). Both conflicts were additive and both sides were
kept: NEWS bullets under the same `### Mixture models` heading, and two
independent `test_that` blocks in `test-mix.R`.

Re-run after the merge: `test-mix.R` 61/1, `test-saem-mix.R` 63/4,
`test-npag-mixture.R` 20/0. All five failing assertions are this PR's own, from
`2e97d579e`, and are the documented rxode2 dependency above -- they need an
rxode2 carrying #1358, which the locally installed 5.1.7 does not. `git blame`
confirms none of them came from the merge.


mattfidler:
## Antigravity review summary

Three rounds against `agy` (`gemini-3.1-pro-high`), each a single narrow
question. **Three findings were real and are fixed in a follow-up commit**; the
rest were checked against the code and dismissed.

### Real findings, fixed

1. **"The two vectors can differ in ways other than the mixture probabilities."**
   They can, and this is the important one: `saemParamsToEstimateCov()` also
   drops the **mu-referenced covariate parameters**. So the eta-shift is *not
   mixture-specific* -- an ordinary covariate model that declares `tcl.wt`
   before `tv` had the volume's eta land on `tcl.wt` and the volume get none.
   Measured end to end on `theo_sd`: the fit table's `v` had **1** distinct
   value before the fix and **12** (one per subject) after, so `IPRED` was
   carrying no volume between-subject variability at all. Long-standing, and
   nothing to do with mixtures. Now covered by its own test.

2. **The unresolvable-name fallback silently reinstated the bug.** It fell back
   to the raw position, which is precisely what was wrong. It is now a `stop()`
   with a clear message -- generating a silently mis-paired model is worse than
   failing.

3. **The widened retry guard risked swallowing real errors silently.** The
   underlying concern was right, but the sharper problem was the opposite one:
   an rxode2 without nlmixr2/rxode2#1358 does not *reject* the iCov at all, it
   just never reads a mixture out of the expanded model -- so nothing errors and
   the zeros were completely silent. The table step now checks whether the model
   it is about to solve reports a mixture and puts it in `$runInfo` when it does
   not. Verified: fires on the old rxode2, silent on the fixed one, silent for
   non-mixture fits.

### Verified and dismissed

- *"Filtering `nonMuEtas` out shifts the index"* -- `.saemEtaTrans()` applies
  that same filter **before** computing the index (`nonMu = TRUE`); the
  replacement replicates it exactly, so the two vectors line up.
- *"Filtering THETA names by ETA names is a namespace confusion"* --
  `saemParamsToEstimate()` appends `nonMuEtas` into that vector itself, so the
  filter is removing what it put there. Pre-existing and intentional.
- *"PRED uses a different, mixture-stripped model, so the new guard will warn on
  a valid PRED table"* -- PRED and IPRED are solved from the **same** model
  object (`R/resid.R:279` and `:282`); only the parameter vector differs (etas
  zeroed). Confirmed by measurement: no warning on either a SAEM or a FOCEi
  mixture fit against the fixed rxode2.
- *"Duplicate names in the value vector would only update the first match"* --
  `iniDf` parameter names are unique.
- *"Deleting `.mixFixTable()` leaves the 0s uncorrected on an old rxode2"* --
  correct, and deliberate. What it "corrected" was itself wrong: it wrote the
  per-subject component into the column holding `mixnum`, which is the component
  *count*, and it only ever fired for columns literally named `me`/`mn`/`mu`.
  Finding 3 is the honest replacement -- the fit now says the mixture did not
  reach the table instead of handing back plausible-looking wrong numbers.

### Testing after the review

`test-mix.R` (56), `test-saem-mix.R` (54), `test-saem-mix-regress.R` (4),
`test-npag-mixture.R` (20), plus `test-cwres.R` (32),
`test-resid-ode-fallback.R` (7) and `test-npde.R` (21) since `resid.R` is on
every method's table path -- 0 failures, 0 skips. The rest is left to CI.

PR #1054, opened 2026-09-10
feat(mix): mixture-proportion standard errors, an exact mixture gradient, and est="vae" support

mattfidler:
> **Depends on nlmixr2/rxode2#1358** (merged upstream) for the mixture columns
> to reach the fitted table at all. A local checkout needs rxode2 from source
> to see the mixture-table tests pass -- the installed 5.1.7 predates it.

Started as "make `est="vae"` fit a `mix()` model". Establishing whether it fit
one *correctly* meant checking the mixture gradient against a finite difference,
which turned out to be wrong in FOCEi too -- and fixing that made the mixture
proportions' information matrix available, which they never had. So this PR is
now three things.

## 1. `est="vae"` fits a mixture model

It could not before: the fit stopped during assembly, so only the training loop
had ever run with a mixture. It now follows the same convention as the
gradient-based methods -- every one of the `nSub x nMix` subject-component pairs
is characterized by the encoder at its own eta, membership is the posterior
responsibility, and the proportions are estimated on the mlogit scale through
their analytic gradient, consumed by the same Adam loop that trains the encoder.

The objective was also wrong in a way no existing test could see: it
marginalized a **square root** likelihood (`exp(-obj/2)` negated twice). Both
errors cancel exactly at `nMix == 1` *and* at identical components with uniform
`pi`, which is why `p1 = 0.5` in the one existing mixture test never caught it.

## 2. The mixture-proportion gradient was wrong -- in FOCEi

`mixGrad()` short-circuits the finite difference in `numericGrad()`, so a wrong
value is invisible under the default derivative-free `outerOpt="bobyqa"` and
silently wrong under every gradient-based one. Two independent defects:

- **the missing `-2`.** `foceiObjFromLik0() = -2*foceiLik0()`, but `mixGrad`
  returned the gradient of `lik`. That flips the sign at *every* `nMix`,
  including 2 -- `lbfgsb3c`'s first step went uphill, the line search rejected
  it, and the proportions never moved off their `ini()` values;
- **the diagonal-only Jacobian.** `rxode2::dmexpit()` returns only the diagonal,
  so every `m != l` term was dropped -- additionally wrong from `nMix >= 3` up.

`pi = mexpit(t)` is a softmax, so `d(pi_m)/d(t_l) = pi_m(delta_ml - pi_l)`, and
with `sum_m r_im = 1` the whole Jacobian collapses to `-2 * sum_i (r_il - pi_l)`.
Checked against a central difference of the objective, 3 components on `theo_sd`:

| | dOFV/dt1 | dOFV/dt2 |
|---|---|---|
| finite difference of the objective | -7.3939793 | 2.7728427 |
| closed form (this PR) | -7.3939794 | 2.7728427 |
| old `mixGrad()` | +4.7367 | +1.4556 |

This is the NONMEM 7 Technical Guide's own formulation -- eq. (1.194) chained by
(1.197), whose `da/dtheta_a` is the FULL Jacobian. Verified equal to a literal
evaluation of (1.194)x(1.197) to 2e-16. The old diagonal-only form was the
departure from it, not the fix.

## 3. Mixture proportions get standard errors

They were forced out of the covariance entirely (`skipCov`), so `p1` reported
`SE = NA` under every method. NONMEM (7.51)-(7.54) build the mixture
parameters' information as the outer product of the same per-subject scores
`g_ia` the gradient uses -- and nlmixr2est's S matrix already *is* that outer
product, so the block drops in once the per-subject score exists. The
per-subject mixture score is taken analytically, so it costs no extra solves.

**This also repaired the S matrix for mixture models generally.** `foceiS()`
built each subject's score by finite-differencing component 0's likelihood
rather than the marginal `log(sum_m pi_m L_im)`, so a parameter entering only
another component got an exactly-zero score. Measured on a 3-component fit,
two of six diagonal entries came out at `1.9e-18` and `6.6e-10`, S was reported
non-positive-definite, and `covMethod="r,s"` silently degraded to `"r"`. The
same model without `mix()` was fine.

Reported on the **probability** scale: the covariance is rotated with the full
mexpit Jacobian `J = diag(p) - p p'` (the off-diagonal terms are what carry the
proportions' correlations across), and the CI is taken on the logit scale,
`expit(logit(p) +/- z*SE/(p(1-p)))`, so it stays inside `(0, 1)` and is
asymmetric. A symmetric interval had a real fit printing
`p1 = 0.648 (-0.045, 1.34)`.

| `covMethod` | proportion SE |
|---|---|
| `"r,s"`, `"r"`, `"s"` | yes |
| `"imp"` | yes |
| `"analytic"` | declines for a mixture (it differentiates one component's conditional likelihood and has no proportion block) and falls back |
| `linFim`/`fim`/`sa` (`saem`) | appended from the fit's own responsibilities, gated on the score statistic `s' I^-1 s` |

`covMethod="imp"` also had to build its Monte-Carlo proposals for every
component rather than component 0 alone, or the proportions' directions were
exactly flat and their SEs came back 0.

## Calibrated against ground truth

Not just internal consistency: 60 replicates of a 2-component model, mean
reported SE **0.06450** against the analytic `sqrt(p(1-p)/n)` = **0.06423**,
with `sd(p1hat)` tracking the realized group fractions (`cor = 0.997`).
`r,s` / `r` / `s` / `imp` agree within 2%.

## Also fixed

- **A proportion now binds to the `mix()` component it was written for.**
  `thetaMixIndex` was built with `which(names(theta) %in% mixProbs)`, which
  returns ascending theta positions -- `ini()` order -- while `ui$mixProbs` is
  in `mix()`-call order. Declaring the proportions in a different order than
  `mix()` uses them ran component 1 on `p2`'s value. At zero iterations with
  `ini({p2 <- 0.20; p1 <- 0.70})`, `$mixProbabilities` came back `0.2 0.7 0.1`
  instead of `0.7 0.2 0.1`. A model whose `ini()` order already matches its
  `mix()` order -- the usual way to write one -- was never affected.
- `fit$etaMat` no longer carries the `mixnum` column, which every consumer
  handing it back as `foceiControl(etaMat=)` compared against the model's
  `neta` and stopped on.
- A mixture fit's `$ui` carries the proportion on the probability scale, so the
  fit is re-fittable rather than failing its own `ini()` validation.
- A fix()ed proportion is no longer given an appended variance for a parameter
  that was never estimated.
- A near-boundary proportion says so in `$runInfo`: the `p(1-p)` factor drives
  its SE toward zero, which is arithmetically right and reads as certainty.

## Review

Ten rounds with an independent reviewer (antigravity, `gemini-3.1-pro-high`),
run to convergence -- the last two clean. Eight real defects found and fixed
(stale `lik[2]` in the central difference; an `nExp`/`nsub` loop bound;
`setCov()` re-rotating an already-rotated covariance, SE `0.06445 -> 0.01548`
per round trip; a dropped block on recompute; the `mixProbs` ordering above; a
double rotation from `.covRecompute`; an N-dependent stationarity tolerance; an
all-or-nothing rotation that left a `fix()`ed proportion's neighbours on the
mlogit scale, measured `0.2652` where the probability scale is `0.0634`). Five
claims were checked and rejected with evidence, including a suggested
`gammaVec[id % nsub]` that would have used component 0's proposal scale for
every component.

CodeFactor: 5 issues -> 0 (the last one refactored, pending re-analysis).

## Verification

482 assertions green across `mix`, `saem-mix`, `vae-mix`, `focei-mix-grad`,
`focei-mix-cov`, `npag-mixture`, `cov-decouple-saimp`, `cov-condition`,
`saem-cov`, `nlmixr2output`, `vae-inner`, plus 269 more over the paths that
exercise the shared log-sum-exp (`imp-tprop` 28, `imp-proposal` 148,
`npag-psi` 76, `vae-elbo` 9, `vae-train` 8), all clean. New `test-focei-mix-grad.R` and
`test-focei-mix-cov.R` are cheap enough to stay out of `.slowBatches`; both were
confirmed to FAIL against the pre-fix code rather than passing either way.

`cov-analytic` is 243/4 -- those four are pre-existing, reproduced on an
untouched worktree containing `origin/main`, and filed as #1055, #1056, #1057.

## Housekeeping

**No duplicated log-sum-exp.** Four implementations of the same max-shifted
sum had accumulated -- `impLogSumExp3()` (pre-existing), `foceiMixLogSumExp()`,
and two inline blocks in `vaeElboStepCpp`/`gVaeThetaObjR` (the latter three
added here). They now share `rxLogSumExpW()` in a new `src/logSumExp.h`, a
header and `static inline` because `impPropLogKernelRecip()` calls it once per
importance sample.

The two callers are not interchangeable and the shared core says so rather
than papering over it: a mixture marginal must **skip** a component that failed
to solve (it contributes nothing, and the caller charges its own
`badSolveObjfAdj`), while the importance-sampling kernel has always let a NaN
**propagate** -- quietly dropping one there would turn a loud failure into a
plausible number. Hence the explicit `skipNonFinite` argument and the two named
entry points, `rxLogSumExpMix()` and `rxLogSumExp()`.

Two smaller repeats collapsed the same way: `diag(p) - p p'` (the mexpit
Jacobian, built in two places) is now `.mixProbJacobian()`, and the S-matrix
"this subject had no usable score" fallback (set `gfull`, flag `hasZero`, dock
`sInfoPer`) is now `foceiSMixNoInfo()`, which gives the `sInfoPer` accounting
somewhere to be explained -- it feeds the `smatPer` gate.

Deliberately left alone: `mixGrad()`'s population score and
`foceiSMixScore()`'s per-subject score share the expression `-2*(r_il - pi_l)`
but not its shape (one sums over subjects, the other stores per subject), and
`vaeLsAdd`/`vaeLsSwap` share two lines across different neighborhood moves.
Extracting either would be noise.

**Two orphaned roxygen blocks put back.** Inserting a function directly above
an existing one's roxygen leaves the block attached to the *new* function and
the old one undocumented. `.vaeSetIniMixProb()` had taken `.vaeUpdateModel()`'s
docs (mine); `.saemIovRestoreUi()` had taken `.saemIovFinalizeCollapsed()`'s
(pre-existing on main, same defect, so fixed here too). Swept `R/` for the
signature -- a block carrying more than one `@noRd`/`@author` -- and these were
the only two.

## Merged `origin/main`

Kept current with `origin/main` throughout, which over the course of this PR
merged #1052 (the mixture-table fixes this branch was developed alongside),
#1053, #1059 and #1061 (the fix for #1058, filed from this branch's work).

Two semantic conflicts git could not see, both caught by tests rather than
markers:

- #1058's fix makes `saem` report proportions already on the natural scale and
  guards the shared back-transform hook with `env$mixProbNatural`. This branch
  adds a *second* back-transform in `.nlmixr2FitUpdateParams()` that had no such
  guard, so a `saem` fit's `ui` came back with `expit(p)` -- 0.4 written as
  0.599. Guarded, and main's own regression test for that identity passes.
- #1052 **removed** `.mixFixTable()`, which this branch still called; that would
  have been a runtime error, not a conflict. Dropped, keeping
  `.mixCovAppendBlock()`.

#1053 also landed on the exact lines where this branch rotates the covariance.
Both kept, with the rotation still *before* `.updateParFixed()` -- it has to be,
or the SEs are derived from the mlogit-scale covariance -- and #1053's refresh
after it. The mixture CI survives that refresh because `.mixParFixedCi()` is
called from `.updateParFixedRefreshSeFromCov()` as well.

A welcome interaction: with #1058 fixed, `saem` now reaches the mixture's
score-zero point, so the gate on the `saem` SE block stops declining and starts
reporting -- with no change to the gate.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -144,7 +144,9 @@
   initial values; from three components up the magnitude was wrong too.  The
   full Jacobian collapses to `-2 * sum_i (r_il - pi_l)`, which is what it now
   uses -- checked against a central difference of the objective to eight
-  significant digits at three components.  Fits left on the default
+  significant digits at three components, and equal to a literal evaluation of
+  the NONMEM 7 Technical Guide's own equations (1.194) and (1.197) for the
+  mixture-proportion gradient.  Fits left on the default
   derivative-free `outerOpt="bobyqa"` never reached this code and are
   unchanged, as are the reported standard errors (a mixture proportion is
   `skipCov`).
--- a/NEWS.md
+++ b/NEWS.md
@@ -152,6 +152,15 @@
   two defects above stopped every such fit during assembly, so only the
   training loop had ever run with a mixture.
 
+- `est="vae"` estimates the mixture proportions.  They were read once from
+  `ini()` and never updated, so the reported proportion was whatever the model
+  started at.  They are now estimated on the mlogit scale through their own
+  analytic gradient -- the same chain `focei` uses (`mixGrad`): each subject's
+  responsibility difference against the last, non-free component, times the
+  `mexpit` Jacobian -- consumed by the same Adam loop that trains the encoder.
+  The gradient agrees with finite differences to 1e-4.  On simulated data with
+  a 3:1 split started from 0.5, the fitted proportion comes back at 0.74.
+
 - `est="vae"` no longer treats a mixture proportion as an ordinary structural
   theta.  `p1` appears inside `mix(a, p1, b)`, so it passed the "does this
   theta appear in a model expression" filter and became a `nonMuTheta`
--- a/NEWS.md
+++ b/NEWS.md
@@ -151,6 +151,12 @@
   across rather than dropped.  Same principle as `covFull` reporting Omega on the
   natural variance scale instead of `chol(solve(omega))`.
 
+  That rotation carries a factor of `p(1-p)`, so a proportion sitting near 0 or
+  1 gets an SE that shrinks toward zero.  It is the right delta-method answer
+  but it reads as certainty, when in fact the symmetric Wald interval has
+  stopped being meaningful there, so the fit's `$runInfo` now says when a
+  proportion is at a boundary.
+
 - The S matrix is no longer singular for mixture models, so `covMethod="r,s"`
   stops silently degrading to `"r"`.  `foceiS()` built each subject's score by
   finite-differencing component 0's likelihood instead of the marginal
--- a/NEWS.md
+++ b/NEWS.md
@@ -181,11 +181,12 @@
   structural parameters and its SEs are mildly optimistic.
 
   The block is only reported when the fit is actually at the mixture's fixed
-  point, `p_l == mean_i r_il`.  An information matrix describes the precision of
-  a maximum-likelihood estimate, and away from that point it is a
+  point, judged by the score statistic `s' I^-1 s` (which is on a chi-square
+  scale, so unlike a tolerance on `mean(r) - p` it does not loosen as the number
+  of subjects grows).  An information matrix describes the precision of a
+  maximum-likelihood estimate, and away from that point it is a
   confident-looking number attached to something that is not one; instead the
-  fit's `$runInfo` says the SE was not computed and by how much the identity
-  fails.  `est="saem"` currently lands far from it (nlmixr2/nlmixr2est#1058), so
+  fit's `$runInfo` says the SE was skipped.  `est="saem"` currently lands far from it (nlmixr2/nlmixr2est#1058), so
   in practice this declines today and will start reporting once that is fixed.
 
 - `est="focei"` estimates the mixture proportions under a gradient-based
--- a/NEWS.md
+++ b/NEWS.md
@@ -165,6 +165,12 @@
   across rather than dropped.  Same principle as `covFull` reporting Omega on the
   natural variance scale instead of `chol(solve(omega))`.
 
+  The confidence interval is taken on the logit scale,
+  `expit(logit(p) +/- z*SE/(p(1-p)))`, from the SE actually reported.  The
+  generic symmetric `est +/- z*SE` interval walks out of `(0, 1)` for a
+  proportion (a fit reported `p1 = 0.648 (-0.045, 1.34)`) and was built from the
+  covariance before the rotation, so it did not agree with the SE beside it.
+
   That rotation carries a factor of `p(1-p)`, so a proportion sitting near 0 or
   1 gets an SE that shrinks toward zero.  It is the right delta-method answer
   but it reads as certainty, when in fact the symmetric Wald interval has
--- a/NEWS.md
+++ b/NEWS.md
@@ -171,6 +171,23 @@
   The augmented sensitivity model differentiates one component's conditional
   likelihood and has no mixture-proportion block at all.
 
+- `est="saem"` mixture fits can report a mixture-proportion SE as well.  `saem`
+  leaves the proportions out of the parameter vector its kernel converges (they
+  are updated by a separate EM step), so its `linFim`/`fim`/`sa` covariance has
+  no mixture rows to extend; the (7.51) block is appended from the fit's own
+  posterior responsibilities.  The cross terms (7.52)-(7.54) are NOT formed --
+  they need per-subject scores for the other parameters on the same footing,
+  which that covariance does not expose -- so the block is uncorrelated with the
+  structural parameters and its SEs are mildly optimistic.
+
+  The block is only reported when the fit is actually at the mixture's fixed
+  point, `p_l == mean_i r_il`.  An information matrix describes the precision of
+  a maximum-likelihood estimate, and away from that point it is a
+  confident-looking number attached to something that is not one; instead the
+  fit's `$runInfo` says the SE was not computed and by how much the identity
+  fails.  `est="saem"` currently lands far from it (nlmixr2/nlmixr2est#1058), so
+  in practice this declines today and will start reporting once that is fixed.
+
 - `est="focei"` estimates the mixture proportions under a gradient-based
   `outerOpt`.  `mixGrad()` supplies an analytic value that short-circuits the
   finite difference in `numericGrad()`, and it chained the per-subject
--- a/NEWS.md
+++ b/NEWS.md
@@ -132,6 +132,22 @@
 
 ## Bug fixes
 
+### Mixture models
+
+- `fit$etaMat` no longer carries the `mixnum` column that `$eta` gains for a
+  mixture fit.  Every consumer that hands it back as `foceiControl(etaMat=)`
+  compared `neta + 1` columns against the model's `neta` and stopped with "The
+  etaMat must have the same number of ETAs (cols) as the model" -- so `$cov`,
+  `addCwres()`, the FO objective and re-fitting a fit were all failing for
+  every mixture fit, in most cases inside a `try()` that swallowed it.
+
+- A mixture fit's `$ui` now carries the mixture probability on the probability
+  scale rather than the mlogit scale it is estimated on.  `fullTheta` is left
+  on the estimation scale and only `$theta` was back-transformed, so `$ui`
+  reported `p1 = -0.847` where `fixef()` reported `0.3`, and re-fitting the fit
+  failed its own `ini()` validation ("the probabilities in a mixture must sum
+  to a number between 0 and 1").
+
 - A `focei`-family fit now reports whether its inner solves actually
   converged.  `fit$env$nTrustInner` breaks the `innerOpt="trust"` per-subject
   Newton solves down by outcome (`calls`, `error`, `notConverged`,
--- a/NEWS.md
+++ b/NEWS.md
@@ -134,6 +134,21 @@
 
 ### Mixture models
 
+- `est="focei"` estimates the mixture proportions under a gradient-based
+  `outerOpt`.  `mixGrad()` supplies an analytic value that short-circuits the
+  finite difference in `numericGrad()`, and it chained the per-subject
+  responsibility sum through the *diagonal* of the `mexpit` Jacobian
+  (`dmexpit()` returns only the diagonal) while dropping the `-2` of the
+  objective scale.  That flipped the sign at every number of components, so the
+  line search rejected the first step and the proportions never moved off their
+  initial values; from three components up the magnitude was wrong too.  The
+  full Jacobian collapses to `-2 * sum_i (r_il - pi_l)`, which is what it now
+  uses -- checked against a central difference of the objective to eight
+  significant digits at three components.  Fits left on the default
+  derivative-free `outerOpt="bobyqa"` never reached this code and are
+  unchanged, as are the reported standard errors (a mixture proportion is
+  `skipCov`).
+
 - `fit$etaMat` no longer carries the `mixnum` column that `$eta` gains for a
   mixture fit.  Every consumer that hands it back as `foceiControl(etaMat=)`
   compared `neta + 1` columns against the model's `neta` and stopped with "The
@@ -155,9 +170,9 @@
 - `est="vae"` estimates the mixture proportions.  They were read once from
   `ini()` and never updated, so the reported proportion was whatever the model
   started at.  They are now estimated on the mlogit scale through their own
-  analytic gradient -- the same chain `focei` uses (`mixGrad`): each subject's
-  responsibility difference against the last, non-free component, reduced
-  against the full `mexpit` Jacobian -- consumed by the same Adam loop that trains the encoder.
+  analytic gradient -- the same closed form `focei`'s corrected `mixGrad()`
+  uses, each component's summed responsibility against its expected count --
+  consumed by the same Adam loop that trains the encoder.
   The gradient agrees with finite differences to 1e-4 at two components and at
   three.  On simulated data with
   a 3:1 split started from 0.5, the fitted proportion comes back at 0.74.
--- a/NEWS.md
+++ b/NEWS.md
@@ -148,6 +148,22 @@
   failed its own `ini()` validation ("the probabilities in a mixture must sum
   to a number between 0 and 1").
 
+- `est="vae"` can fit a mixture (`mix()`) model.  It could not before: the
+  two defects above stopped every such fit during assembly, so only the
+  training loop had ever run with a mixture.
+
+- `est="vae"` no longer treats a mixture proportion as an ordinary structural
+  theta.  `p1` appears inside `mix(a, p1, b)`, so it passed the "does this
+  theta appear in a model expression" filter and became a `nonMuTheta`
+  regression parameter, moved by `bobyqa` against the `(-Inf, Inf)` bounds it
+  carries in `iniDf` -- writing values like `p1 <- 10.6` back into `ini()`.
+  `est="npag"` already excluded them for the same reason.
+
+- `est="vae"` reads the mixture proportion on the scale the inner problem
+  reads it on.  The prepared theta vector held the raw `ini()` probability
+  while the inner problem passes that slot through `mexpit()`, so `p1 = 0.3`
+  was used as `mexpit(0.3) = 0.574`.
+
 - A `focei`-family fit now reports whether its inner solves actually
   converged.  `fit$env$nTrustInner` breaks the `innerOpt="trust"` per-subject
   Newton solves down by outcome (`calls`, `error`, `notConverged`,
--- a/NEWS.md
+++ b/NEWS.md
@@ -134,6 +134,20 @@
 
 ### Mixture models
 
+- A mixture proportion now binds to the `mix()` component it was written for,
+  whatever order `ini()` declares the proportions in.  `mix(a, p1, b, p2, c)`
+  means `p1` is component 1's share, but the theta slots were collected with
+  `which(names(theta) %in% mixProbs)`, which returns ascending theta positions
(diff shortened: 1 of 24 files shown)


mattfidler:
## Antigravity review summary

Five rounds against `agy` (`gemini-3.1-pro-high`), each a single narrow
question. **Three findings were real and are fixed**; the rest were settled
against the code or by measurement.

### Real findings, fixed

1. **The mixture-proportion gradient was wrong beyond two components.** Round 2
   asked whether the `mexpit` Jacobian is genuinely diagonal. It is not —
   `d(pi_m)/d(theta_l) = pi_m (delta_ml - pi_l)` — so applying one scalar per
   free parameter, which is what `mixGrad()` does, drops the off-diagonal
   `-pi_m pi_l` terms. With a single free parameter that is accidentally exact,
   so the two-component finite-difference check passed. Measured on three
   components, analytic vs central difference:

   | | free param 1 | free param 2 |
   |---|---|---|
   | before | 2.250019 | -1.049989 |
   | finite difference | 3.000034 | -2.400019 |
   | after | 3.000013 | -2.400002 |

   Chained properly it collapses to something simpler than what it replaced:
   summing `-r_im/pi_m` against the dense Jacobian over all components gives
   `-(r_il - pi_l)`, because the responsibilities sum to one. So the exact
   gradient is `N*pi_l - sum_i r_il`, with no Jacobian to get wrong. The new
   test uses three components — the smallest case that can see it.

2. **A subject whose components all failed to solve was handed to component 1.**
   `best` was still its initial `0`, so `resp[best] = 1` gave component 1 the
   whole responsibility and silently dragged the proportion estimate toward it.
   With no information the posterior *is* the prior, which is also how saem
   falls back when its own weights underflow.

3. **The reported proportions were one Adam step stale.** Each step reports the
   proportions it *used*, i.e. from before its own update, so the value returned
   and written into `ini()` lagged. Refreshed from the final theta.

### Settled by measurement, not argument

Two findings were pressed twice — that the selected component carries a
"phantom anti-prior" residual `(resp-1)*prior`, and that the prior correction is
applied at the wrong scale. A finite-difference check of the **encoder**
gradient under a mixture settles both at once: analytic against central
difference agrees to **2e-4 relative**. If either residual existed the FD would
not match. That check is now a test.

### Verified and dismissed

- *"Fixing a mixture proportion silently disables all mixture updates"* /
  *"a fixed parameter would be silently overridden"* — ran it: a `fix()`ed
  proportion comes back at exactly its `ini()` value. The fixed value wins at
  the solver level. Now pinned by a test, since nothing asserted it.
- *"Gradients cross-wired if mixture parameters are declared non-sequentially"*
  — `.getMixFromLog` and `op_focei.mixIdx` share one ordering, and the
  three-component test finite-differences **each** free parameter separately,
  which is exactly what a cross-wiring would break.
- *"Indexing a `[N*nMix]` object with a per-subject index"* — round 1 asked this
  directly and found nothing.
- *"`nMix == 1` skips the bad-solve penalty"* — true, and pre-existing for every
  non-mixture vae fit. Changing the non-mixture objective is outside a mixture
  PR and would need its own validation; noted rather than folded in.

### One open modelling point, raised twice and deliberately not changed

The KL and entropy are applied to the **selected** component only, so a
non-selected component's posterior variance is shaped by the data term alone.
That follows directly from the instruction that the components are not combined
and the selected one supplies the per-subject quantities — it is the objective
this PR was asked to implement, not an oversight. Its gradient is exact for that
objective (the FD check above). Applying the KL to every component's own
posterior instead is a defensible alternative and would scale `DKL` by roughly
`nMix`; worth deciding deliberately rather than by review round.

mattfidler:
Follow-up: the same Jacobian defect was in FOCEi, and worse there.

`mixGrad()` (`src/inner.cpp`) short-circuits the finite difference in `numericGrad()` with an analytic value. That value summed the per-subject responsibility contributions and chained them through the **diagonal** of the mexpit Jacobian (`rxode2::dmexpit()` returns only the diagonal), and never applied the `-2` of the objective scale (`foceiObjFromLik0() = -2*foceiLik0()`).

Two separate errors:

- **Missing `-2`** -- wrong sign at *every* number of components, including 2. `lbfgsb3c`'s first step went uphill, the line search rejected it, and the mixture proportions never moved off their initial values.
- **Diagonal-only Jacobian** -- additionally wrong in magnitude from 3 components up. At nMix == 2 the diagonal chain happens to reduce to the correct expression, so only the `-2` was wrong there.

Since `pi = mexpit(t)` is a softmax, `d(pi_m)/d(t_l) = pi_m*(delta_ml - pi_l)`, and with `sum_m r_im = 1` the whole Jacobian collapses to

```
d(-2*log-lik)/d(t_l) = -2 * sum_i (r_il - pi_l)
```

whose stationary point is the EM condition `pi_l == mean_i r_il`.

**Verification.** Central difference of the actual FOCEi objective in mlogit space, 3 components on `theo_sd`:

| | dOFV/dt1 | dOFV/dt2 |
|---|---|---|
| FD of the objective | -7.3939793 | 2.7728427 |
| closed form (new) | -7.3939794 | 2.7728427 |
| old `mixGrad()` | +4.7367 | +1.4556 |

Eight significant digits, against sign disagreement on both coordinates before.

**Scope.** `mixGrad()` is reached only from `numericGrad()`, so the default derivative-free `outerOpt="bobyqa"` never ran it and those fits are unchanged. Reported standard errors are unaffected too -- mixture proportions are `skipCov` (`R/focei.R`).

**Test.** New `tests/testthat/test-focei-mix-grad.R` (2- and 3-component, ~35 s, left out of `.slowBatches` so it runs on every push). Both fit simulated mixture data with everything but the proportions and one component TV `fix()`ed, and assert the objective actually decreased, that the fit landed on the EM fixed point, and that the fixed point is the simulated truth. Verified discriminating: 6/6 pass on the fix, 0/6 on the old code.

One design note worth recording -- the proportion must not be the *only* free parameter. With `npars == 1` the outer optimization never reaches `numericGrad()` at all, so `mixGrad()` never runs and such a test is vacuous (this is how the first draft of the 2-component test passed against the broken code).

Also collapsed `mixGrad()`'s trailing scale chain rule into `dUnscaleParDx()`, which sits directly below it and was already the identical ladder, and dropped an unused `theta` parameter.

Regression run: `mix` 54, `focei-mix-grad` 6, `vae-mix` 11, `saem-mix` 50, `npag-mixture` 20, `vae-inner` 75 -- 216 assertions, 0 failures.

mattfidler:
Independent review (antigravity, `gemini-3.1-pro-high`) run to convergence: **10 rounds, the last two clean**. Eight real defects found and fixed, five claims checked and rejected with evidence.

## Fixed

| # | Defect | Evidence |
|---|---|---|
| 1 | `foceiS()` central leg read the forward leg's `lik[2]` without checking that leg converged -- a failed component leaves it stale from a previous `cpar`, and `foceiMixObjSlot()` combines it into a finite but wrong marginal | forward success vector hoisted to function scope |
| 2 | `impComputeCov()` cleanup loop still ran to `nsub` after the perturbed solves widened to `nExp`, leaving every component but the first cached at the LAST FD perturbation | `src/imp.cpp` |
| 3 | `setCov()` round trip re-rotated an already-rotated covariance | SE 0.06445 -> **0.01548** per round trip; now exact |
| 4 | `.covInstallResult()` dropped the appended mixture block on a recompute | `saem` fell back to `SE = NA` after a `setCov()` |
| 5 | The rotation was keyed on `ui$mixProbs` (mix()-call order) while the covariance rows, `op_focei.mixProb`, `$mixProbabilities` and `se`/`popDf` are all keyed on the **theta slot** (ini order) | with `ini({p2; p1})` + `mix(a,p1,b,p2,c)`: `mixProbs` = `p1 p2`, `names(theta)[thetaMixIndex]` = `p2 p1` -- Jacobian on the wrong rows, SEs swapped |
| 6 | `.covRecompute()` returns a full re-fit's `$cov`, already rotated; `.covInstallResult()` rotated it again | now reports `mixRotated` |
| 7 | The `saem` stationarity gate used an ABSOLUTE `|mean(r) - p| > 0.01`, but the score is `N*(mean(r) - p)` -- so it loosened with N (a score of 1 at N=100, **100** at N=10000) | replaced by the score statistic `s' I^-1 s`, chi-square scale, N-invariant; a test holds the deviation fixed and grows N |
| 8 | A proportion absent from a given covariance (a `fix()`ed one is dropped by `skipCov`) aborted the whole rotation, leaving the REMAINING proportions on the mlogit scale | measured `p2` SE **0.2652** where the probability scale is 0.0634; now rotates the present subset (holding absent coordinates fixed makes the correct Jacobian exactly the submatrix) |

Plus two the review prompted rather than found: a `fix()`ed proportion was being given an appended variance for a parameter never estimated, and a boundary proportion now says so in `$runInfo` -- the `p(1-p)` factor drives its SE toward zero, which reads as certainty when the Wald interval has simply stopped being meaningful.

## Rejected, with evidence

- **`gammaVec[id]` out of bounds in `impComputeCov()`** after widening to `nExp`. It is allocated `arma::vec gammaVec(nExp)` (`src/imp.cpp:1302`) and documented at `:533` as one scale per EXPANDED subject. The suggested `gammaVec[id % nsub]` would be a regression -- component 0's converged proposal scale for every component.
- **The central leg skips failed subjects.** Not reachable: the `gfull` fallback is inside `if (doForward)` and the central leg only runs when `!doForward`, so the two are mutually exclusive.
- **`saem` leaves mixture rows unrotated on the mlogit scale** -- raised three times with three different mechanisms (`covMethod="imp"`, the analytic fallback, the FD fallback). Measured under each: a `saem` mixture fit produces no mixture rows at all (`linFim`, rows `tcl1 tcl2 tv`), so there is nothing to mis-scale.
- **NEWS claims about `est="vae"` and `fit$etaMat` are unsupported** -- flagged three times. Both appear in the reviewed diff as CONTEXT lines (leading space, not `+`); `git log -S` places the `etaMat` one in `7da13ec47` on this same branch.

## Verification

379 assertions green across `mix`, `saem-mix`, `vae-mix`, `focei-mix-grad`, `focei-mix-cov` (61), `npag-mixture`, `cov-decouple-saimp`, `cov-condition`, `saem-cov`, `nlmixr2output`. `cov-analytic` is 243/4, the four being pre-existing failures reproduced on an untouched worktree containing `origin/main` and filed as #1055, #1056, #1057.

Calibration against ground truth still holds: 60 replicates, mean reported SE 0.06450 vs analytic `sqrt(p(1-p)/n)` = 0.06423.

mattfidler:
Two follow-ups: the `fixef()` component mismatch, and the CodeFactor check.

## `fixef()` component mismatch

`mix(a, p1, b, p2, c)` means **p1 is component 1's share**. `ui$thetaMixIndex` is what carries that binding -- every consumer reads `mixIdx[m]` as "component m's theta slot" (`op_focei.mixProb[m]`, `.getMixFromLog()` mapping the slots positionally through `mexpit()`, and the back-transform writing `probs[m]` to slot `mixIdx[m]`).

It was built with `which(names(theta) %in% mixProbs)`, which returns **ascending theta positions** -- `ini()` order. `ui$mixProbs` is in **mix()-call** order. The two agree only when a model declares the proportions in the same order it uses them.

Measured at zero iterations, so the answer is exactly the `ini()` values:

```r
ini({p2 <- 0.20; p1 <- 0.70})     # with mix(a, p1, b, p2, c)

before   $mixProbabilities   0.2 0.7 0.1     # component 1 running on p2's value
after    $mixProbabilities   0.7 0.2 0.1
```

Both `ini()` orders now give the identical, correct result. `match(ui$mixProbs, names(ui$theta))` keeps mixProbs' own order, so the index is component-ordered; no consumer assumed it was sorted (`foceiSetupMixTrans()` scans for a match, `impSetMixThetas()` writes by component, the rest index by `m`).

Worth being clear about the blast radius: this was not just a labelling problem. The **fit itself** ran with the proportions attached to the wrong components. A model whose `ini()` order already matches its `mix()` order -- the usual way to write one -- was never affected, which is why no existing test saw it. `test-mix.R` now pins the invariant with the proportions declared in reverse.

## CodeFactor

Was 5 issues; the skill's `pr 1054` list is now down to 1, and the check is green.

| issue | fix |
|---|---|
| `src/inner.cpp` Very Complex Method (67) -- `foceiS()` | the three mixture branches this PR added extracted to `foceiSMixScore()` / `foceiSMixForward()` / `foceiSMixCentral()` |
| `src/imp.cpp` Very Complex Method (66) -- `impComputeCov()` | FD Hessian bookkeeping moved to `impFdHessian()` |
| `test-vae-inner.R` `%>%` | two converted to `|>` (R >= 4.1 is already the floor, other test files use it) |
| `test-focei-mix-cov.R` unnecessary `c()` | dropped |
| `src/inner.cpp` Complex Method (16) -- `vaeLocalSearchL0()` | add / drop / swap split into `vaeLsAdd` / `vaeLsDrop` / `vaeLsSwap` over shared `vaeLsAddBlock` / `vaeLsDropBlock` |

The last row is pushed but **CodeFactor has not re-analyzed it** -- its status has not moved since `13:48:52Z` and it still reports `src/inner.cpp:22825-22870`, a range that no longer contains `vaeLocalSearchL0` (now at `:22888`). So the remaining count is stale rather than outstanding; it should clear on the next analysis.

## Verification

463 assertions green across `mix` (60), `saem-mix`, `vae-mix`, `focei-mix-grad`, `focei-mix-cov` (61), `npag-mixture`, `cov-decouple-saimp`, `cov-condition`, `saem-cov`, `nlmixr2output`, `vae-inner`. The `vaeLocalSearchL0` refactor is behaviour-preserving -- same neighborhoods, same fixed enumeration order (which is what makes the search deterministic) -- and 313 assertions across `vae-l0learn`, `vae-miqp` and `vae-cov-groups` pass.

PR #1060, opened 2026-09-10
fix(cov): PD-gate the standalone analytic covariance, and diagnose #1055

### LINKED ISSUE #1055: covMethod="analytic": block-Omega covariance is indefinite, giving NaN SEs for the off-diagonal (2 failing assertions in test-cov-analytic.R)
Found while running `test-cov-analytic.R` for unrelated work. Reproduced on a worktree that contains `origin/main` with no local changes, so this is not branch-specific.

## What happens

`test_that("block Omega is handled analytically, with off-diagonal covariance SEs", ...)` (`tests/testthat/test-cov-analytic.R:60`) fails two assertions:

- `test-cov-analytic.R:93` -- `expect_true(all(is.finite(r$se)) && all(r$se > 0))`
- `test-cov-analytic.R:94` -- `expect_true(all(eigen(r$cov, symmetric = TRUE, only.values = TRUE)$values > 0))`

`foceiCovAnalytic()` returns `method == "analytic"` and does produce an off-diagonal `cov.eta.v.eta.cl` row, but the covariance has one negative eigenvalue, so two variances come back negative and their SEs are `NaN`:

```
params: tka tcl tv add.sd om.eta.ka om.eta.cl cov.eta.v.eta.cl om.eta.v
se    : 0.095000 0.037540 0.023218 0.023974 0.079639 0.013051      NaN      NaN
cov diag: 0.00902491 0.00140922 0.00053906 0.00057473 0.00634239 0.00017032 -0.00021951 -0.00083603
                                                                  ^^^^^^^^^^^^  ^^^^^^^^^^^^
eigenvalues: 9.0488e-03 6.6327e-03 1.5766e-03 5.7593e-04 3.9418e-04 2.1955e-04 5.9264e-05 -1.5020e-03
```

The negative eigenvalue is not small relative to the rest of the spectrum -- it is larger in magnitude than three of the positive ones.

## Why this is a defect and not identifiability

The test deliberately replicates `theo_sd` four times (48 subjects) precisely so the block IS estimable; its own comment records that the un-replicated 12-subject data is genuinely indefinite under both the analytic and finite-difference routes, and that replicating it makes the full covariance positive-definite. So on this data the analytic engine is expected to yield a real off-diagonal SE, and the FD route was observed to. That points at the analytic block-Omega (E-basis) second derivatives rather than the data.

## Repro

```r
blk <- function() {
  ini({tka <- log(1.5); tcl <- log(2.7); tv <- log(31.5)
       eta.ka ~ 0.6
       eta.cl + eta.v ~ c(0.3, 0.03, 0.1)
       add.sd <- 0.7})
  model({ka <- exp(tka+eta.ka); cl <- exp(tcl+eta.cl); v <- exp(tv+eta.v)
         d/dt(depot) <- -ka*depot; d/dt(center) <- ka*depot - cl/v*center
         cp <- center/v; cp ~ add(add.sd)})
}
d0  <- nlmixr2data::theo_sd
dat <- do.call(rbind, lapply(1:4, function(k) { .x <- d0; .x$ID <- .x$ID + (k-1)*100; .x }))
fit <- nlmixr(blk, dat, "focei", foceiControl(sigdig = 4, print = 0L, covMethod = ""))
r   <- foceiCovAnalytic(fit)
r$se
eigen(r$cov, symmetric = TRUE, only.values = TRUE)$values
```

## Notes

The test carries `skip_on_ci()`, so CI never runs it -- which is presumably why this has gone unnoticed. `cov-analytic` is in `.slowBatches` batch 2 (weekly), and `skip_on_ci()` also suppresses it there, since the weekly runner sets `CI=true` too.

mattfidler:
Closes #1055.

## What #1055 turned out to be

The reported symptom -- `foceiCovAnalytic()` returning a covariance with one
negative eigenvalue and `NaN` SEs for the off-diagonal Omega element -- is not
the block-Omega E-basis second derivatives.

I built an independent brute-force central-difference Hessian of the FOCEI
objective and evaluated it at the issue's own estimates:

| | min eigenvalue of R |
| --- | --- |
| gold FD Hessian | **-671.23** |
| analytic engine | **-671.5** |

Both give the same two `NaN` SEs. **The negative eigenvalue is real.** At
`sigdig = 4` this fit stops at objective 435.22, against 432.37 at `sigdig = 5`
-- it is not at a local minimum, so the observed information there is
legitimately indefinite:

| sigdig | objective | min eig(R) |
| --- | --- | --- |
| 3 | 436.25 | -4.8e+03 |
| 4 | 435.22 | -6.7e+02 |
| 5 | 432.37 | +1.08e+02 |
| 6 | 433.92 | +1.07e+02 |

The issue notes the finite-difference route gives a real off-diagonal SE at the
same place. It does, but not because it disagrees: `covMethod="r,s"` is the
sandwich `Rinv S Rinv`, which is positive semi-definite by congruence whatever
R's definiteness is. It hides the indefinite R rather than contradicting it.

## The code change

One thing: **`foceiCovAnalytic()` gains the positive-definite gate the other
three analytic installers already have.** `.foceiInstallAnalyticCov()`,
`.covInstallResult()` and `.saemInstallAnalyticCov()` all refuse a non-PD
covariance; the standalone entry did not, so it could overwrite a usable `$cov`
with one carrying negative variances -- and once installed, that is what every
later `getVarCov()` returns. The result now carries `$pd`, the install is gated
on it, and a warning says why when it is skipped (on the cached path too, so a
second call is not silent).

## What this PR does NOT do, and why

An earlier revision of this branch re-solved the FOCEI EBEs to `Phi_eta = 0`
before forming R, on the grounds that the envelope/Schur data term is exact only
at the inner stationary point. That is **removed**. The EBEs are part of the
fit's solution: `fit$objf` was evaluated at them, and re-optimizing them inside
the covariance step yields the curvature of an idealized objective the reported
fit does not sit on. The evidence that appeared to support it did not: the gold
FD re-solves the EBEs at every perturbed parameter vector, so it encoded the
very assumption it seemed to confirm.

The EBEs are genuinely not stationary, and that belongs to estimation, not here.
Scoring the *same* stored EBEs across a tolerance ladder shows it is not an
artifact of the covariance solving tighter than the fit did -- it plateaus:

| solve tol | `max|Phi_eta|` (sigdig 4) | (sigdig 6) |
| --- | --- | --- |
| 1e-3 | 6.36e-03 | 4.39e-03 |
| 1e-6 | 2.30e-03 | 3.88e-04 |
| 1e-8 | 2.30e-03 | 3.85e-04 |
| 1e-10 | 2.30e-03 | 3.85e-04 |
| 1e-12 | 2.30e-03 | 3.85e-04 |

The residual tracks `sigdig`, i.e. the inner solver's own tolerance. The inner
problem itself is right: `src/inner.cpp`'s `lp` is the exact `d(llik)/d(eta)`
including the interaction terms. PR #1059 fixed a sibling instance of this the
correct way -- on the estimation side.

## A false record closed

`test_that("FOCEI analytic R matches the gold FD (block Omega)")` was skipped
with a comment recording 1.2%-377% disagreements as a known FOCEI assembly
defect. Those came from comparing a FOCEI fit's R against a gold built at a
**different (FOCE) fit's** estimates -- a convergence confound. At one common
point the two analytic branches agree to 1.3e-5 with identical eigenvalues. The
test is now a real gold-FD comparison rather than a skip.

## Tests

- the block-Omega SE test pins `sigdig = 6` (a point that IS a minimum) and
  additionally asserts `$pd` and that the covariance really was installed;
- a new test that an indefinite analytic covariance is returned to the caller
  but does not overwrite the fit's usable one, and that `setCov(fit, "analytic")`
  refuses it too;
- `skip_on_ci()` dropped throughout the file: it is in `.slowBatches`, and the
  weekly runner also sets `CI=true`, so those seven tests ran nowhere.

### Pins the trust inner default moved

`innerOpt` now defaults to trust, which lands the mu-referenced foce+ variants in
a different conditional basin. Re-measured:

| pin | old | new |
| --- | --- | --- |
| `foce+` / `focep` | 117.880675 | 112.320500 |
| `mfocep` | 111.731506 | 110.192593 |
| `ifocep` | 111.731506 | 110.152543 |

`focep` still reproduces `foce+` to every digit. `mfocep` and `ifocep` are no
longer identical, so each is pinned separately.

The `foce+` vs FOCEI analytic R comparison is not a stale pin and is not widened
blindly. At the same theta, the same omega, and EBEs agreeing to 6.9e-09, the two
kernels still differ: the FOCE/foce+ kernel uses the general total-derivative
form and carries `sum(gPhi . eta_ab)`, while the FOCEI kernel uses the
envelope/Schur form and drops it. With the stored EBEs at `|Phi_eta| ~ 2e-3`,
that dropped term IS the gap -- both kernels are right about their own
assumption, and it closes only when the inner solve tightens. The test now
asserts the standard errors, which is what a user reads, and keeps a raw-R bound
at the measured level with the mechanism recorded (raw R 1.2e-2 on the small
`[add.sd, tka]` entry; theta and sigma SEs to ~1e-5, the loosest SE to 1.1e-3).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -132,6 +132,24 @@
 
 ## Bug fixes
 
+- `covMethod="analytic"` now re-solves the FOCEI empirical Bayes estimates to
+  their own stationarity, `Phi_eta = 0`, before forming the observed
+  information.  The FOCEI data term is the envelope/Schur form
+  `Phi_pq - M_p' H^-1 M_q`, which is the total second derivative only AT the
+  inner stationary point, and a fit's stored EBEs satisfy that only to the inner
+  tolerance.  Its FOCE sibling already re-solved.  On the block-Omega `theo_sd`
+  fit the analytic observed information disagreed with an independent
+  brute-force central-difference Hessian of the same objective by 3.9e-3;
+  re-solving drops that to 7.7e-5, which is the finite-difference noise floor,
+  and reproduces the reference standard errors to every printed digit.
+
+- `getVarCov()` on a `focei` fit no longer installs an analytic covariance that
+  is not positive definite.  An outer optimizer that stops short of a local
+  minimum leaves an observed information with a negative eigenvalue, which
+  inverts to negative variances and `NaN` standard errors; the fit's existing
+  covariance is now kept and a warning says why.  The live `covMethod="analytic"`
+  seam, `setCov()` and the `saem` installer already guarded this.
+
 - A `focei`-family fit now reports whether its inner solves actually
   converged.  `fit$env$nTrustInner` breaks the `innerOpt="trust"` per-subject
   Newton solves down by outcome (`calls`, `error`, `notConverged`,
--- a/NEWS.md
+++ b/NEWS.md
@@ -143,12 +143,14 @@
   re-solving drops that to 7.7e-5, which is the finite-difference noise floor,
   and reproduces the reference standard errors to every printed digit.
 
-- `getVarCov()` on a `focei` fit no longer installs an analytic covariance that
-  is not positive definite.  An outer optimizer that stops short of a local
+- The standalone analytic-covariance entry point no longer installs a covariance
+  that is not positive definite.  An outer optimizer that stops short of a local
   minimum leaves an observed information with a negative eigenvalue, which
-  inverts to negative variances and `NaN` standard errors; the fit's existing
-  covariance is now kept and a warning says why.  The live `covMethod="analytic"`
-  seam, `setCov()` and the `saem` installer already guarded this.
+  inverts to negative variances and `NaN` standard errors -- and once installed
+  as the fit's `$cov` that is what every later `getVarCov()` returns.  The fit's
+  existing covariance is now kept and a warning says why.  The live
+  `covMethod="analytic"` seam, `setCov()` and the `saem` installer already
+  guarded this; the standalone entry was the one that did not.
 
 - A `focei`-family fit now reports whether its inner solves actually
   converged.  `fit$env$nTrustInner` breaks the `innerOpt="trust"` per-subject
--- a/NEWS.md
+++ b/NEWS.md
@@ -132,17 +132,6 @@
 
 ## Bug fixes
 
-- `covMethod="analytic"` now re-solves the FOCEI empirical Bayes estimates to
-  their own stationarity, `Phi_eta = 0`, before forming the observed
-  information.  The FOCEI data term is the envelope/Schur form
-  `Phi_pq - M_p' H^-1 M_q`, which is the total second derivative only AT the
-  inner stationary point, and a fit's stored EBEs satisfy that only to the inner
-  tolerance.  Its FOCE sibling already re-solved.  On the block-Omega `theo_sd`
-  fit the analytic observed information disagreed with an independent
-  brute-force central-difference Hessian of the same objective by 3.9e-3;
-  re-solving drops that to 7.7e-5, which is the finite-difference noise floor,
-  and reproduces the reference standard errors to every printed digit.
-
 - The standalone analytic-covariance entry point no longer installs a covariance
   that is not positive definite.  An outer optimizer that stops short of a local
   minimum leaves an observed information with a negative eigenvalue, which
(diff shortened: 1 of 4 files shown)


mattfidler:
This is not completely correct as written, the real reason for the change is the trust optimizer giving different solutions.

mattfidler:
Otherwise this does look correct, the fit is stationary with the same objects, just a better fit in general, which is something we have seen with the trust algorithm in general.

PR #1075, opened 2026-09-12
fix(focei): a stalled FOCE+ EBE polish is not a failed likelihood (#1069)

### LINKED ISSUE #1069: est="focep" (foce+) does not converge at the default sigdig=3: bobyqa stalls 4.8 objective units high
Found while fixing #1056.  Reproduced on `origin/main` with no local changes.

## What happens

A one-compartment additive `theo_sd` fit run as `est="focep"` at the DEFAULT
`sigdig` (3) stops at objective **121.560**.  The same model at `sigdig = 4` or
`6` reaches **116.804**.  `est="focei"` and `est="foce"` reach 116.824 at
`sigdig = 3`, so it is specific to `foce+`.

```
est     sigdig   objf        min eig(analytic R)
focep      3     121.56006   -307.9
focep      4     116.80376     26.62
focep      6     116.80341     26.61
focei      3     116.82393     24.95
foce       3     116.82393     24.95
```

The fit knows: `$runInfo` carries

```
last objective function was not at minimum, possible problems in optimization
ETAs were reset to zero during optimization
Hessian reset during optimization
```

and `$message` is "Normal exit from bobyqa".  The last five outer iterations
oscillate in the 5th digit of every parameter around objf 121.5601, i.e. the
search is stuck, not converging.

Because the stopping point is not a local minimum, its observed information is
legitimately indefinite (one eigenvalue -308), so `covMethod="analytic"` refuses
to install it and the fit falls back to `covMethod="r"` -- that is the PD gate
behaving correctly, not a covariance defect.  It is what surfaced this.

## Repro

```r
m1 <- function() {
  ini({ tka <- log(1.5); tcl <- log(2.7); tv <- log(31.5)
  model({ ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
    d/dt(depot)  <- -ka * depot
    d/dt(center) <-  ka * depot - cl / v * center
    cp <- center / v; cp ~ add(add.sd) })
}
f3 <- nlmixr(m1, nlmixr2data::theo_sd, "focep", focepControl(print = 0L, covMethod = "", sigdig = 3))
f4 <- nlmixr(m1, nlmixr2data::theo_sd, "focep", focepControl(print = 0L, covMethod = "", sigdig = 4))
c(f3$objf, f4$objf)   # 121.5601 116.8038
f3$runInfo
```

`foceiControl(interaction = FALSE, foce = "foce+", sigdig = 3)` reproduces it
identically (121.5601, 155 outer iterations), so it is the method, not the
`focep` wrapper.

## Note

`tests/testthat/test-cov-analytic.R`'s `est='focep' installs the full analytic
covariance` test hit this: it used the default sigdig and so asked for an
analytic install at the stalled point.  It now pins `sigdig = 4` and points
here; that pin should come back out once this is fixed.

mattfidler:
Fixes #1069.

## What was wrong

`est="focep"` (`foce="foce+"`) stopped 4.8 objective units high at the default
`sigdig=3` on `theo_sd`: objf **121.560** against **116.804** at `sigdig >= 4`,
with the omegas barely off their starting values, `$runInfo` reporting "last
objective function was not at minimum", and the observed information carrying a
real negative eigenvalue (min eig -307.9) so `covMethod="analytic"` refused to
install it.

The cause is not bobyqa and not the outer tolerances.  FOCE+ defines its EBE by
the truncated-score root with the live conditional R, and because the inner
optimizer minimizes a value whose reported gradient is that truncated score, its
stopping point is not exactly that root.  `refineFocePlusEta()` (added with the
analytic FOCE+ curvature) polishes it with a Newton iteration on the score.

That polish asked for `|score|_inf < 1e-9`, with a `norm < 1e-3 && 0 <=
decrement <= 1e-9` escape hatch.  Its backtracking line search accepts nothing
but a strict decrease in the score norm, so it stops exactly when the score
reaches the noise floor the solve tolerance buys -- at `sigdig=3` that floor is
near `1e-3`, above both gates.  Every subject that reached it had its likelihood
reported as `NA_REAL`.

Instrumented on the issue's own repro: **1933 refinements, 60 hard failures**,
all on one subject, stalling at `norm = 0.0023-0.0048` with `decrement = 2e-7`.
Those NAs are the cliffs that stopped the search -- outer iteration 7 probed
`omega(eta.ka)` upward and read **328.5** against a neighbourhood near 133, so
bobyqa's interpolation model learned that direction was catastrophic and never
moved the omegas again.

`sigdig=4` worked only because its floor sits a decade lower.

## The fix

The eta handed to the polish is the inner optimizer's own answer and is always
usable, so the polish is now best-effort.  Backtracking only ever accepts a
strict decrease, so the current point is the best one seen on every exit path;
each exit now keeps it and evaluates the likelihood there instead of returning
a failure.  The `focePlusSmallStep()` gate is gone -- a stall IS the stopping
criterion, detected rather than guessed at with constants calibrated to one
`sigdig`.

## Result

| | before | after |
|---|---|---|
| `focep` objf, sigdig=3 | 121.5601 (155 outer iters) | **116.8087** (78) |
| `focep` objf, sigdig=4 | 116.8038 | 116.8038 |
| min eig(analytic R), sigdig=3 | -307.9 (analytic refused) | **+27.4** (installs) |
| worst outer evaluation | 328.5 | 138.7 |
| `est="foce"` / `"focei"`, sigdig=3 | 116.8239 | 116.8239 |

`outerOpt="nlminb"` at `sigdig=3` moves from 119.42 to 116.8250 and
`outerOpt="lbfgsb3c"` from 116.8077 to 116.8598, i.e. the cliffs were derailing
more than bobyqa.

## Tests

- `test-focei-foce-plus.R`: new #1069 regression at the DEFAULT sigdig -- objf
  under 118, no outer evaluation more than 25 units above its own start (the
  NA-poisoned probes sat at ~300 against a neighbourhood near 130), and
  `omega(eta.ka)` actually off its start.
- `test-cov-analytic.R`: the `est='focep'` analytic-covariance test drops the
  `sigdig = 4` pin it carried while #1069 was open, as that issue asked.

Green locally, 1065 assertions over 16 files: `cov-analytic`,
`cov-analytic-defaults`, `focei-foce-plus`, `focei-fast-methods`,
`focei-fast-methods-fit`, `mfocei`, `foceiLik`, `focei-ptrs`,
`focei-foce-stride`, `focei-family-control`, `focei-fast-grad`,
`agq-fast-grad`, `focei-full-inner`, `focei-outer-trust`,
`focei-ll-fast-grad`, `lik-contrib-fast-grad`, `fitSim`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,17 @@
   own; they now agree.  On an additive model, where the two methods must
   coincide, the `foce+` and FOCEI observed informations agreed to 1.2e-2 at the
   same EBEs and now agree to 1.1e-13 (#1056).
+- `est="focep"` (`foce="foce+"`) converges at the default `sigdig`.  FOCE+ polishes
+  the inner optimizer's eta onto the truncated-score root it defines its EBE by, and
+  that polish stops once the score reaches the noise floor the solve tolerance buys --
+  near `1e-3` at the default `sigdig=3`, well above the `1e-9` it asked for.  A
+  subject that stalled there reported its likelihood as `NA`, which the outer search
+  read as a cliff: on `theo_sd` it stopped 4.8 objective units high (121.560 against
+  116.804) with the omegas barely off their starting values, and the observed
+  information at that point had a negative eigenvalue so `covMethod="analytic"`
+  refused to install it.  The polish is now best-effort -- the eta it was handed is
+  the inner optimizer's own answer and is always usable, so every exit path keeps the
+  best point found and evaluates the likelihood there (#1069).
 - `saem` refuses a model whose random effect has no population parameter of its
   own -- added to none, or sharing one with another random effect -- naming the
   random effects, instead of fitting the model without them and then failing
--- a/src/inner.cpp
+++ b/src/inner.cpp
@@ -3758,10 +3758,6 @@ static bool focePlusRefinementRequired() {
       op_focei.freezeOde || op_focei.neta == 0);
 }
 
-static bool focePlusSmallStep(double norm, double decrement) {
-  return norm < 1e-3 && decrement >= 0 && decrement <= 1e-9;
-}
-
 static bool focePlusScore(arma::vec &at, arma::vec &out, int id) {
   auto *ind = &inds_focei[id];
   ind->setup = 0; ++ind->nInnerF; ++ind->nInnerG;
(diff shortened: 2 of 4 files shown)


mattfidler:
### Second-reviewer follow-up (Antigravity / gemini-3.1-pro-high)

Three findings; all triaged by measuring rather than arguing.

**1. "`x` is the best point seen on every exit" overclaims.** Real, as a
comment defect. `focePlusScoreJacobian()`'s `x +/- h` probes are points seen,
and one can score lower than `x` while its mirror probe fails. They are
derivative samples at a fixed step, not candidate steps, so adopting one would
be a different algorithm with nothing behind it. The comment now says "best
ITERATE the search accepted" and spells out why the probes are excluded.

**2. "The assemblers drop a score they assume is zero, so the covariance is
silently wrong."** The mechanism is real; the magnitude is what the tolerance
buys. Measured against `test-cov-analytic.R`'s own gold FD (EBEs re-solved to
`|S| < 1e-12`, `tol = 1e-12`):

| sigdig | max rel. err, significant entries of R | max rel. err, SEs |
|---|---|---|
| 3 (default) | 2.7e-3 | 2.5e-4 |
| 4 | 1.0e-4 | 1.4e-5 |
| 6 | 5.9e-6 | 3.1e-6 |

It degrades smoothly with the solve tolerance, as it should. And the
counterfactual is not a better covariance -- before this fix the same fit had
*no* covariance (min eig -308, the PD gate refused). Recorded in the test.

**3. "The new test passes even if the polish is deleted."** Correct about that
test, but the polish is pinned elsewhere, and I measured that rather than
assuming it. Re-running the sigdig=6 gold-FD comparison with
`refineFocePlusEta()` disabled gives **0.252** relative and a **7.5%** SE gap,
and one direction stops being identified -- against 5.9e-6 / 3.1e-6 with it. So
`test-cov-analytic.R`'s existing 3e-4 bound catches a deleted polish by three
orders of magnitude. Noted there so the next reader does not have to re-derive
it.

No OpenMP or state/lifetime findings, which matches my own reading:
`focePlusBacktrack()` only ever replaces `x` with a strict improvement, and
`ind->setup = 0` plus the final `likInner0()` re-solve leaves the subject's
cached solve consistent with the eta written back.

### Separately: cost

The polish runs on every likelihood evaluation of every subject, at ~7 ODE
solves per Newton iteration -- 47,522 score solves on one `theo_sd` fit. That
is why `est="focep"` takes **6.6 s** where `est="foce"` takes **1.7 s** and
`est="focei"` 2.3 s on the same model. Out of scope here (narrowing where it
runs would change the objective's definition), but worth its own issue.

mattfidler:
Correction to the "Separately: cost" section of my previous comment. Those
timings (6.6 s vs 1.7 s) were taken while an unrelated job was saturating the
machine, so they are inflated and the ratio is wrong. Clean, medians of 3 runs
in a steady-state process, model cache warm:

| | wall | outer iters |
|---|---|---|
| `est="focep"` | 1.20-1.35 s | 78 |
| `est="focep"`, polish disabled | 0.50-0.52 s | 60 |
| `est="foce"` | 0.55 s | 60 |
| `est="focei"` | 0.51 s | 60 |

So the gap is ~2.5x, not ~4x, and the polish does account for all of it --
disabled, `focep` costs what `foce` costs. Counters over that one fit: 1072
refinements, 4144 Newton iterations, 23,112 score solves (the earlier 47,522 was
from the pre-fix *stalling* run, which took 155 outer iterations).

Filed as #1078 with the numbers and some directions worth measuring. Nothing in
this PR changes.

PR #1082, opened 2026-09-12
One pool model for the FOCEi inner problem and the analytic outer derivatives

hiddevandebeek:
Follow-up to #1049 (the outer gradient "needs to be in the ode pool so that it can be used efficiently without swapping out memory").

## What this does

`fast=TRUE` now carries the augmented outer model's sensitivities on the **inner model itself**, the way the `combSens` (#958) build does for impmap's theta sensitivities. The solve pool then holds **one** model and nothing is swapped during a fit:

- The outer lines (second-order and non-mu theta sensitivity states, `rx_f1_`/`rx_f2_`/`rx_rvar*`/`rx_rsig*` columns) are appended to the inner model inside `if (rxOuterBlock == 1) { ... }` (`.foceiMaybeAddOuterLines`). The outer slot registers the same compile (matched by parsed md5) and `OdeSwapEsBatch` is a no-op for it, so the event-sensitivity shape is installed once for the whole fit and never switched.
- Inner iterations integrate only the first-order block of that model (`ind->neqOverride`, outer block switched off). The width includes every first-order sensitivity so rxode2's dosing ("jump") sensitivities stay intact.
- The derivative passes (the gradient's re-optimization and the Hessian's) run one full-width solve per subject at the EBEs; the analytic outer gradient and the outer Hessian's base solve **read** that solve (`outerSolveCached`) instead of solving the augmented model themselves. Only the Hessian's third-order probes still integrate.
- A plain objective evaluation stays compact, so `fast=TRUE` evaluates the same function as `fast=FALSE` (identical objective at a fixed point, gradient identical to the separate build).
- `rxOuterBlock = 1` is declared with a default inside the model, so the table/residual/R-covariance solves of the combined model run the full model unchanged and every peer model's positional `param()` layout is untouched.

`foceiControl(outerCombine=FALSE)` keeps the separate outer model. `fit$env$outerComb`, `combInnerNeq`, `nInnerSolveCompact`/`nInnerSolveFull`, `nOuterSolveReused`/`nOuterSolveRun` report what ran.

## Depends on rxode2

Compacting a model against its own wider code needs rxode2's generated `dydt` to skip the derivative lines beyond `_neq[0]` and every generated function to read a state beyond the compacted width as 0 (nlmixr2/rxode2#1366, `rxode2::rxDydtCompact()` probe). Without that rxode2 the fit keeps the separate outer model, so this is safe to merge either way.

## Measured (theo_sd, single-thread CPU, same build, `outerCombine` FALSE vs TRUE)

| | separate | combined |
|---|---|---|
| objective at initial estimates | 133.5736 | 133.5736 |
| nlminb fit total / optimize | 1.81 / 0.56 s | 1.62 / 0.53 s |
| full-width outer solves integrated | 672 | 576 |
| gradient | -0.8900923 ... | identical |

Deterministic under `cores=2` (one real bug fixed on the way: a reused subject has to be bound to the reading thread before its lhs slice is taken, or two subjects last solved by one inner thread share a slice across outer threads). Setup is ~0.2 s heavier per fit because rxode2 re-derives the event-sensitivity map of the (now second-order) inner model on every model construction -- a candidate for caching on the rxode2 side.

## Tests

New `test-focei-outer-comb.R` (build shape, md5 identity, reuse and compaction counts, agreement with central differences and with the separate build, full-fit optimum). Passing: fast-grad, agq-fast-grad, outer-hessian-*, outer-trust, foce-plus, full-inner, inner-conditional, outer-fd-fallback, ll-fast-grad(-fit), lik-contrib-fast-grad, ptrs, warm, reproducible, prior, agq-cov, dde-focei, mtime-fit, cens, cwres, npde, addCwres, cov-analytic (except the two `focep` failures that are pre-existing on `main`).


hiddevandebeek:
@mattfidler is this sort of what you meant? It comes together with some rxode2 changes. If this is in the right direction i will keep digging a bit to make it nice and efficient :)

mattfidler:
No.  This is a bit interesting but I meant load the model and then allow swapping in C++.  The ode pool is the method of swapping between methods but allowing the underlying problem to be allocated once.  If you want we can set up a meeting next week to discuss

hiddevandebeek:
Closing -- not the design meant; will follow up with the odeSwap-based version.

PR #1087, opened 2026-09-13
fix(saem): make the sampling RNG independent of the solve order; guard the ordId mapping

billdenney:
Companion to `rxode2:fix-sortids-throttle-gate`. **These must merge together** — see the bottom.

### 1. SAEM's sampling RNG was keyed to the solve order

rxode2's `par_*()` loops re-seed the per-thread threefry engine once per subject:

```c
int _id = rx->ordId[solveid] - 1;
setSeedEng1(seed0 + _id);
ind_liblsoda0(rx, op, opt, _id, ...);
```

The seed is keyed by subject id — that part is right. But when the solve returns, `_eng[thread]` holds the seed of **whichever subject that thread happened to solve last**, and that is decided by the solve order, i.e. by `rx->ordId`.

SAEM's sampling draws read that same engine: the `do_mcmc` proposals seeded by `_saemSeedDoMcmc()`, and `augmentCensY()`'s `rxTruncNorm()` through `rxNormEng()`/`rxUnifEng()` (`truncNorm.h`). The sampling *seeds* are keyed properly (fit seed folded with kiter/method/chain/endpoint/mixIdx), but the engine they draw from was not.

`nmMcmcRng.h` was written for exactly this, and **2 of the 18 `user_fn()` call sites used it**. The other 16 did not.

Take the restore inside `user_function()` instead, as a destructor so an early return or a throw still restores, and so a nineteenth caller cannot reintroduce the gap. (`imp.cpp` already had this discipline — it seeds per `(iter, id)`, restores after the inner solves, and re-seeds immediately before the SIR draw. SAEM was the outlier.)

**Validation** — 131-subject fit, seeded, 4 threads:

| | result |
|---|---|
| gate as shipped (identity order) | 6/6 runs identical — the contamination is *stable* here, which is why nobody saw it |
| gate corrected, before this PR | **4 distinct objf over 8 runs**; a different value again for each of identity / cost-desc / cost-asc / random / rotation |
| gate corrected, with this PR | **8/8 identical**, and **bit-identical across all five orders** |

Ruled out along the way: not the M3/censoring path (uncensored is equally affected), not chain sharing (`nmc=1` too), not the solve itself (a plain `rxSolve()` is bit-identical under a permuted order).

### ⚠️ This changes SAEM results

The fit above goes **−5728.8317 → −8440.6207**. Today's numbers are reproducible but come from the contaminated stream, so they move. The new value is the order-independent one; the old one was an artifact of which subject a thread solved last. Please sanity-check against your own reference fits before merging.

### 2. Guarded the `ordId` position → subject mapping

`rx->ordId` is a permutation of the `nsub*nsim` subject-**solves**. All 20 of the per-subject parallel loops in `inner.cpp` walk **subjects** and bound themselves by `getRxNsub(rx)`, so what they read out of it means what they take it to mean only when `nsim == 1`. That holds for every FOCEi path today, but nothing said so and nothing checked it. At `nsim > 1` the first `nsub` entries are an arbitrary subset of `1..nsub*nsim`: some subjects visited twice, others skipped, and `inds_focei`/`ebes`/`Hb` indexed past `nsub` — subjects silently swapped, the shape of the bug that made rxode2 give `solveid` a single meaning across ~100 `ind_*()` drivers (nlmixr2est issue 1020).

Every walk now goes through `foceiOrdId()`, which falls back to the canonical order when `nsim != 1`. Unordered is slower; mixed up is wrong.

I reviewed the mapping at all 20 sites while doing it: each maps a position to an id exactly once and then uses the id for the solve **and** for every per-subject array it writes, and the ids handed across to rxode2 (`ind_solve`, `iniSubjectE`, `odeSwapSolveInd`, `calc_lhs`, all via `getRxId`) are ids, never positions — so there is no double mapping at the seam. Verified end to end: a FOCEi fit is bit-identical (objf −6349.6023) under identity, cost-descending, cost-ascending, a random permutation and a rotation.

### Tests

Both new test files assert their convention against the sources, the way rxode2's `test-ind-solve-subject-id.R` does for its half, because ~20 near-copies of one line is how this regresses. Both are mutation-tested — they go red for a raw `getOrdId()` call, for reusing the loop position after mapping it (the subject-mixing bug itself), for a loop bound that is not a subject count, for dropping the `nsim` fallback, and for removing the seed guard.

### Merge order

`rxode2:fix-sortids-throttle-gate` makes `rx->ordId` a real permutation for the first time; without this PR that alone makes SAEM non-reproducible. Merge both, or neither.


mattfidler:
@billdenney I will have to wait for this since I don't see `fix-sortids-throttle-gate` in rxode2

billdenney:
It was merged in https://github.com/nlmixr2/rxode2/pull/1369

mattfidler:
Actually, @billdenney this may be the wrongsolution.  The rng needs to be seeeded so that it independent of which item is solved, just like we do for `rxSolve`.  This does show a bug, but the solution is wrong.  It also shouldn't need a rxode2 change.

Reviewed it some, it seems to be keyed, so it may be OK, but there needs to be some more care taken to make sure these are independent.  It isn't immediately obvious the seeds are independent

mattfidler:
Let me review this a bit more to see if the implementation makes sense.  It should not be independent of the order that the ids are entered.

mattfidler:
These seeds are not correctly seeded based on the threefry setup.  It assumes seed properties that threefry doesn't has and risks collisions.

billdenney:
Well shoot. The goal (as is probably obvious) is to make all methods consistently use a single code path for random number generation and rxode2 interaction with random numbers consistently pulled regardless of number of threads or subject order. By pulling them all together, the hope was issues that have now been found and solved a couple of time will now get a single path to be solved once.

Sorry for extra work here, and hopefully this at least gets closer to the solution.

mattfidler:
No, this is claude mis-understanding threefry.  It has consistently done this because other RNGs don't behave the same way.  It will get there.

mattfidler:
## Review of the RNG seeding, plus the changes made

Reviewed for three properties: (1) the RNG is seeded for each individual and each step, (2) every seed is independent, (3) every seed is in order and can be computed from its position. **None held on the branch as submitted.** The fixes are below. After them, an Antigravity (Gemini) review ran until a round came back clean.

### What was wrong

- **The phi1 and phi0 MCMC blocks shared one seed.** `_saemSeedDoMcmc` folded `(kiter, method, u, k1)` through a multiply hash with no block key. So every iteration the phi0 proposals reused the phi1 noise, and for method 3 the same acceptance uniforms.
- **Neighbouring mixture components collided.** The hash ended in `s += mixIdx`, so component m at `k1` shared a seed with component m+1 at `k1 - 1`.
- **Hashed seeds carry birthday collisions** that sequential threefry keys never need. Distinct keys are already independent streams.
- **The `user_function()` seed guard added in this PR changes no draw.** Every draw site already reseeds immediately before drawing. It is harmless, but it is not what fixes solve-order dependence.
- **`arma::randu` in `npde.cpp` drew from R's RNG.**
- **FOCEi `etaRestart` used a hashed seed** (`inner.cpp`).
- **The macOS build failed on this branch** (`nlm.cpp`: `redefinition of omp_get_num_procs`). `src/rxomp.h` had no include guard, and `nmParallelCore.h` now includes it twice.

### Changes

- **`4803089a5`, `396cce3ba`: sequential, closed-form seeds** (`src/nmSeqSeed.h`). Item `i` of a step (a chain row, subject, observation or support point) is seeded with `seed + offset + i` immediately before its draws. `offset` depends only on the iteration and step. Nothing is hashed and nothing is a running counter, so a fit resumed at an iteration draws exactly the seeds it would have.
  - **SAEM** (`saemSeedLayout`): each iteration is every mixture component's MCMC steps (phi1 then phi0, methods 1/2/3, 20x `nu` at kiter 0), one seed per chain row, then the censored-value draws, one seed per observation.
  - **npb** (`seedBase`): each sweep owns `2*nsub + nMix + 2*K` seeds.
  - **FOCEi `etaRestart`** (`6c6f6a6a9`, `396cce3ba`): restart `k` of subject `id` is seeded with `seed + id*nres + k`.
  - Each sampler owns an exactly computed block `[seed, seed + reserved)`. rxode2's per-subject solve seeds then continue from `seed + reserved` (`nmSeqSeedStart`). This also removes the thread-count-dependent advance from the setup `rxSolve_()`.
- **`498858244`: npde** uses threefry, one seed per observation (`tableControl(seed=)` + row), drawn serially before the parallel loop.
- **`11a41c8eb`:** include guard on `src/rxomp.h`. The macOS build passes.
- **`068c6f795`:** seed-layout test hook split into helpers (a habit-hooks complexity finding).
- **Tests**
  - Re-pinned (`6339bece8`, `9ac0c72b6`).
    - The odeswap zero-blast-radius fit was re-pinned. The new value is bit-identical on a repeat run and at 2 threads.
    - The npb R-hat test is now bounded below by `sqrt((n-1)/n)`, the plain estimator's true minimum, instead of 1; the new seeds give 0.997.
  - New thread-count tests (`6d84af4fd`, `53c76ead2`).
    - SAEM, including censored (M3) data.
    - npde (`addNpde()` at 1 and 2 threads).
    - npb.
    - FOCEi `etaRestart`: it asserts `omegaRestart > 0` first, so the restarts are known to have run.
  - A layout test requires every draw's seed to be `0..n-1` in draw order, and a shorter run's seeds to be a prefix of a longer run's.
- **`d258595d5`:** merged `origin/main`. The only conflict was `NEWS.md`, and both entries were kept.
- **Seeded SAEM, npb and FOCEi-with-restart results change.** They are still reproducible, and identical across thread counts.

### Antigravity review findings

| Round | Finding | Outcome |
|---|---|---|
| 1 | `saemSeedLayout::cens` overlaps the MCMC seeds | Rejected: quoted a formula that is not in the file. The censored-value block starts after `nComp*steps*nM`. |
| 1 | `R::qnorm`/`ISNA` inside npde's OpenMP loop | Rejected: nmath does not report `ME_DOMAIN` errors (NaN, no warning), and `ISNA` is a pure macro. Both predate this PR. |
| 2 | `step()` ignores `k1` for methods 1/2 | Rejected: both MCMC loops `break` after `k1 = 0` for method < 3. |
| 2 | `imp.cpp` `nmStatic` without `nmNoMap` | Rejected: the body writes only per-subject slots, SIR seeds are keyed by `(iter, eid)`, and accumulation is serial. |
| 2 | No test that a resumed fit matches | Covered by the layout prefix test. There is no resume entry point to test directly. |
| 3 | `augmentCensY` gets a 1-based component index | Rejected: all three callers pass 0-based `mHyp`/`jMix`/`-1`. |
| 3 | No thread-count tests for npb, npde, `etaRestart` | **Fixed**, `6d84af4fd`. |
| 4 | npde error handling differs between one thread and several | Rejected: the loop body has its own `try`/`catch` that sets NA. |
| 4 | SAEM thread test has no censored data | **Fixed**, `53c76ead2`. |
| 5 | none | Clean. |

A separate Claude review of the seeding commits also found no defects.

### Verified locally

Every run below was on a scratch build of this branch.
- **Thread counts:** SAEM, npb, npde and FOCEi `etaRestart` give identical results at 1 and 2 threads. npde is also bit-identical at 1 and 4 threads, with and without ties, and `vpcSim()` is identical at 1 and 4 cores.
- **Test files passing:**
  - `test-saem-mcmc-seed-order` 25/25
  - `test-npde` 22/22
  - `test-focei-trust-inner` 113/113
  - `test-npb-fit` 31/31
  - `test-npag-npb-cens-978`, `test-saem-logitnorm-probitnorm-resid`, `test-saem-cov-sa`, `test-iov-same`, `test-saem-loglik-focei`, `test-mix`, `test-saem-odeswap-regression`
- **Not run locally against the final seed-reservation commit:** the other FOCEi pinned-value files (`test-mtime-fit`, `test-impmap`, `test-cov-analytic`, `test-restart`, `test-focei-foce-plus`). CI covers them.
- **CI on the merge commit:** macOS, Windows, devel, oldrel-1, release shard 3/3, pkgdown and CodeFactor passed. Shards 1/3 and 2/3 were cancelled by the newer push. The run on `53c76ead2` is in progress.

PR #1088, opened 2026-09-13
perf(saem): skip three optimizations whose results the SA covariance phase discards

billdenney:
## What this is

With `covMethod="sa"` (the default) and default `nSaCov=500`, `saem_fit()`'s
iteration loop runs `niter + nSaCov` iterations, but printing is gated on
`kiter < niter`. So a default `saemControl(nBurn=200, nEm=300)` fit runs **1000
iterations and prints only the first 500** — the iteration counter appears to
stall at 500 while the fit keeps running. That second half is the dedicated SA
covariance phase, which freezes theta at `theta_hat` by appending `nSaCov` zeros
to the gain sequence `pas` and lets only the MCMC E-step resimulate phi.

Three optimizations keep running throughout that phase, and all three have their
results written back through the frozen gain:

```cpp
x = x + pas(kiter) * (opt - x)    // pas(kiter) == 0  =>  x unchanged
```

The worst of them re-solves the ODE over all `N*nmc` rows on every objective
evaluation, ~500 times over, for nothing.

This PR skips each of them under the exact condition that makes the result dead,
without changing any value that is written.

## The three sites

| commit | site | reachable today | measured effect |
|---|---|---|---|
| `7887a86` | `refinePhi0Lik` | yes, by default | cov phase −58…−62%, `saem` −29…−39% |
| `eee2a05` | `refinePhi1Lik` | no — defensive | none (inert) |
| `752fdba` | 8 × `_saemOpt` residual branches | yes, for non-`add`/`prop` error | `saem` 16.1 s → 15.0 s isolated |

All three key on the same predicate, `pas(kiter) == 0.0`. That is exact rather
than phase-indexed: `pas` is `1/k^a`, never exactly zero for finite `k`, so the
only zeros in it are the ones `nSaCov` appends. The predicate names the actual
condition under which the work is dead, not a phase number.

**1. `refinePhi0Lik`** — the dominant cost. It runs every iteration because
`nonMuTheta` defaults to `"regress"` with `nonMuThetaEvery=1`, and every
objective evaluation is a full ODE re-solve. Instrumented at 5 threads it was
30.1 s of a 49.4 s cov phase. `xmin` is taken as `par0` (`== cur`), so the
closing `fill(cur + 0*(xmin[c] - cur))` is exactly `cur`; the MCOV0 back-solve
and the fixed-entry restore still run.

**2. `refinePhi1Lik`** — same structure, same closing write. **Currently
unreachable**: `_saemPhi1PoolReady` implies a general-likelihood fit, and
`R/saem.R` forces `nSaCov=0` for those, so `pas` is never zero while it runs. I
could not construct a case that exercises the new branch, and it is verified only
as "changes nothing". It is a guard against that coupling changing, and it keeps
the two refinements consistent instead of leaving one silently dependent on an
unrelated file's decision.

**3. Residual-error optimizers** — deliberately narrow. The `switch` cannot be
skipped wholesale: `rmAdd`/`rmProp` write `ares`/`bres` unconditionally rather
than through the gain, `rmProp` mutates `sig2` (which feeds the unconditional
`sigma2[b] = sig2`), and the `resFixed` pins fire before the optimizer. Only the
`_saemOpt` call is skipped, via one `_saemOptOrSkip()` helper that seeds `pxmin`
from `_saemStart` — the same values `_saemOpt` itself starts from — so all eight
branches go through one place and cannot drift apart.

## Verification

Bitwise identical against a **pristine `origin/main` build in a separate
library**, not against an earlier build of this branch. "Bitwise" means
`identical()` on fixed effects, Omega, individual etas, parFixed estimates, SEs
and the full covariance matrix — zero difference, not a tolerance.

- reference model (2 mu-referenced etas + 2 non-mu thetas): identical at 1, 5 and 16 threads
- combined-error variant (exercises fixes 1 and 3): identical
- no-phi0 combined-error model (isolates fix 3): identical

Results are also bitwise identical across thread counts both before and after,
which is what makes the A/B meaningful.

Regression sweep over all SAEM/non-mu test files, before vs after:

| | files | assertions | fail | err | skip |
|---|---|---|---|---|---|
| `origin/main` | 36 | 861 | 0 | 1 | 0 |
| this branch | 36 | 861 | 0 | 1 | 0 |

The single error is pre-existing and unrelated: `test-saem-degraded-fit.R` uses
`testthat::local_mocked_bindings()`, which needs `pkgload::load_all()` and aborts
in `testthat:::dev_package()` before fitting anything when run against an
installed library.

## Timings

Linux, 32-core, R 4.6.1, rxode2 at `origin/main`. Model: 1-compartment IV with an
Emax-over-time central volume, 89 subjects / 1335 observations,
`nBurn=200, nEm=300, nSaCov=500`. Arms interleaved in one binary via a temporary
toggle, 2–3 reps each, means shown.

| threads | cov phase before | after | `saem` before | after |
|---|---|---|---|---|
| 1 | 101.1 s | 40.5 s (−60%) | 172.2 s | 112.2 s (−35%) |
| 5 | 53.5 s | 20.5 s (−62%) | 92.7 s | 56.3 s (−39%) |
| 16 | 37.4 s | 15.8 s (−58%) | 64.1 s | 45.3 s (−29%) |

The cov phase falls from 58% of the loop to 35%. Estimation-phase time is
unchanged within noise, as expected — the guard is false throughout it.

Fix 3 measured alone, on a model with no phi0 parameters so the phi0 refinement
returns immediately and cannot confound (5 threads, 3 paired reps): `saem`
16.1 s → 15.0 s, the fixed arm lower in every pair. Smaller than the phi0 saving
because that optimizer scores precomputed predictions and never re-solves the ODE.

## Not a parallelization change

Worth recording, since that was the first thing looked at. The remaining cov-phase
cost is the MCMC E-step, and it is already parallel where it legitimately can be:
`user_function` parallelizes the solve over all `N*nmc` rows, so every MH proposal
already evaluates all subjects in parallel. `do_mcmc`'s own loops over MH sweeps
and phi coordinates are a sequential Metropolis-within-Gibbs chain — sweep `u+1`
proposes from the state accepted at sweep `u` — so parallelizing them would change
the sampler and therefore the results. Measured scaling of the post-fix cov phase
is 40.5 s → 20.5 s → 15.8 s at 1/5/16 threads, already past the useful point at
this problem size. The win here is redundancy removal, not parallelism.

## Commit layout

Five commits, one fix each, plus the measurement scaffolding kept in history:

```
752fdba7b perf(saem): skip the residual-error optimization when its gain is frozen at zero
eee2a05b6 perf(saem): skip the phi1 refinement when its gain is frozen at zero
4b5c73b1a Remove the temporary SAEM phase instrumentation
7887a8688 perf(saem): skip the phi0 refinement when its gain is frozen at zero
85c2301b7 TEMP: instrument saem.cpp iteration phases for timing
```

The first commit adds env-gated (`NLMIXR2_SAEM_TIME=1`) phase timers and the third
removes them again; `git revert 4b5c73b1a` brings them back to re-measure. The
removal was done by restoring `src/saem.cpp` from `origin/main` and re-applying
only the fix hunk, so there is no scaffolding residue.

Net diff: `src/saem.cpp` only, +56 −18.

## Scope caveat for reviewers

The size of the win depends on the model. `refinePhi0Lik` returns immediately when
`nphi0 == 0`, so a fit whose thetas are all mu-referenced sees nothing from fix 1 —
its cov phase is pure MCMC already. The cheap way to size this on any given fit is
to time `saemControl(nSaCov=0)` against the default; the difference is the whole
unprinted phase.


mattfidler:
## Update: phi1 skip dropped, skip counters + test added, merged with main

### The `refinePhi1Lik` skip is removed (1fb4396c5)

`refinePhi1Lik` is back to exactly what `origin/main` has. That skip could never fire, and nothing could test it:

- `refinePhi1Lik` only runs when `_saemPhi1PoolReady` is true. That needs the phi1-inner models that `R/saem.R` attaches only when `.saemGeneralLik(ui)` is true (`R/saem.R:209`).
- For those same general-likelihood fits, `R/saem.R:283` forces `nSaCov = 0L`, because the SA covariance phase result is thrown away for them.
- The frozen-gain condition `pas(kiter) == 0` only happens during the `nSaCov` iterations, since `pas` is `1/k^a` plus the `nSaCov` zeros. So `refinePhi1Lik` and a zero gain never happen in the same fit.

I checked this on a real fit rather than just reading the code. A counter on the phi1 skip stayed at 0 for every `covMethod="sa"` fit, and `saemPhi1RefineN_()` did not move for a normal-error fit. If general-likelihood fits ever get an SA phase, that change should add the skip along with a test that exercises it.

### Test that the skips happen (111719095)

An antigravity review found no correctness problems. It did point out a gap: skipping a result that is thrown away changes no output, so every test would still pass if the skips were removed. To close that:

- `saemGainFrozenSkipN_()` returns `c(phi0=, resid=)`, the number of optimizations skipped because the gain was frozen at zero. It is registered in `src/init.c`.
- A new test in `test-saem-cov-sa.R` fits a model with a non-mu-referenced `tka`, `nonMuTheta="regress"` and add+prop error. It checks that a `linFim` fit skips nothing, that an `sa` fit with `nSaCov=15` skips exactly 15 phi0 refinements and 15 residual optimizations, and that both fits give the same thetas.

A second antigravity round on the updated diff found no issues. Locally, `test-saem-cov-sa.R` (125) and `test-saem-phi1-inner.R` (20) pass.

### Merged with `origin/main`

This brings in #1094's `test-predict.R` fix, which was the original shard 1/3 failure.

### Note on the PR description

The description above still describes three skip sites. Only two remain: `refinePhi0Lik` and the residual-error `_saemOpt` calls. The timing numbers are unaffected, because the phi1 skip never ran.

Issue #1068, opened 2026-09-11
Refined Full Conditional Hessian for inner problem should be used for determinant?


mattfidler:
- Consider if the inner Full Conditional Hessian should drive the derivative as well.  It makes sense to me.

@hiddevandebeek  this is based on your work, can you explore if this makes sense too.

hiddevandebeek:
@mattfidler Yes I will tomorrow

PR #1090, opened 2026-09-15
feat(focei): full conditional Hessian in the Laplace determinant (detHessian="conditional")

hiddevandebeek:
Explores #1068: "should the refined full conditional Hessian be used for the determinant, and drive the derivative as well?"

## What this does

`foceiControl(detHessian = c("focei", "conditional"))`, default unchanged.

- **Objective**: with `"conditional"`, `LikInner2`'s `log|H|` uses the full conditional inner Hessian (the `innerHess2` model's exact `d2pred`/`d2r`, through `calcEtaHessian`'s existing conditional branch) instead of FOCEI's Gauss-Newton expected information. That is the Laplace approximation proper (observed information at the conditional mode, as NONMEM `LAPLACE`). `innerHessian="conditional"` previously only changed the inner optimizer's curvature; the two are now independent switches on the same model.
- **Derivative**: the analytic outer gradient carries the matching third-order term. It reuses the outer Hessian's eta probes (`foceiHessianThird`, which gained a `points` argument: the gradient uses a 2-point central difference, the Hessian keeps its 4-point Richardson rule) and the (f,R) density chain that was inline in the R-matrix kernel, extracted as `foceiDensityChain()` and shared by `foceiRSubjectFR_` and the new `foceiGradSubjectFullFR_`.
- The probe solve itself goes through the outer Hessian's `FoceiHessianModel` (which now takes its `FoceiHessianProbes` directly instead of the Hessian call object) and `foceiHessianDirs()` builds the direction map for both consumers, so the only new code is the per-subject kernel that assembles `g = 2 dPhi + tr(H^-1 dH*)` with `Ht = H` and `dHt = Tn`.
- The analytic covariance and the analytic outer Hessian differentiate the Gauss-Newton determinant, so both decline to their finite-difference fallbacks under the option. FOCE, AGQ and an estimated boxCox/yeoJohnson lambda decline the analytic gradient (FD gradient) for now.

## Numbers (theo_sd, 1-cmt ODE, single thread)

At the FOCEI estimates: FOCEI 116.81, full-determinant Laplace 118.54, AGQ-11 (~exact marginal) 118.13 -- the full determinant is 0.4 above the exact value where FOCEI is 1.3 below.

Fit: FOCEI 1.6 s, `detHessian="conditional"` 2.8 s (18 analytic gradients, no declines), converged objective 118.5085.

Verification: a 1-eta closed-form model matches `2*obj + log|d2obj/deta2| + log|Omega|` to 1e-6; the theo_sd analytic gradient agrees with central differences within 0.5 %; the small-model gradient matches its central difference to 7 digits and vanishes at the converged fit; with M3-censored observations the gradient still matches its central difference (`test-focei-det-conditional.R`). Related test files (`focei-inner-conditional`, `focei-full-inner`, `focei-outer-hessian*`, `cov-analytic`, `focei-fast-grad`) are unchanged apart from the two `focep` failures already present on main.

## Open question

`est="flaplace"`/`"fagq"` report "Full Laplace" but still use the Gauss-Newton determinant (their objective equals plain `laplace`). If the full determinant is what those names should mean, they could set `detHessian="conditional"` as well (and the AGQ gradient would need the node spread from the full H). Left as is here pending your view.


mattfidler:
## CI fix, main merge, antigravity review

Pushed 4 commits on top of c86f5d13f.

### CI failures (`test-predict.R:13/17`, test-coverage + release shard 2/3)
Not caused by this PR. Main is red on the same test. rxode2 5.1.8 (nlmixr2/rxode2@13ca4e5ac) changed `rxSolve(useLinCmt=)` to default `FALSE`, so the one-compartment fixture is now integrated as an ODE instead of solved analytically. That exposes a tolerance difference between the two entry points the test compares:

- `nlmixr2(fit, data, "predict")` inherits the fit's `rxControl` (`rtol=1e-3`)
- `predict()` uses `rxControl()` defaults

Against the exact `linCmt()` solution the relative errors are 5.3e-4 and 1.2e-6, while the test tolerance is 1e-4.

- **be74f242e** compares each entry point under the same solving options. Both pairs agree to 1e-8, and the tolerance is unchanged. The same fix for main is #1094.
- **7fd268366** merges origin/main (clean).

### Antigravity (gemini-3.1-pro-high), 3 rounds
- **Round 1:** no findings.
- **Round 2, `neta == 0` out-of-bounds read:** rejected by measurement. A model with no etas under `detHessian="conditional"` fits with `nAnalyticGradDirect = 0`, so the analytic gradient never reaches `C.G.rows(0, neta - 1)`.
- **Round 2, missing coverage:** confirmed. Every full-determinant test fixed the residual error and compared only `level`, so the sigma (`AR`/`AthR`) and omega components were never checked. The kernel itself was right: with the error estimated, `error` matches its central difference to 7 digits (43.900549 vs 43.90055). `om.chol.1`'s ratio to d(objf)/d(var) is -0.5349923 under both `detHessian="focei"` and `"conditional"`, i.e. the `diagXform="sqrt"` chain factor -4 var^(5/4).
  - **92532c148** adds that test: `level`, `error` and `om.chol.1` against central differences.
- **Round 3:** no findings.

### habit-hooks
Two functions this PR touches were flagged: `foceiDensityChain` (complexity, size) and `foceiHessianThird` (complexity).

- **96abc2e22** splits them into `foceiRhoCoefs` / `foceiDensityG` / `foceiDensityTn` and `foceiHessianProbeEta` / `foceiHessianThirdFill`. Expressions and loop order are unchanged, and no findings remain in the PR's changed lines. The remaining `parse-error` entry is the clang-tidy include-path gap; it also lists `src/armahead.h`, which this PR does not touch.

### Local verification (scratch install of the merged branch)
- `focei-det-conditional` 21/21, `predict` 11/11, `focei-fast-grad` 62, `focei-inner-conditional` 49, `focei-full-inner` 37
- `cov-analytic` 231, `cov-analytic-defaults` 39
- `focei-outer-hessian*` (7 files) 301 total

0 failures.

mattfidler:
Added 63f757cde (same change as #1094): new CI runs could not install rxode2, because rxode2 main now requires `rxode2lincmt`, which is GitHub-only (`Can't find package called rxode2lincmt`). The workflows now install `nlmixr2/rxode2lincmt` before `nlmixr2/rxode2`.

mattfidler:
Also needs tolerance loosening, more coming.

mattfidler:
## Second batch: full determinant for the f* methods, Hessian retry, NEWS

Pushed 63f757cde..ae79c0d91 (merged origin/main again, so #1094/#1093/#1087 are in).

### `flaplace`/`fagq` now mean the full determinant (answers the PR's open question)

`.foceiFullCtl` forces `detHessian="conditional"` alongside `fast=TRUE` and `innerHessian="conditional"`, for all six methods.

Checking the code first settled the AGQ half: in `LikInner2`, `calcEtaHessian` fills `H0`, and **both** the Laplace determinant (`logH0diag`) and the AGQ node spread (`Ginv_5 = inv(trimatu(H0))`) read that same `H0`; the call is `forOptimization=false`, so the `conditional` predicate keys on `detHessian`. The node spread therefore follows the full H automatically and cannot disagree with the determinant. What is still missing is the analytic outer **gradient** under AGQ: it declines on `fullDet && isAgq` and falls back to the FD gradient, which is the node-spread derivative term. Documented rather than hidden.

The label is unaffected (`src/inner.cpp:12846` keys on `innerHessian`).

One existing assertion had to be inverted: `test-focei-full-inner.R` required `flaplace`'s objective to **equal** plain `laplace`'s ("the conditional curvature does not move the marginal objective"). The full determinant moves it by design, so it now asserts they differ, plus that the control carries `detHessian="conditional"`.

### The analytic outer Hessian no longer falls back on the first failed probe

`foceiHessianThird` pinned the probes at `min(1e-12, fit tol)` and returned failure immediately. It now loosens the probe tolerance a rung at a time (`outerOdeRecalcFactor`, up to `outerMaxOdeRecalc` rungs, never looser than the fit's own tolerance) and only then declines to FD -- the escalation the inner problem already takes on a bad solve (`maxOdeRecalc`). Note the rung count comes from the existing controls (5 by default), not a hardcoded 3.

Rungs taken are reported as `$nHessTolRelax`. New `test-focei-outer-hessian-tol.R` drives it through a `NLMIXR2EST_HESS_PROBE_FAIL` test hook (mirroring `NLMIXR2EST_OUTER_FAIL_ID`): with one forced failure the Hessian still arrives and equals the clean one, and the counter is nonzero; with every rung failing it gives up only after spending the budget.

### NEWS

- `agqf`/`magqf`/`iagqf` were missing entirely -- they shipped in v7.0.2 but the `*f` entry never listed them. Added there.
- Rewrote the `flaplace`/`fagq` entry for the full determinant, incl. the AGQ gradient caveat.
- Bug-fix entry for the tolerance escalation.

### `test-predict.R`

The merge left a duplicated `md2 <- ...` line; removed. I had also added a comment claiming `predict()` falls back to `rxControl()` defaults -- that was **wrong**, and 72b51ed44's description is right. Measured one comparison per fresh R session against the cached fixture: bare `predict(fit, newdata)` matches `nlmixr2(fit, newdata, "predict")` exactly (0) and both differ from `rxControl()` defaults by 5.3e-4. My earlier probe built the fit in-session, which behaves differently from a deserialized one. Comment corrected.

### Verification

Clean rebuild (`rm src/*.o` first, since the merge brought new headers), then 14 files: `predict` 12, `saem-no-pop` 3, `focei-outer-hessian-tol` 7, `full-inner` 45, `det-conditional` 21, the seven `outer-hessian*` files, `focei-fast-grad` 62, `agq-fast-grad` 33. 0 failures.

habit-hooks: no findings in the PR's changed lines (scoped to the merge base). Antigravity: rounds 3 and 4 clean; round 4 specifically confirmed the determinant/node-spread consistency and that the retry loop is bounded, re-zeroes its cubes per attempt, and preserves the FD fallback.

PR #1093, opened 2026-09-15
imp/impmap/qrpem: nIter=0 is an E-step-only evaluation; add setOfv(fit, "imp")

### LINKED ISSUE #1091: nIter = 0 in the IMP family
## Summary

`nIter = 0` in the IMP-family controls segfaults R rather than erroring. It
affects `est = "imp"`, `"impmap"` and `"qrpem"` alike, and reproduces on a
stock `theo_sd` model with no other packages involved.

`impmapControl()` accepts `nIter = 0` without complaint but dies at fit time.

## Reproducible example

```r
library(nlmixr2)

one.cmt <- function() {
  ini({
    tka <- 0.45; tcl <- 1; tv <- 3.45
    eta.ka ~ 0.6; eta.cl ~ 0.3; eta.v ~ 0.1
    add.sd <- 0.7
  })
  model({
    ka <- exp(tka + eta.ka); cl <- exp(tcl + eta.cl); v <- exp(tv + eta.v)
    linCmt() ~ add(add.sd)
  })
}

# Segfaults for est = "imp", "impmap" and "qrpem"
fit <- nlmixr2(one.cmt, nlmixr2data::theo_sd, est = "impmap",
               control = list(print = 0, nIter = 0L))
```

Run non-interactively, each of the three exits with status 139 (SIGSEGV):

```
est=imp     exit=139
est=impmap  exit=139
est=qrpem   exit=139
```

`nIter = 1` and `nIter = 2` are fine, so the crash is specific to zero:

| `nIter` | outcome | `objf` | `env$impObj` |
| --- | --- | --- | --- |
| 0 | **segfault (139)** | — | — |
| 1 | ok | 117.440 | 134.926 |
| 2 | ok | 116.860 | 118.790 |
| default (100) | ok | 116.829 | 117.836 |

## Desired behaviour

Can we have an informative error (or, if zero EM iterations is meaningful, a documented
result)? A segfault loses the session and any unsaved work. 

## Background

I was looking for an evaluate-at-fixed-parameters mode for the IMP family,
the equivalent of PsN's `EONLY=1`, which suppresses the M-step so the objective
can be computed at a supplied parameter vector. This is needed for SIR, where every 
candidate must be scored on one fixed surface. `nIter = 0` was a guess at that. 

As far as I can tell there is no expectation-only mode exposed -
`impmapControl()` has no `EONLY` analogue and no `maxOuterIterations`, and
`nIter` counts full EM iterations, each of which runs an M-step and so moves
the population parameters rather than holding them fixed.

If an E-step-only evaluation is reachable some other way it would be great to be
pointed at it...

## Session info

```
R version 4.6.1 (2026-06-24 ucrt), x86_64-w64-mingw32
nlmixr2      7.0.1
nlmixr2est   7.0.3
rxode2       5.1.7
nlmixr2data  2.0.10
```

### LINKED ISSUE #1092: E-step-only (EONLY) evaluation at fixed parameters for imp/impmap/qrpem
## Summary

Add an E-step-only evaluation for the importance-sampling methods (`est = "imp"`,
`"impmap"`, `"qrpem"`): evaluate the objective at a supplied parameter vector without
moving the population parameters, the equivalent of NONMEM/PsN `EONLY=1`.

## Motivation

Raised in #1091.  SIR (sampling importance resampling) needs every candidate parameter
vector scored on one fixed likelihood surface.  Currently `nIter` counts full EM
iterations, each of which runs an M-step, so there is no way to hold the parameters
fixed; `nIter = 0` crashed R instead.

## Proposal

- `nIter = 0` runs the startup MAP and one E-step at the supplied parameters and
  skips the M-step (and any `nBurn` burn-in).
- The importance-sampling objective is reported in `$impObj`, with the E-step
  diagnostics (`$impCondMean`, `$impNeff`, ...) from that single E-step.
- `$runInfo` notes the E-step-only run and where the starting etas came from:
  `etas 0`, `etas from etaMat`, or `etas from the last fit` when refitting from a
  previous fit.
- `setOfv(fit, "imp")` / `setOfv(fit, "impmap")` add that importance-sampling
  objective to any fit (FOCEi, SAEM, ...) as another objective function row,
  consistent with how `setOfv()` already adds FOCEi/FO or SAEM quadrature objectives.

mattfidler:
## What needs fixing

1. **`nIter = 0` segfaulted R** for `est = "imp"`, `"impmap"` and `"qrpem"`. The EM loop never ran, so the post-loop code read per-iteration state that was never filled (`condVar[id]`), and the covariance would have been handed iteration `-1`.
2. **There was no evaluate-at-fixed-parameters mode** for the importance-sampling methods -- the NONMEM/PsN `EONLY=1` equivalent needed to score every candidate on one fixed surface (SIR). `nIter` counts full EM iterations, each of which runs an M-step and moves the population parameters.
3. **`test-predict.R:21` fails on `main`**, so this branch could not go green without repairing it. After #1094, `predict(fit, newdata)` with no solving options inherits the fit's `rxControl` (`rtol=1e-3`), while `nlmixr2(fit, data, "predict", control = rxControl())` uses the defaults (`rtol=1e-6`) -- the assertion compared two different tolerances.

## What this does

- `nIter = 0` is now the E-step-only evaluation: the startup MAP and one E-step at the supplied parameters, no M-step, no burn-in. The importance-sampling objective is in `$impObj`; `$impIter` is 0 and `$impEonly` is `TRUE`.
- `$runInfo` records the run and where the starting etas came from: `etas 0`, `etas from etaMat`, or `etas from the last fit`.
- `setOfv(fit, "imp")` / `setOfv(fit, "impmap")` add that objective to any fit as an `IMP`/`IMPMAP` row, like the existing FOCEi/FO and SAEM quadrature rows.
- `nIter` is validated with `checkmate` (`lower = 0`).
- `setOfv()` is flattened into `.setOfvAdd()` / `.setOfvSaemQuad()` helpers (no behavior change), clearing the deep-nesting lint the new branch made worse.
- `test-predict.R` passes the solver defaults on both sides and asserts the inheritance contract directly. Measured on a one-compartment fit: `predict()` defaults and `nlmixr2(..., "predict")` inheriting the fit's control agree exactly (relative difference 0); either against explicit `rxControl()` defaults does not.

Fixes #1091
Fixes #1092

## Test plan

- [x] The issue's reproducible example exits 0 for `est = "imp"`, `"impmap"` and `"qrpem"`; thetas stay at the `ini()` values.
- [x] `test-impmap.R` (203 passed): `nIter = 0` fits (parameters unchanged, `impObj` equals the first E-step of an `nIter = 1` fit, `$runInfo` notes for fresh and inherited etas) and `setOfv(fit, "imp")`.
- [x] `test-imp-burnin.R` (28 passed), `test-imp-mapiter.R` (20), `test-qrpem.R` (72) unchanged for `nIter >= 1`.
- [x] `test-saem-aic.R` (21 passed), `test-cov-condition.R` (17) for the refactored `setOfv()` branches.
- [x] `test-predict.R` 12 passed, was 10 passed / 1 failed (and fails identically on a plain `origin/main` build).
- [x] CI green on `72b51ed44`: all 10 checks pass, including the shard 2/3 and test-coverage jobs that were failing.


mattfidler:
Merged `main` and added one extra commit that is not part of the #1091 fix.

`test-predict.R:21` fails on `main` itself (main HEAD's own CI shard 2/3 fails the same assertion with identical numbers, and a plain `origin/main` build reproduces it locally), so this PR could not go green without it.

Cause: after #1094, `predict(fit, newdata)` with no solving options inherits the fit's `rxControl` (`rtol=1e-3`), while `nlmixr2(fit, data, "predict", control = rxControl())` uses the defaults (`rtol=1e-6`) -- so the assertion compared two different tolerances. Measured on a one-compartment fit: `predict()` defaults vs `.nlmixr()` inheriting the fit's control agree exactly (relative difference 0), and either one against explicit `rxControl()` defaults does not.

The test now passes the defaults on both sides and asserts the inheritance contract directly (12 passed, was 10 passed / 1 failed).

PR #1094, opened 2026-09-15
test(predict): compare predict entry points at matched solver tolerances

mattfidler:
Fixes the `test-predict.R:13` / `:17` failures on main (R-CMD-check and test-coverage red since b349343b3 / 0e47a6b9e; also on PR #1090).

## Cause

rxode2 5.1.8 (nlmixr2/rxode2@13ca4e5ac) changed `rxSolve(useLinCmt=)` to default `FALSE`. The one-compartment fixture used to be converted to `linCmt()`, so both predict paths were exact and matched. Now it is integrated as an ODE, which exposes a solver-tolerance difference between the two entry points:

- `nlmixr2(fit, data, "predict")` adapts the fit's `rxControl` (FOCEI `sigdig=3` -> `rtol=1e-3`, `atol=1e-6`)
- `predict(fit, data)`, and `predict()` with `fit$control` spread into `...` (no top-level names match an `rxControl()` argument), use `rxControl()` defaults

Measured against the exact `linCmt()` solution (`useLinCmt=TRUE`), the max relative error is 5.3e-4 for `nlmixr2(fit, "predict")` and 1.2e-6 for `predict()`. The test compared them at `tolerance = 1e-4`.

## Fix

Test only. Each comparison now runs both entry points under the same solving options: the fit's `rxControl` passed to `predict()`, and `rxControl()` passed to `nlmixr2(fit, "predict")`. Both pairs agree to 1e-8, so the 1e-4 tolerance is unchanged.

## Verification

`test-predict.R` passes 11/11 locally, run against a build of this commit merged into #1090's branch (main plus that PR, which does not touch the predict code). CI on this PR covers main by itself.


mattfidler:
Added a8972d3ca: every job on this PR failed at dependency install (`Can't install dependency rxode2lincmt (>= 0.1.0)` -> `Could not solve package dependencies`). rxode2 main now requires `rxode2lincmt`, which is GitHub-only, so `nlmixr2/rxode2lincmt` is now installed before `nlmixr2/rxode2` in every workflow that installs rxode2 from GitHub (R-CMD-check, test-coverage, pkgdown, slow-tests, runtime-benchmarks, rhub). This mirrors rxode2's own CI change (nlmixr2/rxode2@57c81c1fe). The failure affects main and every open PR, not just this one.

mattfidler:
All checks green. The first attempt of `ubuntu-latest (release shard 1/3)` crashed mid-suite with `malloc(): unsorted double linked list corrupted (core dumped)`, which is heap corruption, not a test failure. This PR changes only a test file and CI config, so it cannot cause that. The output just before the crash is a `linCmt()` mixture FOCEi fit falling back from the analytic to the "r,s" covariance; `test-focei-mix-cov.R` or `test-mix.R` are the likely files. Re-running the failed job passed, the same shard passed on #1090, and both files ran clean locally twice each under `MALLOC_CHECK_=3`/`MALLOC_PERTURB_`. So it is intermittent, not fixed.

PR #1105, opened 2026-09-18
perf(odeSwap): swap peer models into the solve pool from C++ (follow-up to #1049)

hiddevandebeek:
Follow-up to #1049: "The outer gradient needs to be in the ode pool so that it can be used efficiently without swapping out memory."

## What this changes

Swapping a peer model into the shared FOCEi solve pool (the augmented outer-gradient model, the outer Hessian's probes, the AGQ node model) is now a C call. `odeSwapEsInstall()` used to call back into R (`rxEventSensLoadModel()`) on every batch to re-derive the peer's event-sensitivity shape, so each gradient/Hessian pass crossed into R at every batch boundary. Now the first install of a slot still goes through R (it owns the dim derivation), the six dims are read back once with `rxode2EventSensGetDims()`, and every later swap is `rxode2EventSensLoadFull()` on the model's preserved `trans` -- the API added in rxode2 `9e007dfa0` for exactly this. `.odeSwapInfo()` reports `esInstallC`/`esInstallR`.

The analytic outer gradient's Omega^-1 and its estimation-scale derivatives come from the native Cholesky map the objective already maintains (`foceiHessianOmega`, shared with the outer Hessian) instead of the `rxSymInvChol` closure on every evaluation. The native values are verified against the R handle at first use (1e-8) and the handle stays the fallback; `$env$omegaGradFast` records which one ran.

No optimizer, control or user-facing behaviour changes. Per theo_sd fit the R round trips drop from 84 installs to 4.

## Measured

Clean `R CMD INSTALL` (-O2) of main and this branch, same session, cores=1, `user.self` median of 3 runs, `fast=TRUE`, `covMethod=""`:

| fit | outerOpt | main (s) | branch (s) | path |
|---|---|---|---|---|
| theo_sd | lbfgsb3c | 1.69 | 1.46 (-14%) | identical |
| theo_sd | nlminb | 1.68 | 1.49 (-11%) | identical |
| theo_sd | trust | 2.37 | 2.12 (-10%) | identical |
| theo_md x10 | lbfgsb3c | 3.80 | 3.21 (-16%) | identical |
| theo_md x10 | nlminb | 5.83 | 4.97 (-15%) | identical |
| theo_md x10 | trust | 17.27 | 15.65 (-9%) | identical |

Heap and RSS unchanged (358-361 MB / 422-439 MB). Warfarin's path differs from main at the 15th digit of the om.chol gradient (the native map versus the closure), deterministic on both sides; the objective agrees.

## Tests

`test-odeswap.R` counts C installs across a short fit; `test-focei-fast-grad.R` asserts `omegaGradFast == 1L` next to the existing central-difference comparison. `test-focei-outer-trust.R`, `test-focei-outer-hessian-family.R` and `test-focei-det-conditional.R` are unchanged and pass.

The trust outer optimizer is untouched here; per your comment that is the next step once this is in.


hiddevandebeek:
@mattfidler try number 2; if still not what you mean, lets discuss in a short meeting somewhere next week

mattfidler:
This is more what I was thinking of originally.  I will look at it soon.

mattfidler:
## Merged main + independent review (Gemini 3.1 Pro, via Antigravity)

**Merge:** `origin/main` merged cleanly (7c6160ff7). Main's new `air` format check flagged `tests/testthat/test-odeswap.R`, so it was reformatted with air 0.11.0 (13741cded). Clean `R CMD INSTALL` of the merge builds, and `test-odeswap.R` (119 pass) and `test-focei-fast-grad.R` (67 pass) pass with no failures or skips.

### Round 1 -- 2 findings, both fixed (d94dfd729)

1. **`gradDirectOmega` (`src/inner.cpp`): a failed verification call to the handle had two side effects.** On first use, if the native map succeeded but `gradDirectOmegaHandle` failed (codes 106/107/108), `okR` was false: `_omGradFastState` went to `-1`, which disabled a native map nobody had disproved for the rest of the fit, and then the handle was called a second time. **Fix:** decline right away with the handle's own code and leave the state untried, so verification retries on the next evaluation. When the native map and the handle genuinely disagree, the state still goes to `-1` as before.
2. **Missing coverage of the handle fallback.** The test only asserted `omegaGradFast == 1`. **Fix:** `test-focei-fast-grad.R` now also fits with `NLMIXR2_NO_FAST_OMEGA=1`, which forces the `rxSymInvChol` handle route. It asserts the native map was not used, the analytic gradient was still consumed, and objf/fixef match the fast-path fit.

### Round 2 (final code) -- no remaining issues

- It confirmed the `gradDirectOmega` state machine: verified, mismatch, transient native-map failure after verification, and `nom == 0`.
- **Rejected:** its one claim was that the `esDims` gate (`src/odeSwap.cpp:272`, `esNParam > 0`) sends an "eta-only" event-sensitivity model down the R path for good. In rxode2, `nParam` is `length(map$sensParams)`, meaning ALL first-order sensitivity parameters with etas included, not structural thetas. `nParam2` counts second-order directions over those same parameters, so `nParam2 > 0` means `nParam > 0`. Even a missed case would only fall back to the old R install, not produce wrong results.

The fix commit is small: it changes only `gradDirectOmega` and adds 4 test assertions. A follow-up (`refactor(focei): split gradDirectOmega's first-use verification out`) moves the verification into `gradDirectOmegaVerify` with no change in behavior, which clears a complexity lint flag on the function. Both test files pass again after it.

mattfidler:
looks good to me

PR #1106, opened 2026-09-18
setCov(): a fit-aware cache key, and setCov(fit) <- value

mattfidler:
Follow-up to #1098. It adds two extension points so packages can plug covariance methods into `setCov()`. `nlmixr2sir` uses both for `setCov(fit, "sir")`; that PR is nlmixr2/nlmixr2sir#1.

## `setCovOptions(control, fit)`: a cache key that can see the fit

In #1098 the cache key was `unclass(control)`, so it could not include anything about the fit. A SIR covariance is seeded from another covariance on the fit, so with that key a changed seed would reinstall a stale SIR covariance. `.covOptionsResolve()` now dispatches on the new generic `setCovOptions()`:

- `setCovOptions.default`: `unclass(control)`, the same behaviour as before.
- `setCovOptions.rsControl`: the existing resolution of `foceiControl` options.
- A package adds a method for its own control class to put fit state into the key. It can also leave out options that don't change the result, such as `workers`.

## `setCov(fit) <- value`: install an already-computed covariance

- `setCov(fit) <- m` installs a matrix as `"user"`; `setCov(fit, "name") <- m` uses that name. Either way it gets the same PD check, SE refresh and `covList` stash as `setCov()`.
- `setCovValue(value, fit, method)` is the generic behind it. A method returns `list(cov, method, options)`, and the recorded `options` let a later `setCov(fit, method, control = ...)` reuse it when they match.
- A method can also return `extra`, a named list assigned into the fit env **only after the install succeeds**. `nlmixr2sir` uses it to keep the SIR result as `fit$sir`.

## Tests

`test-cov-setcov-generic.R` covers:
- a seed-dependent key that recomputes when the installed covariance changes
- the default key
- matrix install, rejection and swap
- a `setCovValue()` method with `extra`, including that nothing is stored when the install fails

That file and the other setCov-related suites pass locally: `cov-scope-swap`, `cov-condition`, `cor-no-cov-1038`, `cov-fdfull-install`, `cov-decouple`, `cov-decouple-saimp`, `focei-mix-cov`.

Note: `DESCRIPTION` is still 7.1.0, so downstream packages can't yet require this version. It needs a bump before `nlmixr2sir` can depend on it.


mattfidler:
Not sure why it is tagging me as Matt Fidler, as opposed to Matthew Fidler like everywhere else, but it is fine.

mattfidler:
Not sure why it is tagging me as Matt Fidler, as opposed to Matthew Fidler like everywhere else, but it is fine.

PR #1112, opened 2026-09-19
perf/fix(focei): trust outer evaluates curvature only at acceptable points, and holds a parameter on an active bound

hiddevandebeek:
Two changes to the `outerOpt="trust"` driver (`R/focei.R` only; nlminb/lbfgsb3c paths untouched), one commit each.

## 1. Gradient and Hessian only at points trust can accept

`RcppTrust` reads the gradient and Hessian only at points it accepts, and it can only accept a trial whose value is below the incumbent's (rho >= 1/4, or the termination branch with ftry < f). The driver used to compute the full triple at every trial point, and the analytic outer Hessian is the expensive part of an iteration (~12 objective evaluations), so on perturbed starts up to two thirds of the Hessians were computed and discarded.

`.trustOuterObjfun` now returns value-only (with the incumbent's cached gradient/Hessian) when the trial value is not below the incumbent's, and tracks the incumbent by replaying trust's own acceptance test (same preddiff, rho, fterm/mterm; where the replay could disagree it errs toward "rejected", which only costs an evaluation). A restart at the incumbent returns the cached triple.

Measured on 9 trust/analytic cases (interleaved separate processes, 1 thread, best of 2), main vs this commit: Hessian evaluations 327 -> 166 in total, wall 175 -> 96 s (1.1x-2.6x per case), objectives identical to 3 decimals in 8/9 (warfPK high 458.523 vs 458.728, within that fit's noise floor).

## 2. Hold a parameter on an active bound instead of converging onto it

`outerOpt="trust"` could report convergence on a box bound with a large gradient. `RcppTrust` is unbounded and the driver rejected a trial outside the box as infeasible, which only shrinks the region. On an active bound every Newton step leaves the box, so every iteration is region-limited: the region halves per accepted step, the run converges linearly onto the bound, and `outerTrustFterm` is met there. Traced on `theo_md` from a perturbed omega start, the last 15 accepted steps were 2e-4 down to 7e-10 long with rho ~ 1 and 13 infeasible trials between each, ending with a gradient of (13, 21, 39, -33, -64, -22, -15, 132) and 3 pointless restarts.

Once the model's step has left the box through a parameter twice from the same incumbent, the driver now holds that parameter on its bound: its gradient entry and Hessian row/column are masked (decoupled unit diagonal) so trust's step in it is exactly zero, trust is re-entered from the incumbent on the free parameters, and the parameter is released when the gradient at a converged point pulls it back inside. The Newton-decrement verdict is taken on the free parameters (a KKT point, not a stationary one). `$optReturn$activeBounds` lists the parameters held at the end.

Same 9 cases, before = commit 1 only, after = both. Reference optima from `outerOpt="nlminb"` on the same starts: oralMM 116.86-116.99, theoMdComb 346.63, warfPK 458.9-466.9.

| model | start | before objf | after objf | before nH | after nH | wall before -> after |
|---|---|---|---|---|---|---|
| oralMM | nominal | 116.808 | 116.807 | 21 | 21 | 5.1 -> 5.1 s |
| oralMM | high | **270.832** | **116.808** | 21 | 23 | 5.0 -> 5.2 s |
| oralMM | rnd2 | 116.808 | 116.808 | 22 | 22 | 5.3 -> 5.0 s |
| theoMdComb | nominal | 346.626 | 346.626 | 8 | 8 | 5.9 -> 5.5 s |
| theoMdComb | high | **487.956** | **346.626** | 21 | 20 | 14.9 -> 13.7 s |
| theoMdComb | rnd2 | 346.626 | 346.626 | 11 | 11 | 8.4 -> 8.4 s |
| warfPK | nominal | 452.194 | 452.206 | 21 | 23 | 9.1 -> 9.7 s |
| warfPK | high | 458.728 | 458.529 | 25 | 33 | 13.9 -> 16.5 s |
| warfPK | rnd2 | 454.181 | 454.174 | 16 | 16 | 8.2 -> 7.7 s |

The two bold cases are the defect: an omega variance on its `diagOmegaBoundLower` floor (the perturbed start put it 100x above the optimum). A later nine-start sweep against `origin/main` found this is not occasional -- every fit whose omega landed on the floor stopped 12 to 596 objective units short, and every fit whose omega stayed inside the box was unaffected. With the hold, the other parameters converge with the variance pinned, the gradient then points back inside, the parameter is released and the run finishes at the optimum nlminb/L-BFGS-B reach, at the same cost. The other seven are unchanged (warfPK's small differences are the noisy-objective path, also present before this change; its optima differ from nlminb's in the other direction).

## Tests

`test-focei-outer-trust.R` drives `.trustOuter()` over a bounded quadratic whose minimum is outside the box (reaches (1, 1) exactly with both parameters held and the gradient still (2, 8); under 15 objective calls); over Rosenbrock under `x2 >= 0`, where the bound is held and then released (`activeChanges == 2`, nothing left held -- without the release it stalls on the bound at `x1 = 0.16`); and over Rosenbrock unbounded, counting the callbacks so the lazy skip is visible at all (a trial worse than the incumbent must cost one `fn()` and nothing else). It also sweeps `maxOuterIterations` 1..20 asserting the result stays inside the box and no `call to objfun` warning escapes. Two end-to-end `theo_sd` fits: `tka` capped below its optimum (`activeBounds == 1L`), and `eta.cl ~ 60` putting an omega on its `diagOmegaBoundLower` floor (`activeBounds == 5L`) -- against `origin/main` that second fit stops at 357.16 where `lbfgsb3c` reaches 202.27, reported as converged. Both assert the objective and fixed effects match `outerOpt="lbfgsb3c"`.

On a clean install: test-focei-outer-trust.R 103, test-focei-det-conditional.R 21 pass, 0 failures. `air format --check` clean.

The review that produced the later commits on this branch is in [a comment below](https://github.com/nlmixr2/nlmixr2est/pull/1112#issuecomment-5746540286): two correctness fixes (an out-of-box parameter vector when a hold was cut short, and a spurious `error in ... call to objfun` note reaching `$runInfo`), the test gaps above, and several reviewer claims checked and rejected by measurement.


mattfidler:
## Independent review of this branch

Reviewed with Kilo Code (non-Claude models: DeepSeek, GLM, Kimi, Grok) plus my own reading. Four of six Kilo runs produced nothing usable (one hit a workspace permission boundary, one hung to timeout, one returned empty, one died on an upstream 503 mid-answer); **Kimi** and **Grok** produced full reviews and agreed with each other. Everything below was verified by measurement before acting -- the standalone harness drives `.trustOuter()` directly, so the counts are over randomized runs, not anecdotes.

### Correctness bugs found and fixed (c952d933)

**1. The returned parameter vector could be outside the box.** A hold is signalled by a condition raised from the objective, so `RcppTrust::trust()` returns through its error path, where `argument` is `argument_at_error` -- the trial that *left the box*, not anything it accepted. When the run loop could not re-enter (active-set change budget spent, or iteration budget gone) it took that as the answer. **4 of 400** randomized runs returned an infeasible `x`. `foceiCustomFun` feeds that straight to `foceiOuterFinal`, so e.g. an omega variance below its `diagOmegaBoundLower` floor becomes the reported fit. Now hands back the incumbent and unwinds the hold that was never applied, so `activeBounds` cannot name a bound the returned point is not on. **0 of 400** after.

**2. `error in first/last call to objfun` leaked into `$runInfo`.** When the deliberate `trustOuterSnap` condition lands on trust's post-loop final evaluation (`error_source == 3`), `trust_solve_r` emits that `Rcpp::warning`. **9 of 400** runs. It is this driver's own control flow, not a failure. Now muffled, narrowly: only while `snapPending` is set *and* the message matches.

### Test gaps closed (70c8ce44, f671e977)

**The "held and then released" test never released.** From `(3,2)` on `(x1-2)^2 + 4*x2^2` the Newton direction is toward `(2,0)`, so only `x2` ever leaves the box and `activeChanges` is 1. Confirmed by deleting the release branch: the assertions still passed. Replaced with Rosenbrock under `x2 >= 0`, which stalls at `x1 = 0.16` without the release.

**Nothing could see the lazy skip** -- skipping the gradient and Hessian at a trial that cannot be accepted does not move the optimum. Now counts the callbacks: a worse trial must cost one `fn()` and nothing else. Without the skip, `gr == fn == 28` and the test fails.

**The omega floor had no test**, and that is the bound that actually bit. `diagOmegaBoundLower` puts a diagonal omega's lower bound at its own initial estimate over 100, so a start ~100x above the optimum makes it active. Measured against `origin/main` on `theo_sd` with `eta.cl ~ 60` (floor 0.6, data optimum ~0.11): trust stops at **357.16** where `lbfgsb3c` reaches **202.27** -- 155 objective units, *reported as converged*. Both reach 202.27 now.

### The failure was every on-floor start, not an occasional one

Nine perturbed starts, `origin/main` vs this branch, trust minus the best of nlminb/lbfgsb3c:

| start | om0 | omega on floor | main | branch |
|---|---|---|---|---|
| 1-3 | 0.3, 3, 10 | no | 0.0000 | 0.0000 |
| 4 | 30 | yes | 14.25 | 0.0000 |
| 5 | 30 | yes | 15.22 | 0.0000 |
| 6 | 50 | yes | 595.67 | 0.0000 |
| 7 | 100 | yes | 12.24 | 0.0000 |
| 8 | 100 | yes | 44.80 | 0.0000 |
| 9 | 60 | yes | 154.90 | 0.0000 |

The split is exact: every failure is a case where the omega landed on its floor, every off-floor case is untouched. NEWS updated accordingly (7cf4ff29) -- it previously read "two of nine ... 154 and 141 objective units", which is true of the original grid but reads as a fluke.

### Reviewer claims checked and rejected

- **Cached curvature reaching an accepted point** (the central risk in the lazy scheme). The R-side replay of `rho`/`is_terminate` matches `trust_core_run` exactly: **0 occurrences in 3,676 trials** over random quadratics and Rosenbrock.
- **Kimi: zero tolerances** (`outerTrustFterm = 0`, `outerTrustMterm = 0` -- `checkmate` does allow 0) letting `rho = +Inf` accept a worse trial carrying cached curvature. **0 in 30,011 trials** at those settings. `preddiff == 0` at a trust-region minimizer implies `p == 0`, so the trial *is* the incumbent and takes the exact-match branch, which returns the right curvature.
- **Kimi: the subproblem's hard case moving a held coordinate.** **0 of 2,037** evaluations made with a hold active, across 400 randomized runs including an indefinite Hessian. The masked model keeps the step in a held coordinate exactly zero.
- **Kimi's stale-incumbent sequence** is self-contradictory: it needs `f(P) <= f(T2) < f(T)` together with `f(T) < f(P)`. A stale incumbent sits *above* trust's value, which makes the shortcut strictly harder to reach -- the safe direction, as the code comment says.
- **Unpatched `argument` when `!.atInc`** -- `!.atInc` never fired in **500** randomized runs (59 of which exercised a release). A patch there would be dead code.
- **`infeasibleRun` not reset on release** (my own concern, not a reviewer's): adding the reset changes nothing over 500 runs -- identical releases, worst `activeChanges`, unconverged count, and a bit-identical objective sum. The concern was also unfounded: a release does not change the incumbent, so the two out-of-box steps genuinely are "from the same incumbent".

### Comments the change had made false (70c8ce44)

`.trustOuterBfgs` and `.trustOuterCurvature` both claimed BFGS secant pairs are consecutive objfun calls; with lazy curvature they are consecutive *improving* points. The replay comment said `rho >= 1/4` where the code takes `>`; it now says why the strictness is deliberate. The NEWS headline said "only at points the trust region can accept" -- "could accept" is what the code does.

### State

`test-focei-outer-trust.R` 103 assertions and `test-focei-det-conditional.R` 21 (the only two files touching `.trustOuter`; its single caller is `foceiControl.R:1708`, reachable only via `outerOpt="trust"`), 0 failures. `format-check`, `lint` and CodeFactor green -- `format-check` was red until 8331726f, my additions were not air-formatted.

mattfidler:
## Antigravity (Gemini) review -- four rounds

Four rounds of `agy --model gemini-3.1-pro-high`, re-reviewing after each fix. Findings narrowed each round and the last produced no new reachable wrong-answer path, so I stopped there. Three fixes came out of it (`280238344` -> reverted, `eaed04881`, `b528b8189`, `7ce7dca05`). Every claim below was checked against the code or measured before acting.

### What it got right

**The acceptance replay disagreed with the solver at exactly rho = 1/4** (`b528b8189`). `RcppTrust` rejects on `rho < 0.25`, so at exactly a quarter it accepts; the replay took `>` and disagreed there by construction. The curvature handed back at that trial is still correct -- an improving trial never reaches the value-only shortcut -- but a disagreement is what puts the replayed incumbent out of step with the solver's, and out of step is where the shortcut stops being sound. `>=` is the exact replay. Outcome bit-identical over 500 randomized runs.

**Lagging is not a safe resting state.** Round 1 I had added a margin (`rho > 0.25 * (1+1e-6)`) so the replay could only ever lag behind the solver, on the reasoning that lagging merely costs an evaluation. Round 2 showed that is false: once the two differ, `.pred` is rebuilt from a different base than the solver's `preddiff`, the two rho values stop being the same number, and the replay can then accept where the solver rejects -- lagging cascades into leading, and the margin *widened* the window that starts it. Measured over identical 8090-trial trajectories the margin separated nothing, so it bought no safety for the risk it added. Reverted in `eaed04881`.

**So the fix is the observable, not the arithmetic** (`eaed04881`). Whatever the cause, a run that ends at a point the incumbent cannot name is one where the shortcut may have answered from a point the solver never accepted. The driver now drops the incumbent there -- a re-entry rebuilds it from the solver's own point -- and refuses to read a Newton decrement out of curvature it cannot vouch for, so the run is honestly marked not stationary instead of reporting a confident wrong number. Never fires on any of 500 randomized runs (outcome bit-identical); the test drives it by poisoning the incumbent directly.

**`optReturn$value` was dropped** (`7ce7dca05`). The branch that ends at the incumbent rebuilds the gradient and Hessian from it but not the value, so `optReturn` could report a gradient for a point whose value it had dropped. One line.

### What I checked and rejected

- **"the returned `x` can be out of the box because `.trustOuter` does not project an infeasible `par`"** -- wrong mechanism. An infeasible start raises a clean `parinit not feasible` error, and that rejection predates this PR.
- **"`infeasibleRun` leaks across the `next`/restart re-entries and triggers erroneous snaps"** -- resetting it on both the release and the restart path changes nothing over 500 runs: identical release count, worst `activeChanges`, unconverged count, and a bit-identical objective sum. The premise is also wrong: a release does not change the incumbent, so the two out-of-box steps genuinely are "from the same incumbent".
- **"the snap error path can bypass the out-of-step guard when the failing trial coincides with the incumbent"** -- unreachable. The `if (state$snapPending)` block at `R/focei.R:507` runs *before* `.atInc` is computed and either re-enters or breaks with the incumbent restored.
- **"a missing `value` leaves `$objf` empty"** -- no. `foceiCustomFun` (`src/inner.cpp:10106-10114`) reads `x`, `convergence` and `message` only; `$objf` is recomputed by `foceiOuterFinal`. The fix above is for `optReturn`'s internal consistency, not for `$objf`.

### What remains, stated plainly

The last round's surviving finding is the floating-point difference between the replayed `.pred` (rebuilt from `x - .inc$x`) and the solver's internal `preddiff`. That one is **irreducible at the R/C++ boundary** -- the solver does not expose its step -- so it is contained rather than eliminated: if it ever separates the two, the run ends away from the incumbent, the guard fires, and the result is reported as not stationary rather than as a wrong gradient. 0 separations in 8090 trials, but that is a knife edge randomized testing will not find, which is why the containment exists.

Tests: `test-focei-outer-trust.R` 109, `test-focei-det-conditional.R` 21, 0 failures.

mattfidler:
## Benchmarked on the larger vignette models -- and what it says about this PR

Extending the benchmark from nlmixr2est#924 to `nimo` (QSS TMDD, 3 ODE states) and `wbc` (Friberg, 7 states), models verbatim from the `nlmixr2` vignettes. Objective / outer evaluations / wall:

| | bobyqa (`focei` default, FD) | lbfgsb3c (`foceif` default) | **trust + analytic Hessian** |
|---|---|---|---|
| nimo | **145.2633** / 253 / 30s | 2400.0000 / 21 / 980s | 2400.0000 / **2** / 1143s |
| wbc | **965.4910** / 76 / 4.6s | 8594.6985 / 23 / 83s | 8633.2681 / 41 / 129s |

**Read this as a `fast=TRUE` result, not an outer-optimizer result.** Both `fast=TRUE` arms sit ~16x and ~9x above `focei`, and the outer optimizer does not move it. With *zero* outer iterations -- no search, same initial estimates -- `fast=FALSE` and `fast=TRUE` already disagree (nimo 1484.46 vs 2400; wbc 1364.41 vs 8808.64), because the `fast=TRUE` inner problem collapses two of three ETAs to ~1e-14. Reproduces identically on `origin/main`, so **this PR does not cause it**; filed as nlmixr2est#1115.

### What it does say about this PR

The one thing that separates the two `fast=TRUE` arms is that the trust driver **notices**. On nimo it reported `underConverged = TRUE` and stopped after **2** outer evaluations; `lbfgsb3c` reported nothing and spent 980s reaching the same place. That is the Newton-decrement gate this PR's driver is built around doing its job on a fit that is genuinely broken -- which is the behaviour I would want from a default, independent of the objective it lands on.

It also means the small-model benchmark in the previous comment is currently the only one that can distinguish outer optimizers at all. On that corpus trust reaches the lowest objective on 4 of 5 and closes a reproducible 3.6-unit gap on `oral 2cmt` where `lbfgsb3c` is the outlier (113.14 vs 116.75, with bobyqa 113.18 and nlminb 113.14 agreeing with trust) -- but at 2-4x the wall time.

### Recommendation on making `trust` the `foceif` default

Not yet. The accuracy case rests on five small reconstructed models; the wall-time cost is consistent and real; and until nlmixr2est#1115 is fixed the larger models cannot adjudicate it. Fix #1115, then re-run both tables -- at that point the question is answerable on models that actually stress the optimizer.

Nothing here changes the code in this PR.

hiddevandebeek:
I will look into stalling in the beginning for the fast path coming week. I have some datasets where this happens too.

